PR #86, opened 2016-11-30
Build fixes for Gentoo

junghans:


jbakosi:
Cool! Thanks for porting to gentoo! Looks good to be merged. Onbe minor questions:
PATH_SUFFIX seems to be PATH_SUFFIXES in cmake's find_path(). Is that a typo or both will work for backwards compatibility? (Assuming the singular was supported by an earlier cmake version).

junghans:
You are right, in newer cmake it is called `PATH_SUFFIXES`, will use that!

PR #87, opened 2016-12-01
Pstreams

jbakosi:
Christoph: Can you test if pstream.h can be found on gentoo?


junghans:
```
git fetch origin pull/87/head && git checkout FETCH_HEAD && cmake ../src && make
```
works for me!

PR #90, opened 2016-12-05
System libs

jbakosi:
I have pulled over the branch from your repo and added a new commit, 1f5887c, on top of it.

I did not know about `-DCMAKE_DISABLE_FIND_PACKAGE_<PackageName>`. This makes the `NO_SYSTEM_<PackageName>` options unnecessary. To keep the change sets as small as possible, I just made this one consistent with the rest of the code and will do the switch to `-DCMAKE_DISABLE_FIND_PACKAGE_<PackageName>` in another step.

In the meantime, can you test this the way you originally needed the H5Part finder to work to see if I have screwed it up in 1f5887c?


junghans:
In general, you want Camelcase or not in cmake?

cmake/FindH5Part.cmake
```diff
@@ -1,32 +1,48 @@
...
-#------------------------------------------------------------------------------#
+################################################################################
+#
+# \file      cmake/FindH5Part.cmake
+# \author    C. Junghans
+# \copyright 2016, Los Alamos National Security, LLC.
+# \brief     Find the H5Part library
+# \date      Mon 05 Dec 2016 09:02:53 AM MST
+#
+################################################################################
```
junghans: I am never sure what the point of this block is! Isn't `git blame` more accurate?

jbakosi: The whole header? Doxygen uses it to generate metadata on a file, e.g., http://quinoacomputing.github.io/quinoa/_lin_sys_merger_8h.html#details. Though you are right:

- this does not happen for the Find*.cmake files, and
- `git blame` is another way to do it.

jbakosi: Also, whever I edit a file in vim, it updates the `\date` line using vim's autodate plugin so that part is automatic - but only for me and those who bother to set this up. So I agree this is kinda cumbersome.

cmake/FindH5Part.cmake
```diff
@@ -1,32 +1,48 @@
...
+#
+#  Set H5PART_ROOT before calling find_package to a path to add an additional
+#  search path, e.g.,
 #
-#  H5Part_INCLUDE_DIRS - where to find flecsi.h, etc.
-#  H5Part_LIBRARIES    - List of libraries when using flecsi.
-#  H5Part_FOUND        - True if flecsi found.
+#  Usage:
+#
+#  set(H5PART_ROOT "/path/to/custom/h5part") # prefer over system
```
junghans: Wouldn't `set(CMAKE_PREFIX_PATH "/path/to/custom/h5part")` do exactly the same?

junghans:
Hmm, I think we are bit inconsistent! If the macro is called `FindXXX.cmake` the variables should be called `XXX_LIBRARIES`, while if the macro is called `FindXxx.cmake` the variables should be called `Xxx_LIBRARIES`.

cmake/FindH5Part.cmake
jbakosi: That would force cmake to try to find every library at `/path/to/custom/h5part`. This way one can customize per lib.

jbakosi:
You are right. I never payed too much attention to the naming, but it would be better to make more consistent.

junghans:
I added tut to Gentoo in gentoo/gentoo@387774b1afa312bde6cd02f74f864c1a662ecc99.

junghans:
We are still missing:

- charm (just needs a find module update for gentoo)
- pegtl (see #91 )
- rngsse2(needs quinoacomputing/RNGSSE2#1) 
- testu01 ( see #93 )

jbakosi:
Just merged quinoacomputing/RNGSSE2#1. I will look into #91. Can you pursue #93? How about #92?

junghans:
@jbakosi : please review!

cmake/FindSEACASExodiff.cmake
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2016 Los Alamos Natl. Lab
```
jbakosi: Ah! I understand that the trilinos-provided Find*cmake stuff is unavailable until it is found (as chicken and egg) so we need this. This is pretty stupid, but I don't see a better way of doing this, unless somehow concocting it from the trilinos source under tpl/src/trilinos. I just looked and it does not seem too straightforward to extract some kind of FindSEACASExodus.cmake and FindSEACASExodiff.cmake from it.

I guess adding these two files remains the most maintainable solution for now...

cmake/TPLs.cmake
```diff
@@ -97,19 +104,27 @@ endif()
...
-  message(STATUS "Found SEACASExodus: ${SEACASExodus_LIBRARY_DIRS}")
+  message(STATUS "Found SEACASExodus: ${SEACASExodus_LIBRARIES}")
 endif()
 find_package(SEACASExodiff REQUIRED)
 if(SEACASExodiff_FOUND)
-  message(STATUS "Found SEACASExodiff: ${SEACASExodiff_LIBRARY_DIRS}")
+  message(STATUS "Found SEACASExodiff: ${SEACASExodiff_EXECUTABLE}")
 endif()
 
+#### RNGSSE2 library
```
jbakosi: RNGSSE2 now also has its FindRNGSSE2.cmake and also has some additional logic to help CI testing it being optional. Please merge the latest master.

junghans: I merge this wrong in 068e62884b1561f1be38daf32414ac56a4e9d7f3

cmake/FindTut.cmake
```diff
@@ -0,0 +1,47 @@
...
+#  if(Tut_FOUND)
+#    include_directories(${Tut_INCLUDE_DIR})
+#  endif()
+
+find_package(PkgConfig)
+
+# If already in cache, be silent
+if (Tut_INCLUDE_DIR)
+  set (Tut_FIND_QUIETLY TRUE)
+  pkg_check_modules(PC_TUT QUIET tut)
```
jbakosi: What is the point of the lines 32 and 34? It finds the system-wide `tut` for me on debian with them too.

cmake/FindSEACASExodus.cmake
```diff
@@ -0,0 +1,29 @@
+#------------------------------------------------------------------------------#
```
jbakosi: Unfortunately, the exodus debian package does not provide exodiff, so on debian/ubuntu one still has to build trilinos ;(

cmake/FindTut.cmake
junghans: If the pkg-config file isn't found cmake will say:
```
-- Checking for module 'tut'
--   No package 'tut' found
```
even though it says:
```
-- Found Tut: /usr/include
```
below. Without the `QUIET` in line 32, it would say that every time you run cmake even if Tut was found in an earlier run. In my experience that confuses users a lot, so that helps to keep the noise level down.

jbakosi: Debian has an ancient (2007) version of tut. Is there a way to detect the TUT version so we can reject it if it is not recent enough? In particular, `tut/tut_runner.hpp` is missing in that old version, while the newer one (as well as in ours at `tpl/src/tut') has it.

If there is not good way of detecting the `tut` version, we could include a search for the header 'tut_runner.hpp'. What do you think?

junghans: Usually I would parse the header for some define, but tut doesn't have any of these defines.
Check for `tut_runner.hpp` seems reasonable.

jbakosi:
Okay, I've done my best on debian, and with the above commits, this should be good to go. (I built and tested on my workstation using clang.) Please test on gentoo and let me know if it is okay, then I merge to master. Thanks.

junghans:
Ok, the only thing left to make quinoa a gentoo package is #93 and/or making testu01 optional.

cmake/TPLs.cmake
```diff
@@ -68,6 +68,9 @@ set(RNGSSE2_ROOT ${TPL_DIR}) # prefer ours
 find_package(RNGSSE2)
 if(RNGSSE2_FOUND AND NOT NO_RNGSSE2)
   set(HAS_RNGSSE2 true)  # will become compiler define in Main/QuinoaConfig.h
+else()
```
jbakosi: It is strange that these are defined when `RNGSSE2` is not found...

jbakosi:
This is looking great! Cleaned it up a little bit and fixed setting CHARM_COMPILER, which would have been overwritten when we build our own Charm++. Please build/test on gentoo once again and I'll merge.

cmake/TPLs.cmake
junghans: They are set empty! Otherwise if `RNGSSE2` isn't found, `RNGSSE2_INCLUDES` and `RNGSSE2_INCLUDES` have a value of `NOT-FOUND` and cmake will complain.

jbakosi: Ah. Okay.

junghans:
Works for me!

Issue #11, opened 2015-07-02
Find and add RNG library alternative to MKL and RNGSSE that works with threads


jbakosi:
MKL is non-free and RNGSSE does not work on Mac OS, cuRAND, Random123?

jbakosi:
Pretty much the same Issue #12.

jbakosi:
RNGSSE *does* work on mac and thus it is now the alternative to MKL which is truly optional.

jbakosi:
RNGSSE2 does work on Mac OS X. However, it does not work on PowerPC, e.g., IBM BlueGene/Q. If MKL is not available we have no RNGs.

jbakosi:
We now have 3 RNG libs: MKL, RNGSSE2, and Random123. See https://github.com/quinoacomputing/quinoa/issues/95.

PR #95, opened 2017-01-12
Random123

jbakosi:
Relates to https://github.com/quinoacomputing/quinoa/issues/11.

This adds [Random123](http://www.thesalmons.org/john/random123/releases/latest/docs/index.html) as a new RNG library and hooks up two of the four available RNGs. The ones not hooked up, EASNI and ARS, require the Advanced Encryption Standard (AES) instruction set (IS) on the CPU. While the AES IS is available on modern CPUs, default compilation does not necessarily enables it and since we don't desperately need all RNGs from Random123, I only enabled ThreeFry and Philox.

I believe, they are hooked up to everything that needs RNGs and both unit-, and regression-tested. Also updated the doxygen documentation.

Perhaps the only remaining task here is to make all RNG libraries (MKL, RNGSSE, Random123) optional but require at least one. Note that MKL is proprietary (thus not available everywhere), RNGSSE is almost pure x86 assembly (thus will not run everywhere), while Random123 should be pretty portable. We should detect what's available and what can run on the given architecture and enable/disable code based on that. Other than this, this is good to goo. @junghans, could you please test? (I'll test on TeamCity). Thanks.


junghans:
Added random123 to gentoo in gentoo/gentoo@9a89b76d13658e4ff055b3c0d7b696807b7e672c.

Testing now.

junghans:
When building against Gentoo's Random123, I found that you are using `uniform.hpp`, which isn't actually part of Random123's `include` directory, but lives in examples. I could add it to the installation, but just checking...

junghans:
I added the missing header in gentoo/gentoo@0d601682798febc90360cc8c9ab9e24053b1815d

junghans:
Yes, I install it under Gentoo now too. So I would just go with that workaround for now and fix it later if necessary. I think this is ready to merge otherwise.

junghans:
ok, I will make #90 based on this.

jbakosi:
Yeah, sounds good. I'm working on making RNGSSE2 optional. Basically, we will detect MKL, RNGSSE2, and Random123. The former two will be optional, while the last one (the most portable) will be required (from system or we build it.)

jbakosi:
With https://github.com/quinoacomputing/quinoa/pull/95/commits/bde476e5273d9637637787e110357e76a7a5e2b3, building RNGSSE2 is now conditional on the detected system architecture. If the arch is not x86, RNGSSE2 cannot build, so we don't build it and conditionally remove it from the code, including its unit and regression tests.

In the TPL cmake run, building of RNGSSE2 can be overridden (even on x86 arch) by passing the `-DNO_RNGSSE2=true` cmake arg. This will be useful for testing the absence of RNGSSE2 on x86 architectures, e.g., during CI.

If none of the three RNG libs (MKL, RNGSSE2, and Random123) are found when attempting to build quinoa, we error out during the cmake run.

We already have CI setup for testing the absence of MKL and I will also set it up testing the absence of RNGSSE2.

Can you please build/test? (On my workstation, I tested this so far, with all of MKL/RNGSSE2/Random123) available using gnu/intel/clang and with only MKL and Random123 being available using clang). If all is good, I will merge this afternoon and update TeamCity for testing the absence of RNGSSE2.

PR #98, opened 2017-01-18
Add -z defs link flag

jbakosi:
Using either `ld` or `gold` I get
```
clang: warning: -Wl,-z,defs: 'linker' input unused
```

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/98)


junghans:
I did:
```
CFLAGS="-Wl,-z,defs" CXXFLAGS="-Wl,-z,defs" cmake ../src/ && make
```

jbakosi:
Okay, than this can be forced as given in de06514. However, I still get
```
clang: warning: -Wl,-z,defs: 'linker' input unused
```
but now at compilation of every object (and a bunch a linking errors).

junghans:
Ok, I was using gcc.

junghans:
`CMAKE_SHARED_LINKER_FLAGS="-Wl,-z,defs"` might work!

jbakosi:
Okay, `CMAKE_SHARED_LINKER_FLAGS` is better. Now I just have to fix the linking errors. ;-)

jbakosi:
Shouldn't we add this flag to only the EXE linker flags? Isn't it enough to make sure that the executables are not underlinked?

junghans:
Executable cannot be underlinked as it is the final resolution of symbols. We should add `-Wl,-z,defs` to the `CMAKE_SHARED_LINKER_FLAGS` that should do it.

src/CMakeLists.txt
```diff
@@ -170,6 +170,10 @@ macro (add_compiler_flag flag)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${flag}")
 endmacro (add_compiler_flag)
 
+# Disallow undefined symbols in object files. Undefined symbols in shared
+# libraries are still allowed.
```
junghans: Hmm, I am a bit confused by the comment, isn't `CMAKE_SHARED_LINKER_FLAGS` used for shared libraries?
And aren't object files just archives and hence have always undefined symbols?

jbakosi: I copied this from ld's man page. According to http://stackoverflow.com/questions/1617286/easy-check-for-unresolved-symbols-in-shared-libraries, passing '-Wl,-z,defs' will just force us to specify all dependent libraries on every library's link line, whereas not passing it just leaves the symbols undefined temporarily and those symbols, of course, will have to be present at the final link.

jbakosi: I don't teally see the point of doing this. I thought '-z,defs' will force us to explicitly have to list all dependent libraries that define symbols required by the libraries we link but don't explicitly link -- as it has to be done, in my experience, for static linking, and so this is would be more prudent that way because all requeid libraries would now be listed explicitly on the link line. But based on the above stackowerflo answer that is not the point of this linker flags?

junghans: Ok, I see. I used something similar here:
http://pkgs.fedoraproject.org/cgit/rpms/gasnet.git/tree/gasnet.spec#n146
but maybe `-z,now` is more what we want.

jbakosi: I think `-z,now` does not attempt to resolve symbols when linking a shared library.

I also found in the [Debian Policy Manual](https://www.debian.org/doc/debian-policy/ch-files.html) , that

> Although not enforced by the build tools, shared libraries must be linked against all libraries that they use symbols from in the same way that binaries are. This ensures the correct functioning of the symbols  and shlibs systems and guarantees that all libraries can be safely opened with dlopen(). Packagers may wish to use the gcc option -Wl,-z,defs when building a shared library. Since this option enforces symbol resolution at build time, a missing library reference will be caught early as a fatal build error.

So I think if we want to be prudent, it is `-z,defs` we want but here is the problem. We now have to explicitly link each shared library to those libraries that are required to resolve all symbols in the library we are building. This means, e.g., we need to link to the Charm++ libraries. I started going down the road of finding out which shared object contains the symbols, but then some Charm++ symbols are defined in several Charm++ libraries, e.g.,

```
<charm++_install_dir>/lib_so:$ find . -name '*.so' -print0 | xargs -0 nm -A | grep CmiOutOfMemory | grep -v U
./libmemory-os-lock.so:0000000000002580 T CmiOutOfMemory
./libmemory-os-lock.so:0000000000002ae0 T CmiOutOfMemoryInit
./libmemory-leak.so:0000000000002a10 T CmiOutOfMemory
./libmemory-leak.so:000000000000a520 T CmiOutOfMemoryInit
./libmemory-default.so:00000000000019b0 T CmiOutOfMemory
./libmemory-default.so:0000000000001b60 T CmiOutOfMemoryInit
./libmemory-charmdebug-slot.so:00000000000040e0 T CmiOutOfMemory
./libmemory-charmdebug-slot.so:000000000000abf0 T CmiOutOfMemoryInit
./libmemory-gnu.so:0000000000002720 T CmiOutOfMemory
./libmemory-gnu.so:0000000000009fe0 T CmiOutOfMemoryInit
./libmemory-charmdebug-mmap-slot.so:0000000000004240 T CmiOutOfMemory
./libmemory-charmdebug-mmap-slot.so:000000000000b0d0 T CmiOutOfMemoryInit
./libmemory-hooks-charmdebug.so:0000000000003700 T CmiOutOfMemory
./libmemory-hooks-charmdebug.so:00000000000066d0 T CmiOutOfMemoryInit
./libmemory-os-verbose.so:0000000000002350 T CmiOutOfMemory
./libmemory-os-verbose.so:00000000000029a0 T CmiOutOfMemoryInit
./libmemory-verbose.so:0000000000002810 T CmiOutOfMemory
./libmemory-verbose.so:000000000000a0f0 T CmiOutOfMemoryInit
./libmemory-os.so:00000000000019b0 T CmiOutOfMemory
./libmemory-os.so:0000000000001b60 T CmiOutOfMemoryInit
./libmemory-os-isomalloc.so:0000000000002710 T CmiOutOfMemory
./libmemory-os-isomalloc.so:0000000000002fb0 T CmiOutOfMemoryInit
./libmemory-os-leak.so:0000000000002690 T CmiOutOfMemory
./libmemory-os-leak.so:0000000000002f30 T CmiOutOfMemoryInit
./libmemory-charmdebug-mmap.so:0000000000004200 T CmiOutOfMemory
./libmemory-charmdebug-mmap.so:000000000000ae60 T CmiOutOfMemoryInit
./libmemory-isomalloc.so:0000000000002a90 T CmiOutOfMemory
./libmemory-isomalloc.so:000000000000a4d0 T CmiOutOfMemoryInit
./libmemory-os-charmdebug.so:0000000000003d90 T CmiOutOfMemory
./libmemory-os-charmdebug.so:0000000000006cc0 T CmiOutOfMemoryInit
./libmemory-os-paranoid.so:00000000000026c0 T CmiOutOfMemory
./libmemory-os-paranoid.so:0000000000003110 T CmiOutOfMemoryInit
./libmemory-charmdebug.so:00000000000040a0 T CmiOutOfMemory
./libmemory-charmdebug.so:000000000000aae0 T CmiOutOfMemoryInit
./libmemory-paranoid.so:0000000000002a60 T CmiOutOfMemory
./libmemory-paranoid.so:000000000000a6e0 T CmiOutOfMemoryInit
```

So `charmc` must pick the correct combination of the libraries to link depending potentially complicated logic. We certainly do not want to re-create and maintain that logic ourselves. I also looked at `charmc` which does have a complex logic for generating the correct link line. So probably we would want to link not only the final executables with `charmc` but also the shared libraries, however, I'm not sure how to do that. Switching the compiler only for linking the executables are done via

```
STRING(REGEX REPLACE "<CMAKE_CXX_COMPILER>"
       "${CHARM_COMPILER} -module DistributedLB"
       CMAKE_CXX_LINK_EXECUTABLE "${CMAKE_CXX_LINK_EXECUTABLE}")
```

so I did the analogous thing for creating shared libraries with

```
STRING(REGEX REPLACE "<CMAKE_CXX_COMPILER>"
       "${CHARM_COMPILER}"
       CMAKE_CXX_CREATE_SHARED_LIBRARY ${CMAKE_CXX_CREATE_SHARED_LIBRARY}")
message("S: ${CMAKE_CXX_CREATE_SHARED_LIBRARY}")
```

However, that still does not resolve all Charm++ symbols, and I don't understand why. Any thoughts?

junghans: I saw the same, we could do:
```
set(CMAKE_LINKER ${CHARM_COMPILER} -module DistributedLB)
```
I guess.

jbakosi:
Not a high priority, will continue some other time.

PR #99, opened 2017-01-18
Make Testu01 optional

jbakosi:
As a follow-up to https://github.com/quinoacomputing/quinoa/pull/90#issuecomment-273347288.

### DIFF
--- a/cmake/FindRNGSSE2.cmake
+++ b/cmake/FindRNGSSE2.cmake
@@ -4,7 +4,7 @@
 # \author    J. Bakosi
 # \copyright 2012-2015, Jozsef Bakosi, 2016, Los Alamos National Security, LLC.
 # \brief     Find the RNGSSE2 library
-# \date      Fri 13 Jan 2017 03:44:47 PM MST
+# \date      Wed 18 Jan 2017 02:31:55 PM MST
 #
 ################################################################################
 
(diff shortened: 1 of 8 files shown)


junghans:
When testing I got:
```
cc1plus: warning: /home/junghans/computing/quinoa/tpl/install/gnu-x86_64/include: No such file or directory [-Wmissing-include-dirs]
cc1plus: warning: /home/junghans/computing/quinoa/build/RNGTest: No such file or directory [-Wmissing-include-dirs]
```

jbakosi:
The first one is probably because the cmake variable TPL_DIR is now undefined in `src/CMakeLists.txt'. I guess we should maybe detect the existence of the library around line 79 and exclude it from the `include_directories` below, at line 495.

jbakosi:
I'm not so sure about the second warning.

jbakosi:
Oh, I guess the 2nd warning is because ` ${PROJECT_BINARY_DIR}/RNGTest` is now undefined at line 501 of `src/CMakeListst.txt`. I guess the same strategy would work there too.

jbakosi:
Also, I'm updating my TeamCity build agents with all TPLs parked in some directory which each can be pulled in via environment variables. This will speed up the builds but also will enable testing the presence and absence of system-wide TPLs. As a result, I should have a situation where all TPLs are found on the system so I can test the issues you bumped into above.

junghans:
I will look at the warnings in different pull request.

jbakosi:
The warnings in https://github.com/quinoacomputing/quinoa/pull/99#issuecomment-273626584 are fixed by (fd5ef4b3bb6f9e449bc0ccbca6424cf0f4a04493, cbf9cd69cb965998ef0067db179402ac5c8c103b) and 8101ae4dc1c, respectively.

PR #100, opened 2017-01-19
Find TPLs

jbakosi:
This is a set of changes that I encountered and needed to be fixed while I was setting up system-wide libraries on our TeamCity build agents, also mentioned in https://github.com/quinoacomputing/quinoa/pull/99#issuecomment-273637095. I have also managed to put in a number of small fixes in the meantime. @junghans: can you please build/test?


cmake/FindZoltan2.cmake
```diff
@@ -33,22 +33,19 @@ endif()
...
 else()
   FIND_LIBRARY(Zoltan2_LIBRARY NAMES zoltan2 trilinos_zoltan2
                PATH_SUFFIXES trilinos
-               HINTS ${Zoltan2_ROOT}/lib)
+               HINTS ${Zoltan2_ROOT}/lib $ENV{ZOLTAN2_ROOT}/lib)
 endif()
 
-set(Zoltan2_LIBRARIES ${Zoltan2_LIBRARY} )
-set(Zoltan2_INCLUDE_DIRS ${Zoltan2_INCLUDE_DIR} )
-
```
junghans: I followed this:
https://cmake.org/Wiki/CMake:How_To_Find_Libraries#Writing_find_modules

jbakosi: Okay, that's cool. We can put that back.

junghans:
I think we should use the chance to:

- [ ] Consistent spelling (Capital vs Camel)
- [ ] Consistent naming (plural vs singular)
- [ ] Consistent hints (environment vs. cmake variable or both)

cmake/FindZoltan2.cmake
junghans: I think we should just make it consistent.

jbakosi:
I agree, we need to make these consistent across all TPLs.

cmake/FindZoltan2.cmake
junghans: One _hidden_ advantage of these lines is that, these ones can be overwritten in the main build system, while the ones from `find_library()` and `find_path()` cannot (with `FORCE`) as they are cached.

jbakosi:
Okay, can you please take a look at _only_ https://github.com/quinoacomputing/quinoa/pull/100/commits/0c831a5c8829f2ffb15a32f1ddfcf914c7299713, and see if I managed to make that consistent wrt spelling, naming, etc.? Then I'll do the rest...

jbakosi:
On a related note, there is another problem with Zoltan2 and probably all trilinos packages:

Since now we have our own `FindZoltan2.cmake` we only find the Zoltan2 library but not its dependencies (within trilinos). However, previously, since we always built trilinos, we just pointed to `Zoltan2Config.cmake` (by doing `SET(CMAKE_PREFIX_PATH ${TPL_DIR} ${CMAKE_PREFIX_PATH})`) installed into `TPL_DIR/clang-x86_64/lib/cmake/Zoltan2` by the trilinos install. Putting only libzoltan to the link line is okay with dynamic linking because the linker pulls in the dependencies, but not with static, which requires the explicit list of all dependent libraries too.

For example, this is the output from our `FindZoltan2.cmake`:
```
-- Found Zoltan2: /home/jbakosi/code/quinoa/tpl/install/clang-x86_64/lib/libzoltan2.so
```
and this is the output from the trilinos-provided `Zoltan2Config.cmake`:
```
-- Found Zoltan2: zoltan2;xpetra-sup;xpetra;tpetraext;tpetrainout;tpetra;tpetrakernels;tpetraclassiclinalg;tpetraclassicnodeapi;tpetraclassic;zoltan;teuchoskokkoscomm;teuchoskokkoscompat;teuchosremainder;teuchosnumerics;teuchoscomm;teuchosparameterlist;teuchoscore;kokkosalgorithms;kokkoscontainers;kokkoscore
```

One can imagine the slew of undefined symbols if those are not given during static linking. BTW, this problem was triggered by our Alpine linux docker container build (and only that) that is run during CI.

This is kind of a chicken and the egg problem: we should use the `Zoltan2Config.cmake` provided by trilinos to rely on the trilinos folks to find the dependencies for us. However, this is not available unless trilinos is installed. So we can't just find a trilinos library without knowing the location of a trilinos install.

I don't think we should write our own FindZoltan2.cmake, but instead get a directory where trilinos is installed, e.g., `TRILINOS_INSTALL`. Then in `${TRILINOS_INSTALL}/lib/cmake` should be added to the list in `CMAKE_PREFIX_PATH`, which then will find `Zoltan2Config.cmake` which provides the ability to call `find_package(Zoltan2)`. The same for any other trilinos packages.

Does this make sense?

junghans:
Absolutely makes sense! Let's drop our `FindZoltan2.cmake` and use `find_package(Zoltan2 QUIETLY)` in the TPL build, so the user doesn't get overwhelmed with the long error message that `Zoltan2.cmake` wasn't found.

junghans:
And Gentoo installs the cmake config files of Trillions as well.

jbakosi:
Sounds good. Cmake searches on system paths for `<package>Config.cmake` files and it does find these files on Debian if Trilinos is installed system-wide, so this will work nice. On system where trilinos is installed in a custom location, `-D<package>_DIR=<trilinos_install>/lib/cmake/<package>` enables finding the packages too - I just tested this.

jbakosi:
And `-D<package>_DIR=<trilinos_install>/lib/cmake/<package>` can also be set as environment variables, e.g., by environment modules.

junghans:
`Zoltan2-dev` puts `Zoltan2.cmake` in `/usr/lib/x86_64-linux-gnu/cmake/Zoltan`, so we should add that directory to `CMAKE_PREFIX_PATH` if we build on darwin.

jbakosi:
That directory is not searched by cmake automatically?

junghans:
This seems similar:
http://askubuntu.com/questions/633826/default-installation-prefix-qt5-qt5-widgets-ubuntu
I guess the problem is the `Zoltan2` subdir.

jbakosi:
What do you think about the procedure outlined in the commit message https://github.com/quinoacomputing/quinoa/pull/100/commits/2558b6e0a34b3216cafda5c1c4ec141429cd398a?

cmake/TPLs.cmake
```diff
@@ -110,17 +112,12 @@ if(NOT BUILD_SHARED_LIBS)
...
-set(ZOLTAN2_ROOT ${TPL_DIR}) # prefer ours
 find_package(Zoltan2 REQUIRED)
 
 #### ExodusII library
-set(SEACASEXODUS_ROOT ${TPL_DIR}) # prefer ours
 find_package(SEACASExodus REQUIRED)
 find_package(SEACASExodiff REQUIRED)
 
-#### H5Part library
-find_package(H5Part REQUIRED)
```
junghans: Why did you remove that?

cmake/FindSEACASExodus.cmake
```diff
@@ -1,32 +0,0 @@
-#------------------------------------------------------------------------------#
```
junghans: I would keep that as exodus will be provided by exodusii-devel on debian, which doesn't have a cmake file.

cmake/TPLs.cmake
jbakosi: Because it was there twice - probably a result of a bad merge.

jbakosi:
No I just don't understand why they did _not_ include any cmake variable in `SEACASExodiffConfig.cmake` that would define the `exodiff` executable. So for that we still need a `FindExodiff.cmake`.

junghans:
I was thinking about it again, even with libexodus, we still need `exodiff` and hence exodus from trilinos, so let's drop it.

jbakosi:
Yeah, I'm pulling back FindSEACASExodiff.cmake, probably with a different name. Argh!

jbakosi:
Alright, I have attempted to make all `Find<package>.cmake` files conforming to the same style and also to KitWare recommendations and I have also updated the Zoltan2/Exodus/Exodiff stuff. I woud like to be done with this PR. Could you please review? I realize the changes here can really be tested for correctness when we subject to a bunch of different kinds of builds for which the TeamCity build matrices are partly already set up. I have only done some basic manual sanity checking during these updates on two different machines, one that has some TPLs missing on the system and thus we build some, and another one that finds every TPLs.

So @junghans can you please take a quick look and perhaps build & test on gentoo, then I'll merge and get on to the more thorough TeamCity build matrices (unfortunately, I can only do this latter on the master branch due to a TeamCity plugin configuration issue.)

junghans:
With RNGSSE2 I get:
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
RNGSSE2_LIBRARY
    linked by target "inciter" in directory /home/junghans/computing/quinoa/src/Main
    linked by target "walker" in directory /home/junghans/computing/quinoa/src/Main
    linked by target "unittest" in directory /home/junghans/computing/quinoa/src/Main

-- Configuring incomplete, errors occurred!
```

junghans:
Fixed

jbakosi:
Yes, I missed that. Thanks. I'm sure there will be some others too. Also, I have not tested with the any of the optional libs not being there, e.g. `TestU01`, but is what I want to do with a build matrix that I have already setup.

junghans:
Building it now, found (#101), but that is trilinos.

junghans:
And merge #103 if you like.

Issue #102, opened 2017-01-20
library names are a bit too generic


junghans:
quinoa libaries names, e.g. libMesh are very generic and will certainly conflict with other libraries when installed in `/usr/lib`.

junghans:
I suggest `libXXX` to `libquinoa_XXX`, we can do that by:
```
set_target_properties(XXX PROPERTIES LIBRARY_OUTPUT_NAME libquinoa_XXX)
```

PR #106, opened 2017-01-23
Specialize libnames

jbakosi:
As a response to #102. Also reduced some duplicated cmake code.


junghans:
LGTM. Will test it tomorrow!

jbakosi:
Cool, I'll merge it because even if I screwed up, it is easy to fix and we have no users. ;-)

PR #108, opened 2017-02-27
Adding travis support

junghans:


junghans:
@jbakosi, things to fix:
- [ ] Clang supports coverage, why is cmake not liking it?
- [ ] Using charmc for linking yields:
```
x86_64-pc-linux-gnu-g++: error: unrecognized command line option '-fcolor-diagnostics'
86_64-pc-linux-gnu-g++: error: unrecognized command line option '-Weverything'
```
- [ ] `0% tests passed, 95 tests failed out of 95`, huh?
- [ ] gcc flags are always `-O0 --coverage`

jbakosi:
> - [ ] Clang supports coverage, why is cmake not liking it?
> - [ ] gcc flags are always `-O0 --coverage`

I have never tried generating code coverage with clang (as never had a use for it), but I'm not against it seting it up. With the help of http://logan.tw/posts/2015/04/28/check-code-coverage-with-clang-and-lcov/, 0333eb1bc92 enables coverage for clang which now works, but with same caveats. See the commit message for details.

jbakosi:
```
x86_64-pc-linux-gnu-g++: error: unrecognized command line option '-fcolor-diagnostics'
86_64-pc-linux-gnu-g++: error: unrecognized command line option '-Weverything'
```
These are clang-only arguments, guarded by, in `src/CMakeLists.txt`:
```
if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")  # clang-specific settings
...
```
Why does it get added to gcc builds? Does the above matches "Clang"?

jbakosi:
> - [ ] `0% tests passed, 95 tests failed out of 95`, huh?

I don't know why the regression test fail. Try to run a single test with, e.g.,
```
ctest -R walker:Dirichlet_pe1
```
 and see what fails during that single test.

cmake/DetectCodeCoverage.cmake
```diff
@@ -20,7 +26,8 @@ if ( GCOV AND
      LCOV AND
```
junghans: @jbakosi : I think the logic here is wrong, whenever `gcov`, `lcov`, `genhtml` and `sed` are found, coverage is enabled. It seems to me this block should be wrapped in some like `if(ENABLE_COVERRAGE)`.

docker/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+USER root
+RUN chown -R quinoa:quinoa /home/quinoa/quinoa /home/quinoa/.ccache
+USER quinoa
+
+WORKDIR /home/quinoa/quinoa
+RUN mkdir build
+
+WORKDIR build
+RUN ccache -z
+RUN cmake ${COVERAGE:+-DCODE_COVERAGE=ON} -DLCOV=/bin/true -DGENHTML=/bin/true ../src
```
junghans: @jbakosi : `${COVERAGE:+-DCODE_COVERAGE=ON}` means only add `-DCODE_COVERAGE=ON` to cmake if `COVERAGE` is set in the environment.

cmake/DetectCodeCoverage.cmake
junghans: Plus for coverage on travis, we don't need `lcov` and `genhtml`.

junghans:
For the warning is problem is that charm got build with gcc and hence charmc is using gcc, but quinoa is build with clang and injects '-fcolor-diagnostics' and `-Weverything`. Maybe we can inject there flags only in the compile line, but not the link command.

cmake/DetectCodeCoverage.cmake
jbakosi: Yeah that would work too, but the rationale behind the current logic is intended to be: if any of those tools are found and the compiler supports coverage, we enable code coverage. Does this make sense?

jbakosi: Note that even if code coverage is enabled (by the cmake variable `CODE_COVERAGE = on` ) that still does not mean the code coverage targets, .e.g, `unittest_coverage` will be built. (By default they will not be built.)

docker/Dockerfile
jbakosi: In light of the above, that is not how this is intended to be used.

jbakosi:
>For the warning is problem is that charm got build with gcc and hence charmc is using gcc, but quinoa is  build with clang and injects '-fcolor-diagnostics' and -Weverything. Maybe we can inject there flags only in the compile line, but not the link command.

I have not so far mixed TPL and code compilers -- I only used the same for both. Is that important to be able to do? I guess for charm, the answer is yes, as that provides a wrapper around whatever compiler it was used to be built with.

cmake/DetectCodeCoverage.cmake
jbakosi: > Plus for coverage on travis, we don't need lcov and genhtml.

That how would that generate coverage?

junghans:
`charmc` has a `-cc` and `-c++` option to overwrite the compilers it was built with.

cmake/DetectCodeCoverage.cmake
jbakosi: More bg info on the logic: Currently we have three code coverage targets: `unittest_coverage`, `regression_coverage`, and `test_coverage`, where the last one is the coverage for both. (I wanted to see detailed (line-)coverage depending on the type of tests.)

jbakosi:
> `charmc` has a `-cc` and `-c++` option to overwrite the compilers it was built with.

Okay, I did not know that, so I guess that wold enable using different compilers for charm and code.

cmake/DetectCodeCoverage.cmake
jbakosi: The three blocks starting at https://github.com/quinoacomputing/quinoa/blob/master/script/update_doc.sh#L59
shows how we generate coverage currently.

junghans: I am not sure if it is a good idea to always in inject `-O0` if these tools are found, as it makes the code slow.

I use Travis with [codecov](codecov.io), which just a parses all the gcov output files.

jbakosi: With https://github.com/quinoacomputing/quinoa/pull/108/commits/ac5d540031ef3c519c756af5ca1772158141bcfe we no longer inject `-O0`.

jbakosi: > I use Travis with [codecov](codecov.io), which just a parses all the gcov output files.

Cool! Will codecov.io work with Charm++ stuff and can it filter out unwanted files, such as https://github.com/quinoacomputing/quinoa/blob/master/cmake/CodeCoverage.cmake#L80 does?

docker/Dockerfile
junghans: Ok, we could use `${COVERAGE:+-DLCOV=/bin/true}` instead.

cmake/DetectCodeCoverage.cmake
junghans: Yes, `codecov` has an exclusion list, too.

docker/Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+RUN make VERBOSE=1 -j2
+RUN ccache -s
+RUN ../script/run_tests.sh
+RUN make install DESTDIR=${PWD}
+RUN cd .. && if [ ${COVERAGE} ]; then \
+  if [ ${CC} = clang ]; then \
+    $HOME/.local/bin/codecov --gcov-exec "llvm-cov gcov"; \
```
junghans: @jbakosi: that is where we would add the exclusion option. Or we put this command inside cmake.

junghans:
Ok, I fixed most of the issues.

@jbakosi, can you defined a smaller set of tests to be run by travis, e.g. `script/run_ci_tests.sh`.

jbakosi:
The best would be to run all tests. I see travis got stuck with `walker:SkewNormalPDF_pe8`. You can label that STRINGENT in https://github.com/quinoacomputing/quinoa/blob/master/regression/walker/SkewNormal/CMakeLists.txt
as is done similarly in https://github.com/quinoacomputing/quinoa/blob/master/regression/rngtest/CMakeLists.txt#L9 and https://github.com/quinoacomputing/quinoa/blob/master/script/run_tests.sh will skip those (the stringent tests run nightly anyway).

junghans:
Travis has two restrictions, max build time need to be less than  60 mins and there needs to be output every 10 mins. The latter currently makes the test fail.

jbakosi:
> The latter currently makes the test fail.

You can fix that by labeling some tests `STRINGENT`.

jbakosi:
We can also add a new regression test label, e.g., `TRAVIS`, and make the https://github.com/quinoacomputing/quinoa/blob/master/script/run_tests.sh take the `-LE` argument.

docker/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN cmake \
     ../src
 RUN make VERBOSE=1 -j2
 RUN ccache -s
-RUN ../script/run_tests.sh
```
jbakosi: We should still run the unit tests:
```
RUN ./charmrun +p2 Main/unittest -v
```

docker/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN cmake \
     ../src
 RUN make VERBOSE=1 -j2
 RUN ccache -s
-RUN ctest -E "walker:SkewNormalPDF_pe8"
```
jbakosi: I really like this solution to exclude tests only for travis, instead of labeling them in any way. This way the exclusion only affects travis and does not exlude tests in any other venue. Awesome!

junghans: Plus, new tests will be picked up automatically.

jbakosi:
The default build type is `Debug`. However, `Release` will run the tests about 3-4x faster. I think it may be enough to test `Release`. (I'm testing `Debug` using TeamCity.) That way all tests might finish within an hour.

docker/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN cmake \
     ../src
 RUN make VERBOSE=1 -j2
 RUN ccache -s
-RUN ctest -E "(walker:SkewNormalPDF_pe8|walker:DiagOrnsteinUhlenbeckPDF_pe8)"
```
jbakosi: I would do `Release` build instead and not exclude these. See https://github.com/quinoacomputing/quinoa/pull/108#issuecomment-283411905.

docker/Dockerfile
```diff
@@ -36,7 +36,7 @@ RUN cmake \
...
+RUN ctest -E "_pe[45678]"
```
jbakosi: I would do `Release` build instead and not exclude these. See https://github.com/quinoacomputing/quinoa/pull/108#issuecomment-283411905.

junghans: I am still playing with how much we can get done in 50 mins on Travis.

jbakosi: BTW, the `_pe*` tests attempt to brute-force test asynchronous logic: the runtime system will schedule some calls differently depending on the load of the machine the tests are running on. So if they fail sometimes and apparently in a random fashion or some tests stall forever, that's an indication of an asynchronous logic bug, which can be pretty hard to debug. Bottom line: it is nice to include as many of such tests.

jbakosi: Okay, but I still think, `Release` is sufficient on travis considering time limitations there.

docker/Dockerfile
```diff
@@ -0,0 +1,51 @@
+FROM quinoacomputing/buildenv:gentoo
```
jbakosi: Can you rename this file to differentiate better from the other docker files in that directory? E.g., `Dockerfile.quinoa-build-travis` or something like that.

junghans: Sure!

jbakosi:
Okay, thanks. Let's see how this does then I'll merge to master.

PR #110, opened 2017-03-08
Add -Werror build

junghans:


jbakosi:
Okay, I think this one will pass. I guess we want
```
env:
  matrix:
    - DISTRO=gentoo DOCKERHUB=true WERROR=ON
    - DISTRO=gentoo COVERAGE=ON
```
instead of
```
matrix:
      - DISTRO=gentoo DOCKERHUB=true
￼      - DISTRO=gentoo COVERAGE=ON
      - DISTRO=gentoo WERROR=ON
```
Correct? If so, @junghans, you can also merge.

PR #132, opened 2017-04-20
Re-enable Intel TeamCity builds

jbakosi:
This now works, but I wanted us to see the various steps that were required to accomplish this. In particular, the commit messages have some interesting details that you guys might also be interested in (if for nothing else, but other projects).

All (now) 80 TeamCity builds in the Linux matrix are also successful.


junghans:
@jbakosi You can merge this!

PR #133, opened 2017-04-20
drop \date

junghans:
git grep --files-with-matches '\\date' | xargs sed -i '/\\date/d'


junghans:
Hmm, I don't really feel like going through >400 files to decide this, so let's just drop `\date`, which is anyhow the main cause of merge conflict.

jbakosi:
I'm fine with just dropping '\date' too, but I don't think you would have to do the differentiation between the '\author's in the functions and the '\author's in the file header manually. You could just use 'head' to selectively grep and delete those '\author' lines that appear within the first, e.g., 5-6 lines of each file. (That is also how vim's 'autodate' plugin works to automatically update the dates.)

PR #135, opened 2017-05-03
Updated travis notification email

junghans:
$ travis encrypt "abc@def.com" --add notifications.email.recipients


jbakosi:
Next we against develop. Thanks.

PR #136, opened 2017-05-12
fix endian.h collision with glibc

junghans:
glibc comes with an endian.h as well. And on case insensitve file-
systems this might ge picked up, so rename ours.


junghans:
Done

PR #140, opened 2017-05-24
Update to PEGTL 2.0

jbakosi:
In conjunction with https://github.com/quinoacomputing/quinoa-tpl/pull/10.

May also require updates to the docker images: ubuntu, ubuntu-verde, and gentoo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Currently, Quinoa consists of the following tools:
 
    ```
    apt-get install cmake gfortran gcc g++ openmpi-bin libopenmpi-dev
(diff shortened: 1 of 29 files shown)


jbakosi:
@junghans: The gentoo docker image needs a rebuild (I believe without any change to the docker config file), but I'm having problems with rsync proxy at this time:
```
!!! getaddrinfo failed for 'rsync.gentoo.org': [Errno -3] Temporary failure in name resolution
```

jbakosi:
Gentoo installs exodus but does it install exodiff? See https://travis-ci.org/quinoacomputing/quinoa/jobs/236363849

junghans:
I guess that is due to gentoo/sci@a8ec95d3c4789c41491d775365b9a4cdbb6bf511!

junghans:
Gentoo builds, alpine docker image needs an update though.

jbakosi:
I think alpine will require a merge to master on the tpl and the cmake repos because it uses master.

PR #142, opened 2017-06-01
re-enable clang builds

junghans:


jbakosi:
Actually, I have encountered and have fixed this error before by updating to clang 4.0.1 on the TC build agents! Can gentoo do that?

PR #141, opened 2017-05-30
ASC mesh reader

jbakosi:
Rudimentary ASC mesh reader to facilitate reading meshes used by Jacob Waltz's Chicoma code.

PR #145, opened 2017-06-07
Vort

jbakosi:
The name of this branch is also now somewhat of a misnomer, sorry. These are changes toward being able to run the various manufactured solution cases we are working on. However, even the vortical flow test case does not seem to be correct at this point, so that's next on my list. The reason for creating this PR is not to grow this larger than what it is now because then it will be even harder to review.

Changes at a high level, please see the individual commit messages for more details:

* Added (doxygen) documentation on three new inciter test examples under `doc/examples`, the input and meshes are now also committed to a new [examples](https://github.com/quinoacomputing/quinoa-examples) repository.

* A "new" regression test, `asc2exo`, and related files are really from an earlier merge that was already reviewed, so please just ignore this showing up in the diff for `regression/meshconv/exo_output/CMakeLists.txt` and associated files in the same directory. Same for `regression/meshconv/gmsh_output/CMakeLists.txt` and `regression/meshconv/netgen_output/CMakeLists.txt`. Also, everything related to `ASCMeshReader` has been reviewed in [that previous PR](https://github.com/quinoacomputing/quinoa/pull/141), so just please ignore everything ASC mesh reader related (tests and code). (Background: ASC is a new simple mesh format that we can now use for mesh convergence studies and I have access to the nice regular meshes in Jacob's paper. We use meshconv to convert to exodus and run them in inciter.)

* Added user-configurable output for error norms (with two types of errors, L_2 and L_inf). This is really the majority of the code in this PR. See more details in the commit message for https://github.com/quinoacomputing/quinoa/pull/145/commits/91821d3edfc7bc938094abcc2c4636297c727c02.

* I also fixed a nasty bug in https://github.com/quinoacomputing/quinoa/pull/145/commits/40abfa98076a787ce708dfe93295637727b787e4, regarding the computation of errors.


jbakosi:
Great! Thanks Francisco, I greatly appreciate your time.

The vortical flow, nonlinear energy growth, and the Rayleigh-Taylor test problems will now come pretty soon as well, since it appears I have just found and fixed the last problem with them.

Thanks again.

Issue #131, opened 2017-04-19
Set up a CI build matrix entry using Charm++'s Randomized Message Queueing


jbakosi:
See how to build Charm++ with [Randomized Message Queueing](http://charm.cs.illinois.edu/manuals/html/charm++/F.html#SECTION05610000000000000000).

jbakosi:
Additional info from @stwhite91:
>... here's how to actually build with randomized message queues:
```
   ./build charm++ netlrts-linux-x86_64 --with-prio-type=int --enable-randomized-msgq --suffix randq-debug -g
```
> Use of randomized message queues requires the message priority type to not be a bitvector (the default), and the `--suffix` option will be concatenated to the build directory, so this build directory will be named `netlrts-linux-x86_64-randq-debug/`.

jbakosi:
Will do:
- [ ] travis build + unit, & regression test suite run with gnu/release
- [ ] teamcity build + unit, & regression test suite run with gnu/release, with rngsse2, testu01, mkl (turning these three on ensures the most tests)

jbakosi:
https://github.com/quinoacomputing/quinoa-tpl/commit/3802cb471c5ee630f946228b69908c447d6a29e7 enables building Charm++ with randomized queue, using:
```
cmake -DCHARM_EXTRA_ARGS="--with-prio-type=int;--enable-randomized-msgq;--suffix;randq-debug" <path-to-CMakeLists.txt>
```

jbakosi:
@stwhite91: Does randomized queue work (and makes sense) with optimized builds?

stwhite91:
Yes it works and makes sense. It's just a different data structure than the usual FIFO queue with priorities.

jbakosi:
Progress:

Using Clang 4.0.1, libc++, OpenMPI 2.0.2, and Cham++ with randomized message queue:
- [x] `ctest -j36 -R meshconv`
- [x] `ctest -j36 -R rngtest -LE extreme`
- [x]  `ctest -j36 -R walker -E PDF` (skipping those estimating PDFs)
- [ ] `ctest -j36 -R walker`: estimating PDFs fail with
```
------------- Processor 0 Exiting: Called CmiAbort ------------
Reason: mis-matched client callbacks in reduction messages
```

jbakosi:
[...]

PR #146, opened 2017-06-22
Duplicate linksys signal fix

rspavel:
Requesting Code Review and approval of merge of bugfix in LinksysMerger to develop branch. Please follow standard code review practices as there is a good chance I missed something and I don't want to introduce a different error.

During analysis of LinksysMerger it was noticed that the same signal was called twice, sequentially. Upon further analysis it was clear that this was due to multiple when() statements dependent on the signal. I have yet to check the Charm++ model to see if this is undefined behavior, but it is at least ambiguous and decreases code clarity. I changed the duplicate rhsbc_complete() signals to be unique for each when() statement.

### DIFF
--- a/src/LinSys/LinSysMerger.h
+++ b/src/LinSys/LinSysMerger.h
@@ -1,7 +1,7 @@
 // *****************************************************************************
 /*!
   \file      src/LinSys/LinSysMerger.h
-  \author    J. Bakosi
+  \author    Los Alamos National Laboratory
   \copyright 2012-2015, Jozsef Bakosi, 2016, Los Alamos National Security, LLC.
   \brief     Charm++ chare linear system merger group to solve a linear system
   \details   Charm++ chare linear system merger group used to collect and
@@ -218,7 +218,7 @@ extern CkReduction::reducerType DiagMerger;
 //!   group's elements are used to collect information from all chare objects
 //!   that happen to be on a given PE. See also the Charm++ interface file
 //!   linsysmerger.ci.
-//! \author J. Bakosi
+//! \author Los Alamos National Laboratory
 template< class HostProxy, class WorkerProxy, class AuxSolver  >
 class LinSysMerger : public CBase_LinSysMerger< HostProxy,
                                                 WorkerProxy,
--- a/src/LinSys/LinSysMerger.h
+++ b/src/LinSys/LinSysMerger.h
@@ -1003,7 +1003,8 @@ class LinSysMerger : public CBase_LinSysMerger< HostProxy,
               r[i] = 0.0;
         }
       }
-      rhsbc_complete(); rhsbc_complete();
+      rhsbc_complete_rhs();
+      rhsbc_complete_aux();
     }
 
     //! Build Hypre data for our portion of the solution vector
--- a/src/LinSys/linsysmerger.ci
+++ b/src/LinSys/linsysmerger.ci
@@ -1,7 +1,7 @@
 // *****************************************************************************
 /*!
   \file      src/LinSys/linsysmerger.ci
-  \author    J. Bakosi
+  \author    Los Alamos National Laboratory
   \copyright 2012-2015, Jozsef Bakosi, 2016, Los Alamos National Security, LLC.
   \brief     Charm++ module interface file for merging a linear system
   \details   Charm++ module interface file for merging a linear system. See more
@@ -200,7 +200,7 @@ module linsysmerger {
       entry void wait4lhs() {
         when lhsbc_complete() serial "hyprelhs" { hyprelhs(); } };
       entry void wait4rhs() {
-        when rhsbc_complete() serial "hyprerhs" { hyprerhs(); } };
+        when rhsbc_complete_rhs() serial "hyprerhs" { hyprerhs(); } };
 
       entry void wait4hypresol() {
         when hypresol_complete(), hyprerow_complete() serial "sol" { sol(); } };
@@ -220,7 +220,7 @@ module linsysmerger {
         when asmsol_complete(), asmlhs_complete(), asmrhs_complete()
           serial "solve" { solve(); } };
       entry void wait4aux() {
-        when rhsbc_complete(), auxlhs_complete(), auxrhs_complete()
+        when rhsbc_complete_aux(), auxlhs_complete(), auxrhs_complete()
           serial "aux" { auxsolve(); } };
 
       entry void wait4solve() {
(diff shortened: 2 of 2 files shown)


src/LinSys/linsysmerger.ci
```diff
@@ -200,7 +200,7 @@ module linsysmerger {
       entry void wait4lhs() {
         when lhsbc_complete() serial "hyprelhs" { hyprelhs(); } };
       entry void wait4rhs() {
-        when rhsbc_complete() serial "hyprerhs" { hyprerhs(); } };
+        when rhsbc_complete_rhs() serial "hyprerhs" { hyprerhs(); } };
```
jbakosi: This is very good. I did not even thought about naming these differently, which does make the client code more readable. Thanks for this, this is awesome! There might be other duplicates too.

BTW: The reason behind these duplicate signals is my (most likely limited) understanding is that these behave like messages in a queue so when they can be consumed by multiple `when` SDAG statements, there must be multiple "messages" in the queue. If not, the incorrect asynchronous logic happens which can lead to errors, incorrect results, and stalling. More info about this is in the commit message of 5663ed146b2d12bac1654ca5fc53b4767e003622, in the paragraph starting with _This commit also revamps the asynchronous logic in LinSysMerger_. In particular, please see also my discussion on the Charm++ mailing list on this topic at https://lists.cs.illinois.edu/lists/arc/charm/2016-09/msg00002.html. The correct way of using SDAG triggers in a multiple fashion is certainly still a little bit still unclear to me.

In summary, the change does make the code cleaner, and I think the asynchronous logic stays correct. I also welcome any more insight on how to understand (think about) this issue better.

src/LinSys/LinSysMerger.h
```diff
@@ -1,7 +1,7 @@
 // *****************************************************************************
 /*!
   \file      src/LinSys/LinSysMerger.h
-  \author    J. Bakosi
+  \author    Los Alamos National Laboratory
```
rspavel: As this project becomes more complex and has more developers, having a giant list of authors on every function becomes unwieldy. 

Also, the functionality you are talking about is provided by git (and can be used with the github interface) https://git-scm.com/docs/git-blame

jbakosi: I like the _original_ author (not a list of authors) showing up directly in the source file as well as in the doxygen documentation page. That does not even require a `git blame`. (Some folks do not even know about that git command.)

The rationale for the _original_ author (and not a list) is that that person is most likely to be the one familiar with the design of (and the need for) a given file. `git blame` shows who edited every single line in the file -- a very different, and in some cases way too much, information.

rspavel: We might want to take this offline, but I disagree that the original author will have the best understanding of why a code block written by twelve different people does what it does. This is why when you look at the Linux kernel you don't immediately send Linus an email about the methodology behind gamepad support.

Similarly, I make a lot of mistakes when I code (that is why I like code reviews and regression tests. I can force someone to double check me). git-blame is the tool to use when we find out that I added an off by one error under certain cases two years ago. The original author has no idea that I forgot what language I was in and might think it was intentional due to weird interloop dependencies. That is what git-blame is for and adding additional information just leads to delays if the wrong person is asked.

And, speaking more philosophically: Open source codes are a community project and can have any number of authors. I would prefer no author blocks at all because maybe down the line someone from a random University will be using quinoa and contribute a feature or a bug fix that some random postdoc who drew the short straw and got code maintenance responsibilities approved.
Github is nice because it provides a great list of all contributors and git-blame is great because it lets you figure out which of those were naughty when you spend a few hours debugging and nail down a problematic statement. And the original author may move on to bigger and better things and it might be some random person who is now in charge of code maintenance and approval of pull requests.

While more about copyrights (which are a necessary "evil" when you have open source projects that are owned and financially supported by an entity), https://opensource.com/law/14/2/copyright-statements-source-files provides a moderately well written argument on the subject.

jbakosi: These are good points, including the blog post. However, adding an author field as "Los Alamos National Laboratory" has completely zero information content.

rspavel: Agreed. LANL as an author block is pretty not useful. As per this discussion I have fixed with PR 147 and the future ones in the cmake and TPL repos

src/LinSys/linsysmerger.ci
jbakosi: Ah! I think there is something we are missing with the way we understand multiple use of SDAG triggers/messages.

I was just able to produce an example, that does _not_ work if you use unique names for the triggers, but works with using the same names.

@rspavel: can you please verify that you can reproduce the following?

Check out 715b8a82fe3081bcdbea64a036be0be25ecc16db, compile and run
```
ctest -j36 -R asynclogic_u0.5_pe8 -V
```
It should run fine.

Then check out a156faa5cc60ba26354a28a96055026b39bd46b4, compile and run the same, which hands for me.

The only difference is that a156faa5cc60ba26354a28a96055026b39bd46b4 uses uniquely named SDAG triggers.

Note that the logic requires `lhs_complete()` _and_ the completion of LinSysMerger::addbc(). I believe both versions should work, but for some reason only the version using the _non-uniquely-named_ SDAG triggers works.

This is beyond my understanding at this point...

I have no illusions that the asynchronous SDAG logic in LinSysMerger is correct, because the regression test suite fails with a randomized-queue-build of Charm++. That is why we need to get #131 working and in the meantime not screw up much.

Because of this, out of caution, I am reverting this to not-approved. Sorry, I really liked the uniquely-named triggers...

rspavel: So just to make sure I understand: The alterations in this PR are not problematic, but alterations in the same vein in a different portion of the code might be?

I can look into it (I suspect it may be related to quinoa taking radically different code paths based upon preprocessor flags), but I would suggest either accepting this as individually correct (as it seems to be) or changing the nature of the PR to indicate that its scope has drastically changed to an entire codebase fix rather than a quick "That looks wrong" feature branch. Either one works and is a project management choice, but it is useful for documentation purposes to understand what the purpose of a PR is.
And it would help me to know if I am tasked with fixing the entire codebase or just the portion I noticed while working on a different feature branch altogether.

jbakosi: Because we could find at least a single example that _does exactly the same change_ as this PR suggests and fails, that renders the change in this PR incorrect until we can convince ourselves to the contrary. Note that passing all regression tests is not a proof of correctness, since coverage is less than 100%.

You are not required at all to fix this problem. I would like you to try to reproduce it so I can be sure, that at least the problem can be reproduced elsewhere. I simply just wanted to alert you about the issue, that's all. If you have a better understanding of the issue, I welcome any insight, if not, that's totally fine too, but I don't think we should put in this change unless we can convince ourselves that it is correct.

PR #148, opened 2017-06-27
Add the option of storing output in ROOT framework for inciter.

adityapakki:
Requesting code, that enables writing to ROOT for Inciter, be reviewed and merged into the develop branch. Following are the capabilities added in this patch.

0. Make ROOT optional throughout the project.
1. Write Inciter run output in ROOT file layout using RootMeshWriter class.
2. Convert the output from point 1 into ExodusII layout utilizing ExodusIIMeshWriter through FileConvWriter class.
3. Add regression tests for the above two classes.


jbakosi:
This decreases coverage, most likely, because ROOT is optional and thus everything ROOT-related (which is most of this PR), is `ifdef`d out.

@junghans: Do you think it is possible to add ROOT to the gentoo travis container? (Or that will increase the size and/or the build time too much.)

I will install ROOT on the TeamCity build agents so that ROOT is always available there.

jbakosi:
@adityapakki Can you please also open a PR for https://github.com/quinoacomputing/quinoa-tpl/tree/root as well as https://github.com/quinoacomputing/cmake-modules/tree/root?

adityapakki:
@jbakosi Quick question, so I create pull requests for quinoa-tpl & cmake-modules from root to master ? Is that what you are expecting ?

jbakosi:
@adityapakki Yes. Thanks.

jbakosi:
@junghans: After you also approve, I will do the merge in concert with the cmake and TPL repos, when I have successfully installed ROOT on the TeamCity build agents. (I will simply enable ROOT on those agents for all builds.)

jbakosi:
@junghans Can you please take look at this? I'd like to merge when the travis builds passed.

jbakosi:
@rfbird Yeah, sorry. The context is that I have been getting ready to merge this PR, I pulled in the latest master from the cmake repo. Then the build broke because the PEGTL version file is not found, which was pulled in from the cmake repo (which was reviewed but not tested, since we don't have CI for that separately). So I reverted in this repo and this root branch the commit that I found to be the culprit until we find out why the PEGTL version check does not work. I also added a new PR requesting reverting the version check in the cmake repo.

It _is_ getting confusing for me to work with multiple repositories + subtree + submodules, but, hey, if it was simple and easy we would not be interested, right? ;-)

rfbird:
OK, as long as there is a plan!

> We choose to go to the Moon in this decade and do the other things, not because they are easy, but because they are hard; because that goal will serve to organize and measure the best of our energies and skills, because that challenge is one that we are willing to accept, one we are unwilling to postpone, and one we intend to win

jbakosi:
Nice!

Yes, there is a plan, the situation is clear in my head. It does take some effort and energy to keep it that way, though, with the current setup and a multi-dev team.

cmake/FindRoot.cmake
```diff
@@ -0,0 +1,62 @@
...
+                                               $ENV{ROOT_ROOT}/lib)
+find_library(ROOT_NET_LIBRARY NAMES Net HINTS ${ROOT_ROOT}/lib
+                                               $ENV{ROOT_ROOT}/lib)
+find_library(ROOT_IMT_LIBRARY NAMES Imt HINTS ${ROOT_ROOT}/lib
+                                               $ENV{ROOT_ROOT}/lib)
+find_library(ROOT_MATRIX_LIBRARY NAMES Matrix HINTS ${ROOT_ROOT}/lib
+                                               $ENV{ROOT_ROOT}/lib)
+find_library(ROOT_MATHCORE_LIBRARY NAMES MathCore HINTS ${ROOT_ROOT}/lib
+                                               $ENV{ROOT_ROOT}/lib)
+
```
junghans: Honestly, I would do that in a for loop.

cmake/LICENSE
```diff
@@ -0,0 +1,39 @@
+Copyright (c) 2016, Los Alamos National Security, LLC
```
junghans: 2017

jbakosi: Yes, I will update the year soon. If we wait a few months, we can update 2 years in a single step. ;-)

cmake/FindRoot.cmake
jbakosi: Sure, I will fix this.

cmake/LICENSE
jbakosi: Yep, we will update the copyright year everywhere in a next PR.

cmake/FindRoot.cmake
jbakosi: See https://github.com/quinoacomputing/cmake-modules/pull/10 for a solution.

cmake/LICENSE
jbakosi: See https://github.com/quinoacomputing/quinoa/pull/151, https://github.com/quinoacomputing/cmake-modules/pull/11, and https://github.com/quinoacomputing/quinoa-tpl/pull/17.

PR #149, opened 2017-06-28
Progress on manufactured solution cases testing the numerical methods in inciter

jbakosi:
Apologies in advance for a huge change set. (As you can see from the commit dates, this is a work of roughly a month). Changes at a high level bulleted below. Please look at the individual commit messages at least for 7a4a3b84 and 8bf19a74, which contain important additional details on the two main changes: (1) in the async logic of LinSysMerger and (2) adding time-dependent Dirichlet BCs, respectively.

Note that this is a work in progress. I copy-paste the to-do list from the bottom of 8bf19a74 at the bottom of the high-level bulleted list.

**High level description of changes:**
- Add time-, and space-dependent Dirichlet boundary conditions (BCs).
- Revamp how the Dirichlet BCs are assigned to mesh nodes.
- Change `LinSysMerger`'s asynchronous logic, as querying BCs is now NOT required during startup, i.e., before t=0, but only during time-stepping, but now at every time step. This also changes the polymorphic interface to `PDE::initialize()` and that of its "children", `Transport` and `CompFlow`.
- Add artificial viscosity, only used for compressible flow, required for stabilization of the numerical method for nonlinear equations.
- Add new manufactured solution case: Taylor-Green. Does not work yet and not yet tested.
- Improve on how the limit coefficient is computed during flux-corrected transport at Dirichlet boundaries. E.g., take care of a corner case, when every node of a cell happens to be on a Dirichlet boundary.
- Some code cleanup, e.g., remove unused variables.
- Some doxygen warnings cleanup.

**TODO:**

- This is a step towards getting the nonlinear energy growth (NLEG) test case working, and as such it has changes to `CompFlowProblemNLEnergyGrowth` and other `CompFlowProblem`s. However, only the time-dependent Dirichlet BCs and their assembly is tested and not the correctness of the NLEG solution (and this not even with regression tests).
- `CompFlow` (e.g., NLEG) is not part of the regression test suite.
- There are no regression tests that would test Dirichlet BCs, neither time-independent nor the new time-dependent ones.
- While the previous way of constructing the data structures for BCs worked with initial uniform refinement, the current time-dependent one does not yet work with it. (This does not fail any regression tests, however, since the existing tests that exercise AMR do not have Dirichlet BCs.)


src/DiffEq/MixMassFractionBeta.h
```diff
@@ -184,6 +184,7 @@ class MixMassFractionBeta {
     //! \param[in,out] particles Array of particle properties
     //! \param[in] stream Thread (or more precisely stream) ID
     //! \param[in] dt Time step size
+    //! \param[in] t Physical time
```
rspavel: Physical time of what?

src/Inciter/Carrier.C
```diff
@@ -120,7 +120,7 @@ Carrier::Carrier( const TransporterProxy& transporter,
 //!   mesh chunk we operate on
 //! \param[in] filenodes Map associating old node IDs (as in file) to new node
 //!   IDs (as in producing contiguous-row-id linear system contributions)
-//! \param[in] edgenodemap Map associating new node IDs ('new' as in producing
+//! \param[in] edgenodes Map associating new node IDs ('new' as in producing
 //!   contiguous-row-id linear system contributions) to edges (a pair of old
 //!   node IDs ('old' as in file). These 'new' node IDs are the ones newly
```
rspavel: At multiple points in this file: What does "contiguous-row-id linear" mean? You probably want to rephrase this to make it clear that this is related to the zoltan(?) reordering and partitioning that is going to be passed to hypre and indicate if this is a global ordering (I think yes?)

src/Inciter/Carrier.C
```diff
@@ -153,6 +153,32 @@ Carrier::Carrier( const TransporterProxy& transporter,
...
+    Assert( n < m_gid.size(),
+            "Local IDs must be lower than the local number of grid points" );
+    linnodes[ i.second ] = n;
+  }
+
+  // Access all side sets and their old node IDs (as in file) from LinSysMerger
+  auto& oldside = m_linsysmerger.ckLocalBranch()->side();
+
+  // Create map that assigns the local mesh node IDs mapped to side set ids,
+  // soring only those nodes for a given side set that are part of our chunk of
```
rspavel: soring =? sorting

src/Inciter/Carrier.C
```diff
@@ -374,118 +397,80 @@ Carrier::bc()
...
-                            { return p.first; } );
-    // for all node IDs we contribute to in the side set associate BC values and
-    // whether they are set for a component (for all PDE components)
-    std::unordered_set< std::size_t > bcnodes;
-    for (auto o : s.second) {
-      auto n = own(o);
-      if (n.first && inset(n.second,s.second) && ubc) {
-        assign( b, n.second );          // assign BC values to nodes
-        bcnodes.insert( n.second );     // store BC nodes
+  // Vector of pairs of bool and boundary condition value associated to mesh
```
rspavel: An unordered_map is not (necessarily) a vector

src/Inciter/Carrier.C
```diff
@@ -1205,33 +1189,32 @@ Carrier::velocity( std::size_t e )
...
-      auto lid = p->second;
-      for (ncomp_t c=0; c<b.nprop(); ++c)
-        b(i,c,0) = std::abs( m_dul(lid,c,0) + m_a(lid,c,0) );
-      ++i;
-    }
+  auto& dirbc = m_linsysmerger.ckLocalBranch()->dirbc();
+
+  if (dirbc.empty()) return true;
+
+  for (const auto& s : m_side)
```
rspavel: More documentation of what this chain of nested for loops is actually doing

src/Inciter/FluxCorrector.C
```diff
@@ -396,8 +396,8 @@ FluxCorrector::lim( const std::vector< std::size_t >& inpoel,
 //! \param[in] Ul Low order solution
 //! \param[inout] Q The maximum and mimimum unknowns of elements surrounding
 //!   each node
-//! \param[in,out] Limited antidiffusive element contributions scatter-added to
-//!   nodes
+//! \param[in,out] A Limited antidiffusive element contributions scatter-added
```
rspavel: Not a heavy fortran programmer, but what is the distinction between the inout and in,out intent?

src/Inciter/Partitioner.h
```diff
@@ -387,6 +388,8 @@ class Partitioner : public CBase_Partitioner< HostProxy,
     //! \param[in] p The PE uniquely assigns the node IDs marked listed in ch
     //! \param[in] cn Vector containing the set of potentially multiple chare
     //!   IDs that we own (i.e., contribute to) for all of our node IDs.
+    //! \param[in] ce Map associating a vector of chare IDs to edges (at which
```
rspavel: What does this mean? Presumably this is what chares own vertices that we are connected to but that are not locally owned? What does initial mesh refinement have to do with this?

src/Inciter/Partitioner.h
```diff
@@ -1260,6 +1265,7 @@ class Partitioner : public CBase_Partitioner< HostProxy,
                         CkCallback(CkReductionTarget(Transporter,load), host));
     }
     //! Compute average communication cost of merging the linear system
+    //! \param[in] host Host to signal to
```
rspavel: Are these computation tasks or wrappers for charm messages? Do we want to couple our compute so closely with our communication?

src/Inciter/Partitioner.h
```diff
@@ -1270,6 +1276,7 @@ class Partitioner : public CBase_Partitioner< HostProxy,
     }
     //! \brief Compute standard deviation of the communication cost of merging
     //!   the linear system
+    //! \param[in] host Host to signal to
```
rspavel: Are these computation tasks or wrappers for charm messages? Do we want to couple our compute so closely with our communication?

src/LinSys/LinSysMerger.h
```diff
@@ -970,16 +970,14 @@ class LinSysMerger : public CBase_LinSysMerger< HostProxy,
               "Nonzero values of distributed matrix on PE " +
               std::to_string( CkMyPe() ) + " is incomplete: cannot set BCs" );
 
-      for (auto& r : m_lhs) {
-        auto it = m_bc.find( r.first );
-        if (it != end(m_bc)) {
-          auto& diag = tk::ref_find( r.second, r.first );
+      for (const auto& n : m_bc) {
```
rspavel: Documentation for what this group of nested for loops with single character variable names is doing?

src/PDE/CompFlowProblem/TaylorGreen.h
```diff
@@ -0,0 +1,302 @@
+// *****************************************************************************
+/*!
+  \file      src/PDE/CompFlowProblem/TaylorGreen.h
+  \author    J. Bakosi
```
rspavel: Aren't we removing these to follow a more traditional open source development model?

src/PDE/CompFlowProblem/RayleighTaylor.h
```diff
@@ -0,0 +1,521 @@
+// *****************************************************************************
+/*!
+  \file      src/PDE/CompFlowProblem/RayleighTaylor.h
+  \author    F. Gonzalez
```
rspavel: Aren't we removing these to support a more traditional open source development model?

src/PDE/CompFlowProblem/NLEnergyGrowth.h
```diff
@@ -0,0 +1,462 @@
+// *****************************************************************************
+/*!
+  \file      src/PDE/CompFlowProblem/NLEnergyGrowth.h
+  \author    F. Gonzalez
```
rspavel: Aren't we removing these to support a more traditional open source development model?

src/PDE/CompFlowProblem/UserDefined.h
```diff
@@ -0,0 +1,187 @@
+// *****************************************************************************
+/*!
+  \file      src/PDE/CompFlowProblem/UserDefined.h
+  \author    J. Bakosi
```
rspavel: Aren't we removing these to support a more traditional open source development model?

src/PDE/CompFlowProblem/VorticalFlow.h
```diff
@@ -0,0 +1,308 @@
+// *****************************************************************************
+/*!
+  \file      src/PDE/CompFlowProblem/VorticalFlow.h
+  \author    J. Bakosi
```
rspavel: Aren't we removing these to support a more traditional open source development model?

src/Inciter/FluxCorrector.C
jbakosi: This is not a fortran but a doxygen comment. I believe in the past one was required to use `inout` then they switched the syntax to `in,out`, then now doxygen, I believe, understands both. But will put in a comma so it is consistent.

src/Inciter/Partitioner.h
jbakosi: `ce` is a map that associates a vector of chare IDs to edges. That is, the map key is an edge (an array of two integers, node IDs of mesh vertices, see UnsMesh::Edge) and the map value is a vector of chare IDs. This map is empty if no initial mesh refinement happened.

src/Inciter/Partitioner.h
jbakosi: I'm not sure if I understand your question. The reason why this is done this way, is explained starting at line [1209](https://github.com/quinoacomputing/quinoa/blob/mms/src/Inciter/Partitioner.h#L1209).

src/Inciter/Partitioner.h
jbakosi: I'm not sure if I understand your question. The reason why this is done this way, is explained starting at line [1209](https://github.com/quinoacomputing/quinoa/blob/mms/src/Inciter/Partitioner.h#L1209).

src/PDE/CompFlowProblem/TaylorGreen.h
jbakosi: Yep, thanks for catching this.

src/Inciter/Carrier.C
jbakosi: _storing_, thanks for the catch.

src/Inciter/Carrier.C
jbakosi: Thanks, attempted to clarify it in 86cc28a3.

src/Inciter/Carrier.C
jbakosi: > Vector of pairs of bool and boundary condition value associated to mesh node IDs at which the user has set Dirichlet boundary conditions for all PDEs integrated

Since it is associated, it is an associative container, a map.

src/Inciter/Carrier.C
jbakosi: Thanks. This is really useful. Addressed in 63371176.

jbakosi:
Thanks @rspavel for your comments. They are very useful. I think I have attempted to address all of them. Please let me know if something is still unclear.

PR #151, opened 2017-07-08
Update copyright year to 2017

jbakosi:
See script/update_copyright.sh for how the copyright was updated.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,7 +1,7 @@
-Copyright (c) 2016, Los Alamos National Security, LLC
+Copyright (c) 2017, Los Alamos National Security, LLC
 All rights reserved.
 
-Copyright 2016. Los Alamos National Security, LLC. This software was produced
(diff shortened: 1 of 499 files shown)


LICENSE
```diff
@@ -1,7 +1,7 @@
-Copyright (c) 2016, Los Alamos National Security, LLC
+Copyright (c) 2017, Los Alamos National Security, LLC
```
junghans: Wouldn't it be 2016-2017?

jbakosi: I have no idea what is better.

jbakosi: I guess we should do the range as that is consistent with the copyright in the individual files.

junghans: My thinking was, 2016 is when LANL took over and hence the copyright holds from 2016-now (2017).

jbakosi: Yeah, makes sense. I will fix it.

PR #152, opened 2017-07-19
Correct asynchronous logic and enable testing with randomized queues

jbakosi:
This takes care of https://github.com/quinoacomputing/quinoa/issues/131.

The reasons for the changes in this PR are rather deep as many of them affect the asynchronous logic. Therefore I have tried my best to describe why the various changes were made in as much detail as I thought it was necessary. Please take a look at the individual commit messages as they contain a lot of information. For that reason I would suggest looking at the diff commit-by-commit.

Besides addressing the various issues revealed by Charm++'s randomized message queues, I also added a new `buildenv` image that has a preinstalled Charm++ build with randomized queues at https://github.com/quinoacomputing/buildenv/blob/master/alpine-rndq. This is also hooked up into travis, see e.g., https://travis-ci.org/quinoacomputing/quinoa/builds/255336762. Additionally, I have also setup the TeamCity build agents and configured 10 new matrix entries with Charm++'s randomized message queues, testing various compilers, standard libraries in debug and release mode, which all pass. The TC code is added by https://github.com/quinoacomputing/quinoa-teamcity/commit/f625e24c56c2cfaa2860ab0bf825f1fbc0244c7c.


src/Inciter/Carrier.C
```diff
@@ -244,9 +244,57 @@ Carrier::vol()
...
+{
+  Assert( V.size() == gid.size(), "Size mismatch" );
+
+  for (std::size_t i=0; i<gid.size(); ++i) {
+    auto bid = tk::cref_find( m_bid, gid[i] );
+    Assert( bid < m_volc.size(), "Indexing out of bounds" );
+    m_volc[ bid ] += V[i];
+  }
+
+  if (++m_nvol == m_msum.size()) {
```
rspavel: What are the Charm++ rules regarding handling counters like this? Do we have a guarantee that entry invocations will always be atomic/serialized (unless otherwise flagged)? I know that in practice this is true, but if we don't have a guarantee then we are setting up a ticking time bomb of race conditions.

This is theoretically a problem elsewhere (which is why I assume it is probably fine), but this is the kind of thing we want to avoid in a fix branch unless we know it will be fine

src/Inciter/Carrier.C
```diff
@@ -244,9 +244,57 @@ Carrier::vol()
...
+
+  // Sum mesh volume to host
+  tk::real V = 0.0;
+  for (auto v : m_v) V += v;
+  contribute( sizeof(tk::real), &V, CkReduction::sum_double,
+    CkCallback(CkReductionTarget(Transporter,totalvol), m_transporter) );
+}
+
+void
 Carrier::stat()
```
rspavel: If this is specifically for cell statistics, wouldn't it be smart to rename this to cellStat() or something more indicative?

src/Inciter/Carrier.C
```diff
@@ -908,6 +927,11 @@ Carrier::aec( const tk::Fields& Un )
       for (auto i : n.second) p.push_back( m_p[ tk::cref_find(m_lid,i) ] );
       thisProxy[ n.first ].comaec( n.second, p );
     }
+
+  ownaec_complete();
+  #ifndef NDEBUG
+  ownaec_complete();
```
rspavel: While having special flags for debugging is always nice, having two separate control paths and communication requirements between the debug and mainline path is a recipe for disaster and a way to find very hard to debug problems as we are fundamentally changing the logic of the program when we debug.

src/Inciter/Partitioner.h
```diff
@@ -1283,7 +1284,7 @@ class Partitioner : public CBase_Partitioner< HostProxy,
     //! Signal back to host that we are ready for partitioning the mesh
     void signal2host_setup_complete( const CProxy_Transporter& host ) {
       Group::contribute(
-        CkCallback(CkIndex_Transporter::redn_wrapper_partition(NULL), host ));
+        CkCallback(CkIndex_Transporter::redn_wrapper_part(NULL), host ));
```
rspavel: Is there no way to handle the dependencies outside of a barrier?

src/Inciter/Transporter.C
```diff
@@ -329,7 +332,7 @@ Transporter::load( uint64_t nelem )
 }
 
 void
-Transporter::partition()
+Transporter::part()
```
rspavel: What benefit is there in renaming partition to part? Based upon the description neither name was particularly meaningful but ``partition'' contained much more semantic meaning

src/Inciter/Transporter.h
```diff
@@ -126,7 +133,7 @@ class Transporter : public CBase_Transporter {
     //! \brief Reduction target indicating that all Partitioner chare groups
     //!   have finished setting up the necessary data structures for
     //!   partitioning the computational mesh and we are ready for partitioning
-    void partition();
+    void part();
```
rspavel: See above

src/Inciter/Transporter.h
```diff
@@ -195,7 +202,7 @@ class Transporter : public CBase_Transporter {
     void rowcomplete();
 
     //! Reduction target summing total mesh volume
-    void vol( tk::real v );
+    void totalvol( tk::real v );
```
rspavel: In general, we should consider a standardized way to indicate what is and isn't a reduction target. I think some of the stuff in transporter is ``redn_''?

src/Walker/Collector.C
```diff
@@ -29,69 +29,13 @@ CkReduction::reducerType PDFMerger;
 using walker::Collector;
 
 void
-Collector::chareOrd( const std::vector< tk::real >& ord )
+Collector::chareOrd( const std::vector< tk::real >& ord,
```
rspavel: what is the semantic meaning of "chareOrd"?

src/Walker/Collector.C
```diff
@@ -113,33 +59,46 @@ Collector::chareOrdPDF( const std::vector< tk::UniPDF >& updf,
...
+    std::fill( begin(m_ordinary), end(m_ordinary), 0.0 );
 
     // Serialize vector of PDFs to raw stream
     auto stream = tk::serialize( m_ordupdf, m_ordbpdf, m_ordtpdf );
 
     // Create Charm++ callback function for reduction.
     // Distributor::estimateOrdPDF() will be the final target of the reduction
     // where the results of the reduction will appear.
-    CkCallback cb( CkIndex_Distributor::estimateOrdPDF(nullptr), m_hostproxy );
+    CkCallback c2( CkIndex_Distributor::estimateOrdPDF(nullptr), m_hostproxy );
```
rspavel: "c2" is not a good variable name in any situation and will just make it harder for people to understand the code

src/Walker/Collector.C
```diff
@@ -113,33 +59,46 @@ Collector::chareOrdPDF( const std::vector< tk::UniPDF >& updf,
   for (const auto& p : tpdf) m_ordtpdf[i++].addPDF( p );
 
   // If all chares on my PE have contributed, send partial sums to host
-  if (m_nopdf == m_nchare) {
+  if (m_nord == m_nchare) {
+
+    // Create Charm++ callback function for reduction
+    CkCallback c1( CkReductionTarget( Distributor, estimateOrd ), m_hostproxy );
```
rspavel: And neither is "c1" (I noticed this after seeing "c2". If there is a "c0" then same applies)

src/Walker/Collector.C
```diff
@@ -113,33 +59,46 @@ Collector::chareOrdPDF( const std::vector< tk::UniPDF >& updf,
...
+
+    m_nord = 0;
   }
 }
 
 void
-Collector::chareCenPDF( const std::vector< tk::UniPDF >& updf,
-                        const std::vector< tk::BiPDF >& bpdf,
-                        const std::vector< tk::TriPDF >& tpdf )
+Collector::chareCen( const std::vector< tk::real >& cen,
```
rspavel: Again, while not a good name removing "PDF" removes semantic meaning from the function name

jbakosi:
@rspavel wrote:
> As always, I probably missed things and generally only flagged an issue once rather than at every occurrence.

That's quite alright. I appreciate any insight anyone can offer.

> I don't like that our solution was to basically add a bunch of barriers, but if that is the preferred way to fix race conditions then so be it.

I don't like that either and since it is **not** my _preferred_ way to fix race conditions this way, I do not appreciate that assume so. Beside generalizing the criticism of the specific changes in this PR to "preferred way", I would value proposal(s) of different solution(s) a lot more. In other words: Do you have a better idea? That would be more constructive, interesting, and infinitely more useful.

> But I definitely feel we should not be renaming functions and variables to make deriving their meaning even harder. Text is cheap (especially compared to our dependencies) and having variable/function names with semantic meaning is always preferred.

I agree.

> While I am far from a good software engineer and am all for making some scientific code with a future technical debt, we should at least try to not make things harder for future contributors as we modify the code.

It is hard to disagree. However, the contribution value of such general statements are immeasurable. Also, I do not appreciate the cynical tone.

src/Inciter/Carrier.C
jbakosi: Charm++ manual, Sec. _Basic Concepts_:
> Ordinary entry methods are completely non-preemptive -- Charm++ will not interrupt an executing method to start any other work, and all calls made are asynchronous.

src/Inciter/Carrier.C
jbakosi: I pondered on this myself as well. It computes statistics derived from edge-lengths and cell volumes. Since these are so far the only kind of statistics we compute in _inciter_,  I thought `stat` is pretty much unambiguous.

src/Inciter/Carrier.C
jbakosi: I could not agree more. Any suggestions on how to do this better?

src/Inciter/Partitioner.h
jbakosi: I don't understand you question. Can you elaborate?

src/Inciter/Transporter.C
jbakosi: You are correct. However, calling this function is now only a necessary but not sufficient condition for starting partitioning. So keeping the name would be misleading. Do you have a better idea?

src/Inciter/Transporter.h
jbakosi: Yeah. I feel the same. Any suggestions?

src/Walker/Collector.C
jbakosi: Given by first line documenting the function:
_Chares contribute ordinary moments and ordinary PDFs_

src/Walker/Collector.C
jbakosi: That used to be `c1` and `c2` in a previous attempt of solving this problem which required two reductions starting from the same function to avoid name clashes. Now that is not required, but who cares?

jbakosi: I just notice that there _are_ `c1` as well as `c2`. So how can I improve on this?

src/Walker/Collector.C
jbakosi: I agree, but I don't care enough about it.

src/Walker/Collector.C
jbakosi: Changed because now that function not only contributes central moments but central PDFs as well.

src/Inciter/Partitioner.h
rspavel: You are signaling completion with a blocking reduction. For all intents and purposes, a blocking reduction is a barrier. Hence we are using barriers to block between phases.

The alternative is to use finer grain signals and messages to indicate when the data required to proceed is available and proceed then.

Reductions should always be the last resort as that is a barrier. If you truly have no way to not use it (likely because we are still in the MPI-space) then go for it. But the general reason to use an asynchronous task based runtime/advanced programming model/whatever is to take advantage of constructs to avoid those. Unless you truly need the entire global state to be at a certain point to continue (which should only be true for when we rely on MPI libraries), figure out what data you need and proceed.

The simplest way to think of this is a stencil as that is one of the most fundamental patterns. The traditional method is to put a barrier after every time step. What should be done is to instead have the computation of a tile wait until all inputs (generally ghost cells) are provided and proceed then. This lets you avoid bulk synchronous parallelism and potentially get fun things like wavefront parallelism-esque executions. Obviously you can't always do that (maybe the data you need is a reduction on dt), but it should always be targeted until determined to not be beneficial.

src/Inciter/Transporter.h
rspavel: Maybe ``redn_''? I don't like prefixes but you did that elsewhere and it at least is indicative

src/Walker/Collector.C
rspavel: My point was that it is a bad function name that makes reading the code a lot harder.

src/Walker/Collector.C
rspavel: Give them a semantic meaning that indicates what they are for?

src/Inciter/Carrier.C
rspavel: Fair enough. That's what I figured but it is always good to check

src/Inciter/Carrier.C
rspavel: The general idea is to not need deep knowledge of the entire system to understand a block of code.

In this case, let's assume we output data in a manner similar to the following (even if we don't, this is a thought exercise):

```
void printStats()
{
foo.stat();
bar.stat();
man.stat();
}
```

Now, that means nothing. If we go one step better we would have

```
void printStats()
{
inciterArray.stat();
linksysArray.stat();
partitionerArray.stat();
}
```

That is a lot better. Now we know what kinds of objects we are getting data from. So if we understand what kind of data inciter, partitioner, and linksys generate, we are golden.

But why stop there?

```
void printStats()
{
inciterArray.printCellStatistics();
linksysArray.printHypreStatistics();
partitionerArray.printStandardDeviationOfPartitionSize();
}
```

Or something similar. I don't have the deep understanding of quinoa that you do so I am kind of reaching based upon my vague understandings of what those chares did the last time I did a deep dive.

I can't speak to its overall value but I was told to read The Practice of Programming by Kernighan and Pike and it seems pretty useful for explaining methodologies for general coding practices. There are obviously much better books out there, but this one is short and full of examples.

src/Inciter/Partitioner.h
jbakosi: Okay, thanks for the clarification.

This call signals the host that the mesh graph has been read and the cell centroids have been computed by the given PE. This is necessary but not sufficient to call Zoltan, since that also needs the number of chares we need Zoltan to partition the mesh into. The number of chares is computed in `Transporter::load()` another `reductiontarget` with necessary contributions from all `Carrier` chares. Only when both are ready can we call Zoltan. Do you agree?

It is, of course, possible to do individual (point-to-point) calls to replace both reductions, but they will not use Charm++'s tree-based aggregation, and they are a lot less perfomant (I tested this). However, both the total number of chares (into which we want to partition the mesh) and the fact that the mesh has been read and its centroids have been computed must be completed before we can call Zoltan.

What you write is general and obvious. In light of the above two paragraphs with specific information on the problem and its constraints on correctness, can you offer a better alternative?

src/Walker/Collector.C
jbakosi: They are both call-backs used immediately in their following lines and never again. Plus every line is commented. I refuse to believe that you do not understand their meaning.

src/Walker/Collector.C
jbakosi: I understood your point. Do you have a better alternative?

src/Inciter/Partitioner.h
rspavel: If we need to use a barrier because of our reliance on MPI, go for it. That was my question. I apologize if that was obvious, but I don't have the deep understanding of quinoa you do. In fact, most of my comments are to that effect. Hero programmers, regardless of what some people say, are a bad thing and should be avoided like the plague. Code should be understandable and easy to follow

So if the solution of a barrier makes sense, go for it. Sometimes you actually DO want a barrier (particularly when you need quiescence to change to MPI)

As for reductions versus p2p: Yes, if you are doing a pattern that is a reduction then it is hard to beat a reduction. And if you truly need one, go for it. Same with barriers. But if you can avoid the barrier, you should. Because even if you do need everyone to communicate in a tree pattern, you can potentially get your signal, send your signal, and then get back to work. The alternative is having to wait until everyone has received the signal and has the result (or dummy value if it is just a barrier) which involves a lot of idle time and is the graph of choice for runtime developers arguing the utility of a tool.

Maybe when MPI's non-blocking reductions are more ubiquitous that will become the gold standard as it combines the best of both worlds. But as it is right now, we need to weigh the pros and cons of everything. Hence my asking if this was the approach we should take.

src/Inciter/Transporter.C
rspavel: Based upon the comment description and code inside, something along the lines of initiatePartitioning()? Or even "reportMeshTiming()". Although the part_complete() at the bottom makes me confused as to whether this is actually the signal for the completion of partitioning? But the text suggests this was just mesh reading?

src/Walker/Collector.C
rspavel: Not really because I don't actually understand what is going on.

But again, based on the text, something like sumOrdinaryMomentsAndPDFs()? If length is an issue (and that is pretty heinous) sumOrdMomsAndPDF() or along the same lines. That is much more indicative of what that function does than chareOrd()

src/Inciter/Transporter.h
jbakosi: That's not a bad idea. But note that where there is `redn_` so far in the code, it is not my choice, but because we need to call a Charm++ macro (which starts with `redn_`). If we do this rename, though, I feel should be done consistently everywhere, so in another PR.

src/Inciter/Carrier.C
rspavel: The issue is that it is not immediately obvious, at least to a non-domain expert, what "inciter" generates. If someone were to tell me "Go in and add a counter for the number of cells that change by more than 10% in any given time step" I would have no idea where to look and would need to add a lot of sentinels and do a lot of grepping. Whereas, if I go check the statistics aggregation section and see "Ah, inciterArray prints cell statistics. I should go look at whatever inciterArray is" I am going to spend more time doing meaningful work.

As it stands now, I would have to hope that you are still being paid to work on this project and have time to help me find it. And if I were actually an external collaborator contributing to an open source project, I am not going to even bother to do that.

That is the reason for software engineering. Making code better and faster is nice, but it is more about ensuring that future workers on a project need <whatever> science knowledge, not quinoa knowledge.

But you are in charge and if you think this approach is the one we should take, disregard this. Just understand that the time we save by not fixing this now is paid tenfold during onboarding of new developers and reduced contributions from the open source community.

src/Walker/Collector.C
jbakosi: I agree, both are pretty heinous.

Have you thought about how you would _talk_ **in person** about this code? (Would that be harder and more time-wasting with the longer function names?)

src/Inciter/Partitioner.h
jbakosi: Yes, unfortunately, in this case, I'm afraid, we need the barriers because Zoltan is MPI.

src/Walker/Collector.C
rspavel: I think I would probably make fun of whoever wrote the name sumOrdMomsAndPDF(), but I think it would also save time over having to explain what "chareOrd()" means. 

But I also wouldn't use either function name if I were discussing this. I would give a very high level talk about the algorithm we use and just say "And then we have each collector perform a partial sum of the ordinary moments and PDFs prior to a global reduction".

But I also know that if I am looking through the code, not having to cross reference the manual to understand what chareOrd() does would be nice. Which is really what I am thinking of.

src/Inciter/Transporter.C
jbakosi: Yeah, I agree that his is somewhat misleading, because it is really a leftover from when this was not only a necessary but also a sufficient condition for partitioning (which was incorrect).

Timing is not the main job of `Transporter::part()`.

What it really means is that a necessary condition is met and that condition is the mesh read and computing centroids on all PEs (if a geometric partitioner is configured by the user) have been completed. It does not initiate partitioning. I'm not sure what is a good single word for that, so I punted and kept it `part()` which, I agree, is pretty awful.

src/Inciter/Carrier.C
jbakosi: Alright! Good points. How about `meshStat()`? (Does that differentiate enough from _particle_ statistics in `walker`?)

jbakosi:
This was opened 19 days ago. I will merge it tomorrow morning, unless there is a concern.

PR #153, opened 2017-07-20
Expose C++ language standard to user

jbakosi:
This PR proposes to expose the configuration of the C++ language standard to the user via `src/CMakeLists.txt`. Currently, our minimum required standard is C++11. For some reason, the macports-installed ROOT on mac using the gnu compilers, however, requires C++14 for the ROOT header files and errors out if `-std=c++11` is used by the compiler, while it builds fine with `-std=c++14`. Thus this change exposes this setting so that the mac (TeamCity) CI builds, enabling ROOT, can explicitly override the default, which remains C++11. See https://github.com/quinoacomputing/quinoa-teamcity/commit/657a5c612f5402699826f7ca3a7377c781c635bc for how this is used.

Tested the cmake logic by passing `-DCMAKE_CXX_STANDARD=98`, which correctly triggers a fatal error.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -133,11 +133,16 @@ execute_process(
 )
 string(REGEX REPLACE "[\r\n]" "" BUILD_DATE "${BUILD_DATE}")
 
-# Set the requirement for the C++ standard
-set(CMAKE_CXX_STANDARD 11)
+# Set the requirement for the C++ standard. If user sets it, use that, if not,
+# require C++11 by default..
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 11 CACHE STRING "Required C++ language standard. See https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html for supported values")
+elseif(${CMAKE_CXX_STANDARD} EQUAL 98)
+  message(FATAL_ERROR "Minimum required C++ standard: C++11")
+endif()
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-message(STATUS "Required language standard: C++${CMAKE_CXX_STANDARD}")
+message(STATUS "C++ language standard: C++${CMAKE_CXX_STANDARD}")
 
 # Get git revision
 include(GetGitRevisionDescription)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -133,11 +133,16 @@ execute_process(
 )
 string(REGEX REPLACE "[\r\n]" "" BUILD_DATE "${BUILD_DATE}")
 
-# Set the requirement for the C++ standard
-set(CMAKE_CXX_STANDARD 11)
```
junghans: Wouldn't it be easier to put the `11` into and option and set it to `14` with MacPorts?

jbakosi: I'm not sure if I understand. We are not calling cmake from macports. This is called from TeamCity, which has a cmake command line on which now this is exposed.

jbakosi:
Well, this may be a dead-end for what I'm trying to do... (that is enabling the ROOT-related regression tests on mac CI builds), because I get the following runtime errors, running the root-regression tests, with the macports-installed `root6` package:

```
Error in <TTree::Branch>: The class requested vector<__cxx11::basic_string<char,char_traits<char>,allocator<char> > >) for the branch "variables" is an instance of an stl collection and does not have a compiled CollectionProxy. Please generate the dictionary for this collection (vector<__cxx11::basic_string<char,char_traits<char>,allocator<char> > >) to avoid to write corrupted data.
107:
107:  *** Break *** segmentation violation
107: [/usr/lib/system/libsystem_platform.dylib] _sigtramp (no debug info)
107: [<unknown binary>] (no debug info)
107: [[...]/build/gnu/IO/libquinoa_meshio.dylib] tk::RootMeshWriter::writeNodeVarNames(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&) const [...]/src/IO/RootMeshWriter.C:173
107: [[...]/build/gnu/Inciter/libquinoa_inciter.dylib] inciter::Carrier::writeMeta() const [...]/src/Inciter/Carrier.C:812
107: [[...]/build/gnu/Inciter/libquinoa_inciter.dylib] inciter::Carrier::setup(double) [...]/src/Inciter/Carrier.C:387
107: [[...]/build/gnu/Inciter/libquinoa_inciter.dylib] inciter::CkIndex_Carrier::_call_setup_marshall6(void*, void*) RNGTest/../Inciter/carrier.def.h:1881
107: [[...]/build/gnu/Main/inciter] CkDeliverMessageReadonly (no debug info)
```

I just noticed that the default `-std=` argument for building the `root6` macports package is `cxx17`, which is probably the reason, there is some `__cxx11`-prefixed symbols are missing and I get the same problem even when I can pass `-std=c++14` to our TPL and code builds.

Instead, I'm going to rebuild the macports package with `-cxx17`, which will default to `-std=c++11`. That should generate the above missing symbols... If that works out, I will simply close PR without merging, but I guess it is still good to have this documented so we can come back if we ever need to expose the CXX_STANDARD on the cmake command line.

jbakosi:
> Kudos for the excellent error ...

While that builds the macports package fine, I still get the error above.

jbakosi:
Managed to build root on mac using clang manually. This is not needed for now.

Issue #12, opened 2015-07-02
Implement a fall-back RNG library


jbakosi:
Need a fall-back RNG (Boost?, STL?) guaranteed to work everywhere.
(RNGSSE's assembly does not work on ARM, Intel's MKL is commercial.)

jbakosi:
Pretty much the same Issue #11.

jbakosi:
ARM support is not critical at this time. RNGSSE is a drop-in replacement for MKL.

PR #155, opened 2017-07-31
Corrected Copyright in Files Post License Transfer

rspavel:
Following discussions, corrected copyright in files to address status relative to when the license of the code transferred. Files created after the transfer date were never under the previous license and have been updated accordingly.

I do suggest we remove copyright from all non-license files to save ourselves time on copyright range changes in the future, but that is a problem for a different discussion and is less of an immediate concern.

List of affected files generated via 
`git log --diff-filter=A --name-status --oneline  c63ca7ae816fc52f6796a5cf4f0dbe365619f1ff..928819e304edcb379f2884da73b0b239b494d3a0`

Not guaranteed to have gotten every file (another reason to remove copyrights altogether), but haven't found any that were created post-transfer that still exist and don't have an updated copyright line.

### DIFF
--- a/cmake/BuildShared.cmake
+++ b/cmake/BuildShared.cmake
@@ -1,7 +1,7 @@
 ################################################################################
 #
 # \file      cmake/BuildShared.cmake
-# \copyright 2012-2015, J. Bakosi, 2016-2017, Los Alamos National Security, LLC.
+# \copyright 2016-2017, Los Alamos National Security, LLC.
 # \brief     Set default value for building shared libs if none was specified
 #
 ################################################################################
--- a/cmake/BuildType.cmake
+++ b/cmake/BuildType.cmake
@@ -1,7 +1,7 @@
 ################################################################################
 #
 # \file      cmake/BuildType.cmake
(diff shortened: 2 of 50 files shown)


junghans:
Should the files in `cmake/` be fixed in https://github.com/quinoacomputing/cmake-modules?

rspavel:
Probably, and I can migrate those changes there shortly. This was a largely automated process and was primarily intended as a way to spark a discussion of the much easier solution (remove copyright blocks from source files)

rspavel:
@junghans Updated cmake-modules with a PR (#12)

jbakosi:
Even if we decide to care about this this much (which I don't), the changes clearly do not reflect your intention: most of these changes are to files that were clearly created before the transfer of copyright. (Their creation date is probably not correct, which I also could not care less about.)

Also, I'm in favor of keeping this copyright info in the files. They are easy to update using `script/update_copyright.sh`.

I do not understand why this is of "immediate concern". Can we please concentrate on capability instead of cosmetics?

Issue #3, opened 2015-07-02
Add logger


jbakosi:
Useful? Requirements?
- Pantheios?
- Glog?
- Boost.Log?
- github/gabime/spdlog?

jbakosi:
Probably not very necessary.

PR #156, opened 2017-08-02
Initial Merge of AMR Functionality 

rfbird:
Replaces existing 1:8 hard coded initial refinement with the same, but using the TetAMR library 

Seems to pass AMR Tests 

This is the first step, to get the TetAMR code merged in, before we move AMR to be before partitioning


src/Inciter/AMR/Loggers.h
```diff
@@ -0,0 +1,19 @@
...
+#define trace_out while(0) std::cout
+#endif /* ENABLE_TRACE */
+
+#ifdef ENABLE_DEBUG
+#define debug_out std::cout << "__DEBUG: "
+#else
+#define debug_out while(0) std::cout
+#endif /* ENABLE_DEBUG */
+
+#endif
```
jbakosi: I don't think we should perform anything major as part of this PR, just wanted to let folks now that I have also thought about adding some kind of logging facility, see https://github.com/quinoacomputing/quinoa/issues/3.

Should we care about logging as much as pulling in a TPL for it do it well? Or just pull our own but useable from the whole code and not only from AMR? Any thoughts? (Perhaps discuss in #3).

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+#define TETAMR_AMR_DATA_H
+
+#include <vector>
+#include <cstddef>
+
+#include "types.h"
+
+namespace AMR {
+    // Types
+    using real_t = tk::real;
```
jbakosi: `tk::real` is defined in [Base/Types.h](https://github.com/quinoacomputing/quinoa/blob/master/src/Base/Types.h#L17). Thus I think this needs an `#include <Types.h>`, otherwise this header (as is right now if compiles) is order dependent. Or that is `types.h`? If so, let's combine!

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+
+    const size_t DEFUALT_CHILD_NUMBER = 0; // Used to mark "doesn't have children"
+    const size_t DEFAULT_NUM_CHILDREN = 0; // Default to no children
+
+    // TODO: Populate this enum with the available refinement cases
+    enum Refinement_Case { initial_grid, one_to_two, one_to_four, one_to_eight,
```
jbakosi: As we already discussed, it would probably be better to convert these enums to type-safe ones.

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+        two_to_eight, four_to_eight, none };
+
+    enum Edge_Lock_Case {unlocked, locked, intermediate, temporary};
```
jbakosi: Same for this: make enum type-safe.

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+        size_t A;
+        size_t B;
+        real_t refinement_criteria;
+        bool needs_refining; // TODO: This could possibly be deduced implicitly
+        bool needs_derefining; // TODO: Marge this with needs_refining
+        bool is_dead;
+        Edge_Lock_Case lockCase; // TODO: Refactor this to match _ style?
+
+        // Explicit Empty Constructor
+        Edge_Refinement() {
```
jbakosi: If it is `explicit`, let's tell the compiler too.

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+            A = 0;
+            B = 0;
+            refinement_criteria = 0.0;
+            needs_refining = false;
+            needs_derefining = false;
+            is_dead = false;
+            lockCase = Edge_Lock_Case::unlocked;
```
jbakosi: Let's do this with an initializer list instead. No big deal, but avoids a needless copy.

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+        }
+
+        // Explicit copy Constructor
+        Edge_Refinement(const Edge_Refinement& a) :
```
jbakosi: I don't see a copy assignment. Let's add it and implement the copy constructor in terms of the copy assignment and same for move constructor and assignment. Or even better: Can we get rid of all of the constructors including the copy constructor? That way the compiler auto-generates everything. This should work as the state is simply POD.

src/Inciter/AMR/Refinement_State.h
```diff
@@ -0,0 +1,167 @@
...
+            lockCase(lockCase_in)
+        {
+            // Empty, all implicit.
+            // Could add logic here to reconcile needs_refining and needs_derefining
+        }
+    };
+
+    //stop it being copied around?
+
+    class Refinement_State {
```
jbakosi: The same as [above](https://github.com/quinoacomputing/quinoa/pull/156/files#r131698931) should work here too: I would do `struct` and only put in the state and perhaps a constructor with an initializer list with the defaults. Then let the compiler auto-generate everything else as needed.

src/Inciter/AMR/active_element_store.h
```diff
@@ -0,0 +1,54 @@
...
+        public:
+            /**
+             * @brief Function to add active elements
+             *
+             * @param id The master_id this should map to
+             */
+            void add(size_t id)
+            {
+                // Check if that active element already exists
+                assert( !exists(id) );
```
jbakosi: Please use `Assert` instead to make it consistent with the rest of the code (for better or worse). `Assert` is defined in [Base/Exception.h](https://github.com/quinoacomputing/quinoa/blob/master/src/Base/Exception.h#L36) and allows passing a message and will generate a stack trace in `DEBUG`.

src/Inciter/AMR/edge.cpp
```diff
@@ -0,0 +1,9 @@
+#include "edge.h"
+
+// Implementation of the friend function, extending osstream and not an edge_t
+// member
+std::ostream& operator<<(std::ostream& os, const edge_t& e)
+{
+    os << e.data.first << "-" << e.data.second;
+    return os;
+}
```
jbakosi: Can we use `.C` as the extension? Just to make it consistent with the rest of the code (for better or worse).

src/Inciter/AMR/edge.h
```diff
@@ -0,0 +1,80 @@
+#ifndef AMR_edge_t_h
+#define AMR_edge_t_h
+
+#include <iostream>
+#include <ostream>
+
+class edge_t {
+    using edge_ = std::pair<size_t, size_t>;
```
jbakosi: Minor thing: Let's use `m_` for member data to make it consistent with the rest of the code. Is that disgusting?

src/Inciter/AMR/edge.h
```diff
@@ -0,0 +1,80 @@
...
+        friend std::ostream& operator<<(std::ostream&, const edge_t&);
+
+    public:
+
+        edge_ get_data() const {
+            return data;
+        }
+
+        // Constructors
+        edge_t()
```
jbakosi: This could be probably just deleted.

src/Inciter/AMR/edge.h
```diff
@@ -0,0 +1,80 @@
...
+        {
+        }
+        edge_t(size_t A, size_t B)
```
jbakosi: I would try to use initializer list to avoid a copy.

src/Inciter/AMR/edge.h
```diff
@@ -0,0 +1,80 @@
...
+        {
+            data = e.data;
+        }
+
+        // TODO: Sanity check this
+        edge_t& operator=(const edge_t& e)
+        {
+            data = e.data;
+            return *this;
+        }
```
jbakosi: I bet both the copy constructor and assignment can be just deleted and the compiler will auto-generate them.

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
+#ifndef AMR_edge_store_h
+#define AMR_edge_store_h
+
+namespace AMR {
+
+    class edge_store_t {
+        public:
+            // TODO: convert this to an unordered map with a custom hash (can lift from Quinoa)
```
jbakosi: Yep! ;-)

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
...
+            /**
+             * @brief Given nodes A and B, generate an edge between them
+             *
+             * @param A First node
+             * @param B Second node
+             * @param lock_case Lock case for new edge
+             */
+            void generate(size_t A, size_t B, Edge_Lock_Case lock_case)
+            {
+                assert(A != B);
```
jbakosi: Use `Assert`.

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
...
+                key_list[0] = key; // TODO: Is it OK to use copy assignment here?
+
+                key = nodes_to_key(A,C);
+                key_list[1] = key;
+
+                key = nodes_to_key(B,C);
+                key_list[2] = key;
+
+                return key_list;
+            }
```
jbakosi: I wonder if not generating `A->B`, `B->C`, `C->A` instead will generate the orientation which will result in negative Jacobians.

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
...
+
+            /**
+             * @brief function to take a list of edge and mark them all
+             * as needing to be refined
+             *
+             * @param ids List of ids to mark for refinement
+             */
+            void mark_edges_for_refinement(std::vector<node_pair_t> ids) {
+                for (size_t i = 0; i < ids.size(); i++)
```
jbakosi: This can be done with a range-based for loop to simplify.

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
...
+             */
+            void unmark_for_refinement(edge_t key)
+            {
+                    assert( exists(key) );
+                    get(key).needs_refining = false;
+            }
+
+            // TODO: Document this (and implement!)
+            void mark_edges_for_derefinement(std::vector<node_pair_t> ids) {
+                for (size_t i = 0; i < ids.size(); i++)
```
jbakosi: Also with range-based for loop.

src/Inciter/AMR/edge_store.h
```diff
@@ -0,0 +1,329 @@
...
+             *
+             * @param tet The tet to generate edge pairs for
+             *
+             * @return A list (array) of edge keys which can be separated out to
+             * name the two composing node ids
+             */
+            // TODO: Should this return a pointer/reference?
+            edge_list_t generate_keys(tet_t tet)
+            {
+                // FIXME : Generate these with a (2d) loop and not hard code them?
```
jbakosi: I would leave it as is. Could probably be done with a loop, but at this low level, I would definitely want that as a compile-time unrolled loop instead of a runtime loop (which the compiler can choose not to unroll). We could look into that but AFAIK it is not trivial, see e.g., https://www.codeproject.com/Articles/857354/Compile-Time-Loops-with-Cplusplus-Creating-a-Gener.

src/Inciter/AMR/id_generator.h
```diff
@@ -0,0 +1,146 @@
...
+             * @param parent_id The id of the parent
+             *
+             * @return The list of children ids
+             */
+            child_id_list_t generate_child_ids(size_t parent_id, size_t count = MAX_CHILDREN)
+            {
+                child_id_list_t c;
+                c.resize(count);
+                c[0] = parent_id; // TODO: Remove this. Horrific hack to suppress warning
+                for (size_t i = 0; i < count; i++)
```
jbakosi:
```
for (auto& i : c) i = get_next_tet_id();
```

src/Inciter/AMR/id_generator.h
```diff
@@ -0,0 +1,146 @@
...
+            // (64 - 20) / 3  = 14
+
+            // This basically says the number of tets which can be in an initial grid
+            // A sensible value is 2^20 (1,048,576) for big simulations, and anything
+            // smaller for toy problems
+            size_t START_TET_ID = 1024;
+
+            // Constructor to reset START_TET_ID on the new value
+            morton_id_generator_t() : id_generator_t() {
+                // TODO: Is there a nice way to remove this code duplication
```
jbakosi: How about
```
class id_generator_t {
    private:
        std::size_t START_TET_ID;
        std::size_t start_id;
        std::size_t next_tet_id;

    public:
        // Constructor
        explicit id_generator_t( std::size_t start_tet_id = 0 ) :
            START_TET_ID( start_tet_id ),
            start_id( start_tet_id ),
            next_tet_id( start_tet_id )
        {}
};

class morton_id_generator_t : public id_generator_t
{
    public:
        morton_id_generator_t() : id_generator_t(1024) {}
};
```

src/Inciter/AMR/id_generator.h
```diff
@@ -0,0 +1,146 @@
...
+             * @brief Helper function to convert from parent id to (base) child
+             * id
+             *
+             * @param parent_id Id of parent
+             *
+             * @return Id of child
+             */
+            static size_t get_child_id(size_t parent_id)
+            {
+                size_t child_id = (parent_id << ID_SHIFT);
```
jbakosi: I would just do
```cxx
return parent_id << ID_SHIFT;
```

src/Inciter/AMR/id_generator.h
```diff
@@ -0,0 +1,146 @@
...
+             * @return The list of children ids
+             */
+            static child_id_list_t generate_child_ids(size_t parent_id, size_t count = MAX_CHILDREN)
+            {
+                child_id_list_t c;
+                c.resize(count);
+                for (size_t i = 0; i < count; i++)
+                {
+                    c[i] = get_child_id(parent_id, i);
+                }
```
jbakosi: I would do it with a range-based for loop. See https://github.com/quinoacomputing/quinoa/pull/156/files#r131712662.

src/Inciter/AMR/id_generator.h
```diff
@@ -0,0 +1,146 @@
...
+             *
+             * @param parent_id id of the parent
+             * @param offset offset into the parent list (i.e number of child)
+             *
+             * @return the id of the given child
+             */
+            static size_t get_child_id(size_t parent_id, size_t offset)
+            {
+                // Try detect overflow
+                assert( parent_id <= get_parent_id(std::numeric_limits<size_t>::max()));
```
jbakosi: Even if the assert passes, the return statement may still overflow. Correct?

src/Inciter/AMR/marked_refinements_store.h
```diff
@@ -0,0 +1,113 @@
...
+#define AMR_marked_refinements_store_h
+
+#include <map>
+#include "Refinement_State.h"
+
+namespace AMR {
+
+    class marked_refinements_store_t {
+        private:
+            std::map<size_t, Refinement_Case> marked_refinements;
```
jbakosi: In a second round, we should probably test if we can get away with `std::unordered_map`.

src/Inciter/AMR/marked_refinements_store.h
```diff
@@ -0,0 +1,113 @@
...
+             * @return The marked refinement case for the given tet
+             */
+            Refinement_Case& get(size_t id)
+            {
+                return marked_refinements.at(id);
+            }
+
+            void erase(size_t id)
+            {
+                marked_refinements[id] = Refinement_Case::none;
```
jbakosi: Interesting. I wonder what is the difference between `Refinement_Case::none` and a nonexistent `id`.

src/Inciter/AMR/marked_refinements_store.h
```diff
@@ -0,0 +1,113 @@
...
+             * @param id The id of the tet to mark
+             * @param r The refinement decision for the tet
+             */
+            void add(size_t id, Refinement_Case r)
+            {
+                // Check if that active element already exists
+                if (exists(id))
+                {
+                    if (marked_refinements[id] != r)
+                    {
```
jbakosi: `exists()` already does a search in `marked_refinements`, so let's not repeat it with `operator[]`.

One solution that comes to mind is to return a `boost::optional` from `exists()` instead of a simple `bool`. `boost::optional` can then be queried by its implicit conversion-to-bool operator and if true, the value behind `boost::optional` can be accessed by `boost::optional<T>::operator*()`. As `T` I would simply use `Refinement_Case`, but perhaps as a typesafe enum. `boost::optional` should be immediately available from boost (we already use it) and `std::optional` will come with C++17, which I believe (hope) will not be very different from `boost::optional`.

Also, this way existing client code of `exists()` should not need to change, because of the existence of `boost::optional<T>::operator bool()'.

src/Inciter/AMR/mesh_adapter.cpp
```diff
@@ -0,0 +1,813 @@
...
+        node_store.m_graphsize = graph_size;
+    }
+
+    void mesh_adapter_t::init_with_nodes(coord_type* m_x, coord_type* m_y, coord_type* m_z, size_t* graph_size)
+    {
+        init_node_store(m_x, m_y, m_z, graph_size);
+        //init(); // TODO: This also needs to call init if you want node support to work
+    }
+
+    void mesh_adapter_t::init(std::vector<size_t> m_tetinpoel, size_t num_nodes) {
```
jbakosi: I would not use `m_` for function arguments. This is only used for class member data. Just to keep it consistent with rest of the code.

jbakosi: Also, what is the reason for passing the vector by value? If it is not modified (and it isn't) I would pass it by const-ref: `const std::vector<size_t>&`. If it is to be really "consumed", i.e., moved from, then the way to pass it as an rvalue reference: `std::vector<size_t>&&`. Both of these avoid copying.

src/Inciter/AMR/mesh_adapter.cpp
```diff
@@ -0,0 +1,813 @@
...
+                            //  locked edge and will thus ignore it
+                            if (tet_store.edge_store.get(key).needs_refining)
+                            {
+                                num_to_refine++;
+                                //trace_out << "key needs ref " << key << std::endl;
+                            }
+                        }
+                    }
+
+                    // TODO: Should this be a reference?
```
jbakosi: Probably okay, it is a simple enum, cheap to copy.

src/Inciter/AMR/types.h
```diff
@@ -0,0 +1,48 @@
...
+const size_t NUM_TET_FACES = 4;
+const size_t ID_SHIFT = 3;
+const size_t MAX_CHILDREN = 8;
+const char KEY_DELIM = '-';
+
+using real_t = tk::real;
+using coordinate_t = std::array<real_t, DIMENSION>;
+using tet_t = std::array<size_t, NUM_TET_NODES>;
+
+using node_pair_t  = std::array<std::size_t, 2>; // TODO: should this be a std::pair
```
jbakosi: Not too bad as `std::array` either, because then the elements can be accessed as `[0]` and `[1]` instead of the more verbose `.first` and `.second`.

src/Inciter/AMR/node_connectivity.h
```diff
@@ -0,0 +1,165 @@
...
+                for (size_t i = 0; i < size(); i ++)
+                {
+                    std::cout << i << ": A " << get(i)[0] << " B " << get(i)[1] << std::endl;
+                }
+            }
+
+    };
+}
+
+#endif // guard
```
jbakosi: This class seems like an edge (connectivity) for me. Is that correct? Even if that is eventually needed as such it is probably store the state `nodes` as a std::vector< std::size_t > as that will guarantee contiguous memory storage but `std::vector< std::array< std::size_t, 2 > >` does not. The access logic addressing into it should be pretty straightforward to change I think.

But I also wonder if this is at all will be necessary and we could use something building on [UnsMesh::Edge](https://github.com/quinoacomputing/quinoa/blob/master/src/Mesh/UnsMesh.h#L31) and related derived types following.

src/Inciter/AMR/node_store.h
```diff
@@ -0,0 +1,245 @@
...
+#include <cmath>
+
+#include "types.h"
+#include "tet_store.h"
+
+// TODO: make this have a base class to support multiple generator schemes
+// using the policy design pattern
+namespace AMR {
+
+    class node_store_t
```
jbakosi: I wonder how this might be possible to merge to the existing [UnsMesh](https://github.com/quinoacomputing/quinoa/blob/master/src/Mesh/UnsMesh.h) class.

src/Inciter/AMR/node_store.h
```diff
@@ -0,0 +1,245 @@
...
+    {
+
+        private:
+            coord_type* m_x;
```
jbakosi: If we keep this as a separate "node store". it is probably good practice not store any state by reference (or pointers) as that, in general, will be more problematic with migration. (Value-types are simpler and less error prone to pup.)

src/Inciter/AMR/node_store.h
```diff
@@ -0,0 +1,245 @@
...
+            coord_type* m_z;
+
+            // We really don't want people to pass this by value..
+            // (Because we store refs in here, which are consts..)
+            // NonCopyable & operator=(const NonCopyable&) = delete;
+            node_store_t(const node_store_t& c) = delete;
+            node_store_t& operator=(const node_store_t&) = delete;
+
+        public:
+            node_store_t() { } // default cons
```
jbakosi: ... and the default constructor leaving the state pointers uninitialized: a recipe for disaster. ;-)

src/Inciter/AMR/node_store.h
```diff
@@ -0,0 +1,245 @@
...
+             * @param x_in X coord to check
+             * @param y_in Y coord to check
+             * @param z_in Z coord to check
+             *
+             * @return The id of the node if it exists, -1 if it doesn't.
+             */
+            int check_node_exists(real_t x_in, real_t y_in, real_t z_in)
+            {
+                const real_t eps = 1e-7;
+                for (size_t i = 0; i < size(); i++)
```
jbakosi: Yep, as the comments above indicate, this is a linear search and we probably want steer clear of it.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+
+// TODO: make this have a base class to support multiple generator schemes
+// using the policy design pattern
+
+namespace AMR {
+
+    class refinement_t {
+        private:
+
+            const size_t DEFAULT_REFINEMENT_LEVEL = 0; //TODO: Is this in the right place?
```
jbakosi: We will read this out from user input. I will help (or will do it) in a second pass.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+            const size_t MIN_REFINEMENT_LEVEL = DEFAULT_REFINEMENT_LEVEL;
+
+            // TODO: Tidy up edge store references here
+            tet_store_t* tet_store;
+            node_connectivity_t* node_connectivity;
```
jbakosi: Just a note: try not to use reference semantics in class state - will be more complex and/or problematic with migration.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+
+            /**
+             * @brief Method which takes a tet id, and deduces the other
+             * parameters needed to perform a 1:2
+             *
+             * @param tet_id The id to refine 1:2
+             */
+            void refine_one_to_two(size_t tet_id)
+            {
+                edge_list_t edge_list = tet_store->generate_edge_keys(tet_id);
```
jbakosi: Could use `auto` instead of `edge_list_t`, though some might prefer to see the explicit type.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+                node_pair_t nodes = find_single_refinement_nodes(edge_list);
```
jbakosi: Could use `auto` instead of `node_pair_t`, though some might prefer to see the explicit type.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+             * into the form needed for the main 1:2 refinement method
+             *
+             * @param tet_id The id to refine 1:2
+             */
+            void refine_one_to_two(
+                    size_t tet_id,
+                    std::string edge_key
+            )
+            {
+                std::vector<std::string> nodes = util::split(edge_key,KEY_DELIM);
```
jbakosi: I think it would be better to use two `int`s as the edge-end points, in edge_key, instead of a string, e.g., `std::array< std::size_t, 2 >`, as in [UnsMesh::Edge](https://github.com/quinoacomputing/quinoa/blob/master/src/Mesh/UnsMesh.h#L32), which will eliminate the non-type-safe stuff behind `util::split()`.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+                /// Split existing tet into two new tets
+
+                // The two new tets will be the same, but for each an edge will
+                // be cut, losing an edge  replaced by E
+
+                tet_t new_tet1;
+                tet_t new_tet2;
+
+                // Create a new tet that is based on the original
+                copy_tet(&new_tet1, &original_tet);
```
jbakosi: This seems like it would read more naturally with a copy assignment instead. Since `tet_t` is a simple `std::array< std::size_t, 4 >`, this should be done by _not_ defining the copy assignment and the compiler will generate the correct one doing deep-copy as required.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+
+                // Replace all node ids in tet that were pointing to A with new_node_id
+                std::replace(new_tet1.begin(), new_tet1.end(), edge_node_A_id,
```
jbakosi: I like it that you are using a standard algorithm. However, in this case, I think it should probably wrapped into a one-liner function with a great name for better readability.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+
+                if (!check_allowed_refinement(tet_id)) return;
+
+                // Split every edge into two
+                // Makes 4 tets out of the old corners and 3 near mid-points
+                // Make 4 out of the midpoints
+
+                // For Tet {ABCD} need to know all (non-repeating) node pairs
+                // {AB} {AC} {AD} {BC} {BD} {CD}
+                // This can either be hard coded, or generated with a 2d loop
```
jbakosi: Probably better and simpler to just keep it unrolled as it is. See also https://github.com/quinoacomputing/quinoa/pull/156/files#r131709593.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+             * you want to make a tet that's very similar to an existing one
+             *
+             * @param out The tet to store the copy
+             * @param original The tet to copy the data from
+             */
+            void copy_tet(tet_t* out, tet_t* original)
+            {
+                // NOTE: This will do a deep copy, so is safer than it may look
+                *out = *original;
+            }
```
jbakosi: I think the best is, well, to not do anything. ;-) And let the compiler auto-generate the copy constructor, copy assignment, as well as the move constructor and the move assignment for `std::array<>`.

src/Inciter/AMR/refinement.h
```diff
@@ -0,0 +1,779 @@
...
+             *
+             * @param tet tet of the tet to look for
+             * @param element offset into that tet to look at
+             *
+             * @return tet into data arrays the tet lives
+             */
+            // TODO: Move this (or rename?)
+            size_t tet_id_to_node_id(size_t tet, size_t element) {
+                return tet_store->get(tet)[element];
+            }
```
jbakosi: It would be good to put in an `Assert( element < NUM_TET_NODES, "Indexing out of tet )";` before `return`.

Also, just to double-check: is it the desired behavior that if `tet` does not exist in tet_store an exception is thrown? (I'm not sure, but you will know.)

src/Inciter/AMR/tet_store.h
```diff
@@ -0,0 +1,733 @@
...
+             * @brief Getter for tet element
+             *
+             * @param id Id of tet to get
+             *
+             * @return Copy of the tet
+             */
+            tet_t get( size_t id )
+            {
+                assert( exists(id) );
+                return tets.at(id);
```
jbakosi: `.at()` will throw if `id` does not exist.

src/Inciter/AMR/tet_store.h
```diff
@@ -0,0 +1,733 @@
...
+
+                // Sanity check the storage ids
+                // (this is probably better in a function/2d loop)
+                assert( nodes[0] != nodes[1] );
+                assert( nodes[0] != nodes[2] );
+                assert( nodes[0] != nodes[3] );
+                assert( nodes[1] != nodes[2] );
+                assert( nodes[1] != nodes[3] );
+                assert( nodes[2] != nodes[3] );
+            }
```
jbakosi: Yeah. That will get compiled away in `RELEASE` even in a runtime for loop.

src/Inciter/AMR/tet_store.h
```diff
@@ -0,0 +1,733 @@
...
+             * derefinement)
+             *
+             * @param id id of the tet to delete
+             */
+            void erase(size_t id)
+            {
+                deactivate(id);
+                master_elements.erase(id);
+                tets.erase(id);
+                // TODO: Should this update the number of children here rather than at the call site?
```
jbakosi: I think yes if the number of children is stored here, explicitly or inside a data structure that is stored here, i.e., in this class' state. No?

src/Inciter/AMR/tet_store.h
```diff
@@ -0,0 +1,733 @@
...
+                if (!is_active(id) )
+                {
+                    active_elements.add(id);
+                }
+            }
+
+            /**
+             * @brief Function to add a tet to a list which maintains what is a
+             * center tet. (The need to maintain a list could be replaced by a
+             * geometric check on the tet itself)
```
jbakosi: But that would necessitate dealing with node coordinates, correct?

src/Inciter/AMR/tet_store.h
```diff
@@ -0,0 +1,733 @@
...
+                    }
+                }
+
+                trace_out << "in size " << in.size() << " active size " << active.size() << std::endl;
+                return active;
+            }
+
+            // TODO: These mark methods can probably be a single one to which a
+                // Refinement_Case is passed, depending on how extra edges are marked
+            /**
```
jbakosi: Yeah, it seems like it would be good to do that to simplify and shorten the code follows.

src/Inciter/AMR/util.cpp
```diff
@@ -0,0 +1,108 @@
...
+        {
+            coordinate_t mid_point;
+
+            for(size_t i = 0; i < DIMENSION; i++)
+            {
+                mid_point[i] = (edge_node_A[i]+edge_node_B[i])/2.0;
+            }
+
+            return mid_point;
+        }
```
jbakosi: This is probably going away.

src/Inciter/AMR/util.cpp
```diff
@@ -0,0 +1,108 @@
...
+        {
+
+            coordinate_t mid_point;
+
+            mid_point[0] = (x1+x2)/2.0;
+            mid_point[1] = (y1+y2)/2.0;
+            mid_point[2] = (z1+z2)/2.0;
+
+            return mid_point;
+        }
```
jbakosi: Going away soon.

src/Inciter/AMR/id_generator.h
rfbird: What does this look like when I want an index (`i`) for the `get_child_id` call?

```
  for (auto& i : c) 
  { 
     i = get_child_id(parent_id, some_counter?)
  }
```

src/Inciter/AMR/marked_refinements_store.h
rfbird: Pretty sure we can, I'll add a TODO

src/Inciter/AMR/marked_refinements_store.h
rfbird: Good catch

src/Inciter/AMR/marked_refinements_store.h
rfbird: Nice idea, but to me it very much feels like a premature "optimization". I think we would do better to just have stronger semantics around what is stored. 

At a minimum, we could just probably remove the `exists` call, or replace it with a `find` call.  

I think we can punt on this, and then if we decide to address the global nature of `refinement_state_changed` (imo a bigger problem), this will be natural to address then

src/Inciter/AMR/node_connectivity.h
rfbird: I agree this class needs some love. Yes this is meant to be connectivity, and is basically what was left once I tried to strip any sense of coordinates out into an optional store.

I'm not sure this is the place to go in for a drastic re-write, but i'll put it on my list of things that need to be looked at (perhaps a issue is more appropriate?)

rfbird: On second thoughts, given the inner type is not a point, I think the data should (now) be contiguous?: 

> 23.2.6 Class template vector [vector]
> 
> 1 A vector is a sequence container that supports random access iterators. In addition, it supports (amortized) constant time insert and erase operations at the end; insert and erase in the middle take linear time. Storage management is handled automatically, though hints can be given to improve efficiency. The elements of a vector are stored contiguously, meaning that if v is a vector where T is some type other than bool, then it obeys the identity &v[n] == &v[0] + n for all 0 <= n < v.size().

src/Inciter/AMR/node_store.h
rfbird: Yes, lets discuss this tomorrow. Originally I had them merged, but wanted to try and find a discrete way to separate them as it's not clear if everywhere we need AMR (i.e Partitioner.h) we will have a mesh_t

src/Inciter/AMR/node_store.h
rfbird: See above. This was done like this a temporary measure to soft-wrap mesh_t, but to provide a (partial) abstraction from some of it's implementation details. It's definitely not the best solution ever!

src/Inciter/AMR/node_store.h
rfbird: Great catch. This is a hang over from me preferring stack level storage for `node_store` in the `mesh_adapter`. Perhaps it's time to change that?

src/Inciter/AMR/types.h
rfbird: OK

src/Inciter/AMR/Refinement_State.h
jbakosi: On second thought, I think it is actually a good idea to have a separate types header for AMR. Let's call it `AMRTypes.h` though, or something similar.

src/Inciter/AMR/edge_store.h
jbakosi: Okay.

src/Inciter/AMR/node_connectivity.h
jbakosi: Yeah. `std::vector<T>` itself is guaranteed to be contiguously storing `T`. But if `T` is a container with its own heap-based storage, it will not be contiguous anymore, e.g., a vec of vecs. Having said that a vec of arrays might be. I'm actually not sure. But converting to just a vec of Ts will definitely be contiguous.

This is actually, not a big deal at all at this point; just wanted point out, so that we think about and perhaps put it on a todo list. Most important at this point is to put merge this in if it is correct. We can continue cleaning this up in following passes.

src/Inciter/AMR/Refinement_State.h
rfbird: OK. It's already in the AMR folder, so the include would be pretty transparent (`AMR/types.h` vs `AMR/AMR_types.h`), but sure. It's easy enough to change

src/Inciter/AMR/node_connectivity.h
jbakosi: Okay. Cool. I did not know that.

src/Inciter/AMR/mesh_adapter.cpp
rfbird: Agreed

src/Inciter/AMR/Refinement_State.h
jbakosi: Okay. If we do _not_ include `src/Inciter/AMR` among the list of include dirs in cmake, then one will be forced to do the include as `AMR/types.h' - that's cool.

src/Inciter/AMR/refinement.h
rfbird: Sounds good.

src/Inciter/AMR/refinement.h
rfbird: I personally prefer the explicit semantic, it helps me know how to use the type

src/Inciter/AMR/refinement.h
rfbird: I personally prefer the explicit semantic, it helps me know how to use the type

src/Inciter/AMR/refinement.h
jbakosi: Totally understandable and I'm cool with that.

src/Inciter/AMR/refinement.h
jbakosi: Okay.

src/Inciter/AMR/refinement.h
rfbird: Pretty sure this is a hangover from an old version which used strings and can be safely removed.

jbakosi: Great!

src/Inciter/AMR/refinement.h
rfbird: I agree, and I think that's what I had originally. 

I think I then went on to decide the above is more obviously a deep copy / duplication, in an attempt to be verbose and understandable, given I think the performance will be essentially identical

src/Inciter/AMR/refinement.h
rfbird: heh, `replace_node` already exists.. better use it then

src/Inciter/AMR/refinement.h
rfbird: That sounds good. Right now If the user (i.e we) do something stupid, I want to know about it as soon as possible..

src/Inciter/AMR/tet_store.h
rfbird: Swapping to `[]`

src/Inciter/AMR/refinement.h
jbakosi: Understood.

src/Inciter/AMR/tet_store.h
rfbird: Agreed. I still suspect there is a better way to do this

src/Inciter/AMR/util.cpp
rfbird: I'm not sure what this means, but I'm guessing it's not too important. 

We should likely merge with your code to calculate mid_points or similar, perhaps it's that?

rfbird:
We should see if there is a meaningful way to port over the AMR unit tests @jbakosi

src/Inciter/AMR/util.cpp
jbakosi: No, I think I meant that you will take out coordinates dependence.

jbakosi:
There are *no* AMR unit tests. ;-) Only two regression tests. We should add some unit tests (for util-like functionality), and definitely more regression tests. I can also do that (and/or help with them).

src/Inciter/AMR/util.cpp
rfbird: Ah, OK. Somewhere has to track the coordinates of where the new AMR nodes are added, and the AMR lib isn't an unreasonable place to do this [assume geom tracking is enabled]. If that is true, then this kind of functionality doesn't fundamentally not make sense

jbakosi: Yeah. I'm not sure where is the best to track coordinates. It is appealing to have AMR not know anything about coordinates, but then eventually new edge-nodes have to have new coordinates associated with them. We can think about this more...

jbakosi:
12? I can count only 3, and those are really the same test, but run in serial, parallel _without_ virtualization, and parallel _with_ virtualization.

And I don't know how to _reply_ either, unless it is a _thread_ started on a line of code.

src/Inciter/AMR/Refinement_State.h
jbakosi: That's what thought you meant though. So what is the reason for _not_ making this 'explicit'.

src/Inciter/AMR/Refinement_State.h
jbakosi: Sure, but everything is public...

src/Inciter/AMR/edge.cpp
jbakosi: You mean C (as C and not C++) and CPP (as really C++)? If so, I thought for C one would use `.c`, i.e. lower case.

src/Inciter/AMR/edge.h
jbakosi: Uh! I just realize that `edge_` is a type... Okay, for that I guess you have consistently used the `_t` suffix and I use capital first character. Then why don't you make it `edge_t`?

src/Inciter/AMR/id_generator.h
jbakosi: And this way it will not be on the stack? I'm not sure if I understand? And yes, the magic 1024 is not great, but that wasn't really my concern here.

src/Inciter/AMR/id_generator.h
jbakosi: Oops! Never mind. I did not notice that you need the loop index. In that case, no range-for.

src/Inciter/AMR/marked_refinements_store.h
jbakosi: Okay, that's fine. You see the more important problem to address from the AMR viewpoint. I would do it with a find and an if test then for now.

jbakosi:
As it turned out, it was _I_ who was getting tired: I thought you were talking about the `quinoa`-amr regression tests, which are 3, but you were talking about the `tetamr` regression tests, which I guess are 12. Yes, we should definitely port those over.

PR #176, opened 2017-10-02
Cleanup

jbakosi:
This PR is generally a cleanup and some simplification of the classes in inciter. Partly in preparation for DG, and partly to prepare for solving the PDE system using lumped mass as an option. This seems like a good point for sanity check. Next on my list is making DG and DG selectable based on user input.

For more info, see also the commit messages and issues in the *Done* column of project [Prep4DG](https://github.com/quinoacomputing/quinoa/projects/2).


src/Control/Inciter/InputDeck/InputDeck.h
```diff
@@ -75,8 +75,6 @@ class InputDeck :
                                        kw::pde_diffusivity,
                                        kw::pde_lambda,
                                        kw::pde_u0,
-                                       kw::poisson,
-                                       kw::dir_neu,
```
rfbird: If these are particle specific (and that's why they were removed), is there a nice logical way to group some of these things together?

jbakosi: They are not particle-related. The history of this is that (around a year ago) I wanted to do some research in algebraic multi-grid on top of Charm++, i.e., distributed-memory, asynchronous, AMG and a Poisson equation with inhomogeneous Dirichlet and Neumann boundary conditions (for which I know the analytical solution) would have been a good test problem. I just remove it now, because this is not going to happen in the next couple of years at least.

rfbird:
I know this is always tricky, but I think this PR would have been much cleaner split out into multiple. One for disabling the particles, one for renaming solver, etc

jbakosi:
Yeah, true. I will do it better next time.

src/Inciter/Partitioner.h
```diff
@@ -281,11 +273,13 @@ class Partitioner : public CBase_Partitioner< HostProxy,
...
         inpoel.insert( end(inpoel), begin(c.second), end(c.second) );
       }
-      Group::thisProxy[ frompe ].recv();
+      thisProxy[ frompe ].recv();
     }
 
     //! Acknowledge received node IDs
-    void recv() { if (--m_npe == 0) signal2host_distributed( m_host ); }
+    void recv() {
+      if (--m_npe == 0) contribute( m_cb.get< tag::distributed >() );
```
rfbird: I'm  a really big fan of adding some whitespace to the code, I feel it's pretty dense in places right now :)

PR #180, opened 2017-10-16
CG/DG dispatch

jbakosi:
For details see the commit message for ebaf3eb.


rfbird:
We should keep on eye on the C++17 stuff like constexpr if (https://hackernoon.com/a-tour-of-c-17-if-constexpr-3ea62f62ff65) which should(?) let us simply some of this stuff, removing SFINAE etc

PR #181, opened 2017-10-23
Scheme::PUP + enable DG

jbakosi:
This adds proper Charm++ pack/unpack routines for `inciter::Scheme` together with unit tests.

Note that this now also requires C++14. This is for type-based access to a tuple element using `std::get<T>(std::tuple)`. I'm sure it would be possible to do this based on a compile-time constant using some template magic, which only requires C++11, but there is no real reason to keep the language standard requirement to be C++11 and I'm kinda sick of putting in a lot of template magic. 

Inside in`tk::Variant`, which is the real contribution of this PR, accessing tuple elements really is the simplest using type-based access.

Also our currently tested compilers (clang, gnu, intel) all support C++14 for quite a while now. This will also enable some additional simplifications, such as `std::make_unique` instead of `tk::make_unique` and `auto` as return type.

Objections?


jbakosi:
Ah! I really wanted to create a new PR for the above 3 commits. I thought I would be able to create a new PR even using the same branch. Anyway, it this way now. @junghans and @rfbird: I think we should just do a review in person where I will explain to make it easier.

rfbird:
OK, we can go over this on Wednesday?

---
Robert Bird
CCS-7, Mail Stop B287
Los Alamos National Laboratory
Los Alamos, NM 87545

Phone: +1-505-665-7374

src/Base/Variant.h
```diff
@@ -0,0 +1,109 @@
...
+    friend void operator|( PUP::er& p, Variant< Types... >&& v ) { v.pup(p); }
+    //@}
+
+  private:
+    //! Visitor setting a value of tuple that matches the type of the variant
+    struct getval : boost::static_visitor<> {
+      Variant* const host;
+      getval( Variant* const h ) : host(h) {}
+      template< typename P > void operator()( const P& p ) const {
+        std::get< P >( host->tuple ) = p;
```
rfbird: Do you think we should be documenting which code features require us to be upping our C++ language version?

src/Base/Variant.h
```diff
@@ -0,0 +1,109 @@
...
+
+  public:
+    //! Default constructor for Charm++ (e.g., PUP's receive side)
+    Variant() {}
+
+    //! Initializer constructor
+    //! \param[in] v boost::variant used as input/output
+    //! \details This constructor queries the index at which the variant holds a
+    //!   value and sets that value in tuple mathing the type.
+    Variant( boost::variant< Types... >& v ) : idx( v.which() ), variant(v) {
```
rfbird: Do you think it's worth abstracting (perhaps via a typedef?) to hide the implementation of the detail, to avoid explicitly stating it's a dependency on boost? Not sure if there's a compelling reason for an implementation detail like this to "leak out". When we switch to std::variant it may also make that easier

rfbird:
Just added a few comments I had from looking at the code in the meeting.

I'm not sure where best to put this last comment, but I just wanted to make sure we're pretty clear (in comments in the code) about this new reduction you added between the setup and the stepping being more of a cosmetic fix-up to make charm work than something which is (should be?) functionally required

src/Base/Variant.h
jbakosi: That can be done, but I don't feel the strong need for it in this case. Though, I may not foresee enough at this point too much of a problem with this. I think at this point this is so little code, that as it is, explicitly stating boost::variant here, probably makes it more readable. Also, grepping for `boost::variant` will make switching to std::variant easy. AmI missing something?

jbakosi:
> I'm not sure where best to put this last comment, but I just wanted to make sure we're pretty clear (in comments in the code) about this new reduction you added between the setup and the stepping being more of a cosmetic fix-up to make charm work than something which is (should be?) functionally required

Good point. Yeah, I'm planning to document this stuff a little better than it is right now before merging this PR. The problem is that this stuff is so trial&error that I never feel like I 100% know what I'm doing even when it finally works (or I think it does). Still, putting in some comments about why *I think* some of these reductions are needed may still be better than nothing, s I will do it. This is so fishy, I'm very dissatisfied with it.

rfbird:
For some reason I can't see these comments on GH, so hopefully replying over email is sufficient.

I agree, it's an easy change. In general I think it's a good idea, but perhaps in this case there is indeed no need

---
Robert Bird
CCS-7, Mail Stop B287
Los Alamos National Laboratory
Los Alamos, NM 87545

Phone: +1-505-665-7374

rfbird:
Hmm yea, I think you're right. There's no way of doing this that's "nice enough" to be worth the effort

---
Robert Bird
CCS-7, Mail Stop B287
Los Alamos National Laboratory
Los Alamos, NM 87545

Phone: +1-505-665-7374

PR #184, opened 2018-01-04
Lump

jbakosi:
This concludes the work leading up to a functional lumped-mass solution in inciter. In a nutshell, inciter now has three options for selecting a discretization scheme: `matcg`, `diagcg`, and `dg`. The first one solves the consistent-mass matrix using hypre, the second one solves the lumped-mass matrix inside `diagcg` and does not require a linear solver, and work on `dg` has been started but not yet functional. Regression tests have been added and several steps have been taken to revamp the code to reuse functionality that used to be married with the `matcg` solution, namely, the use of FCT as well as applying Dirichlet boundary conditions (BCs), and computing diagnostics, e.g., L2 errors.

Note that the work on BCs simply abstracted the existing functionality away to a common class and did not attempt to make the matching of the BCs general, see also the comments in the newly added `BoundaryConditions::match()` in `Inciter/BoundaryConditions.C`.


src/Inciter/BoundaryConditions.C
```diff
@@ -0,0 +1,192 @@
...
+      auto it = localnodes.find( o );
+      if (it != end(localnodes))
+        n.push_back( it->second );
+    }
+  }
+
+  return sidenodes;
+}
+
+std::unordered_map< std::size_t,
```
rfbird: Does it make sense to tidy up some of these complex types using typedefs/usings?

jbakosi: Yeah, I agree. But I never really know which way to go with this question. On one hand, hiding these complex types behind a good name shortens the code, but also hides the actual type, which then, IMO, decreases readability. I favor slightly more the more detail, partly because I think that's easier to read (though admittedly there is more to read), and partly because this avoids coming up with good names, which is hard, and I'm lazy.

jbakosi:
@adityakpandare: The `sqrt` is are In `Transporter::Diagnostics()`, the reduction target, at the very end of accumulation (across all PEs) of the sums for the various errors: https://github.com/quinoacomputing/quinoa/blob/lump/src/Inciter/Transporter.C#L641.

src/Inciter/DG.h
```diff
@@ -65,6 +68,12 @@ class DG : public CBase_DG {
     //! Total mesh volume
     tk::real m_vol;
 
+    //! Access bound Discretization class pointer
+    Discretization* Disc() const {
+      Assert( m_disc[ thisIndex ].ckLocal() != nullptr, "ckLocal() null" );
+      return m_disc[ thisIndex ].ckLocal();
+    }
+
```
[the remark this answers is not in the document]
jbakosi: Yes, and also to ensure that the `Assert` always gets called (in `DEBUG` mode).

jbakosi:
@adityakpandare, thanks for the review. It is always great to have another (and another) set of eyes. Please approve.

src/Inciter/Diagnostics.C
```diff
@@ -0,0 +1,141 @@
...
+//  Compute diagnostics, e.g., residuals, norms of errors, etc.
+//! \param[in] d Discretization proxy to read from
+// *****************************************************************************
+{
+  // Store the local IDs of those mesh nodes to which we contribute but do not
+  // own, i.e., slave nodes. Ownership here is defined by having a lower chare
+  // ID than any other chare that also contributes to the node.
+  for (const auto& c : d.Msum())        // for all chares that neighbor our mesh
+    if (d.thisIndex > c.first)          // if our chare ID is larger than theirs
+      for (auto i : c.second)           // store local ID in set
```
rfbird: Not that it matters, but should this be const?

src/Inciter/Diagnostics.C
```diff
@@ -0,0 +1,141 @@
...
+Diagnostics::compute( Discretization& d, const tk::Fields& u )
+// *****************************************************************************
+//  Compute diagnostics, e.g., residuals, norms of errors, etc.
+//! \param[in] d Discretization proxy to read from
+//! \param[in] u Current solution vector
+//! \return True if diagnostics have been computed
+// *****************************************************************************
+{
+  // Optionally collect diagnostics and send for aggregation across all workers
+  if ( !((d.It()+1) % g_inputdeck.get< tag::interval, tag::diag >()) ) {
```
rfbird: Would the clarity of this be increased by breaking it out into multiple lines?

src/Inciter/Diagnostics.C
```diff
@@ -0,0 +1,141 @@
...
+    // where a_i and n_i are the analytical and numerical solutions at node i,
+    // respectively. The final aggregated solution will end up in
+    // Transporter::diagnostics().
+
+    // Diagnostics vector (of vectors) during aggregation:
+    // 0: L2-norm of all scalar components of the numerical solution
+    // 1: L2-norm of all scalar components of the numerical-analytic solution
+    // 2: Linf-norm of all scalar components of the numerical-analytic solution
+    std::vector< std::vector< tk::real > >
+      diag( 6, std::vector< tk::real >( u.nprop(), 0.0 ) );
```
rfbird: Again, perhaps good to get rid of the magic number (6)

src/Inciter/Diagnostics.C
```diff
@@ -0,0 +1,141 @@
...
+        for (std::size_t c=0; c<u.nprop(); ++c)
+          diag[1][c] +=
+            (u(i,c,0)-a(i,c,0)) * (u(i,c,0)-a(i,c,0)) * d.Vol()[i];
+        // Compute Linf norm of the numerical-analytic solution
+        for (std::size_t c=0; c<u.nprop(); ++c) {
+          auto err = std::abs( u(i,c,0) - a(i,c,0) );
+          if (err > diag[2][c]) diag[2][c] = err;
+        }
+      }
+
```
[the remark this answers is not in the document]
rfbird: I agree, at a glance it does appear that he's calculating |x|^2?

src/Inciter/Diagnostics.C
```diff
@@ -0,0 +1,141 @@
...
+    // Append diagnostics vector with metadata on the current time step
+    // 3: Current iteration count (only the first entry is used)
+    // 4: Current physical time (only the first entry is used)
+    // 5: Current physical time step size (only the first entry is used)
```
rfbird: Could be an enum or similar, as part of removing the magic numbers?

src/Inciter/Discretization.C
```diff
@@ -69,23 +73,14 @@ Discretization::Discretization(
   m_nbfac( nbfac ),
   m_bface( bface ),
   m_belem( belem ),
-  m_esuel( tk::genEsuelTet( m_inpoel, 
-                            tk::genEsup(m_inpoel,4) ) ),
-  m_ntfac( tk::genNtfac( 4,
```
rfbird: Magic number (I think I'm just obsessing at this point... :P) ?

src/Inciter/Scheme.h
```diff
@@ -231,9 +231,9 @@ class Scheme : public SchemeBase {
     //!   specifying a non-default last argument.
     template< typename Op, typename... Args, typename std::enable_if<
       std::is_same< Op, tag::elem >::value, int >::type = 1,
-      typename std::enable_if< sizeof...(Args) == 11, int >::type = 0 >
+      typename std::enable_if< sizeof...(Args) == 12, int >::type = 0 >
```
rfbird: Changing these doesnt seem like much fun!

src/Inciter/Transporter.C
```diff
@@ -155,83 +157,50 @@ Transporter::Transporter() :
...
-  auto ss = er.readSidesets();
+  auto sidenodes = er.readSidesets();
 
   // Verify that side sets to which boundary conditions are assigned by user
   // exist in mesh file
   std::unordered_set< int > conf;
   for (const auto& eq : g_pdes) eq.side( conf );
   for (auto i : conf)
-    if (ss.find(i) == end(ss)) {
+    if (sidenodes.find(i) == end(sidenodes)) {
```
rfbird: Big fan of using the longer but more readable variable name 👍

src/Control/Inciter/Options/Scheme.h
```diff
@@ -20,8 +20,9 @@ namespace inciter {
 namespace ctr {
 
 //! Scheme types
-enum class SchemeType : uint8_t { CG=0,
-                                  DG };
+enum class SchemeType : uint8_t { MatCG=0
```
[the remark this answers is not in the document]
jbakosi: Yeah, no good reason. I will just remove it.

src/IO/RootMeshWriter.C
```diff
@@ -2,7 +2,7 @@
 /*!
   \file      src/IO/RootMeshWriter.C
   \author    A. Pakki
```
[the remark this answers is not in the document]
jbakosi: Indeed.

src/Inciter/BoundaryConditions.C
```diff
@@ -0,0 +1,192 @@
...
+    std::size_t c = 0;
+    for (std::size_t eq=0; eq<g_pdes.size(); ++eq) {
+      auto eqbc = g_pdes[eq].dirbc( t, dt, s, coord );
+      for (const auto& n : eqbc) {
+        auto id = n.first;                      // BC node ID
+        const auto& bcs = n.second;             // BCs
+        auto& nodebc = dirbc[ gid[id] ];        // BCs to be set for node
+        if (nodebc.size() > c) {        // node already has BCs from this PDE
+          Assert( nodebc.size() == c+bcs.size(), "Size mismatch" );
+          for (std::size_t i=0; i<bcs.size(); i++)
```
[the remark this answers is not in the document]
jbakosi: Yeah, you are right. I have added the braces for the `for` loop.

A style guide is not a bad idea. It would also be good to enforce this automatically. Can this be done automatically with a customized `clang-format` style in a way that does not result in a huge difference compared to the current formatting?

src/Inciter/DG.C
```diff
@@ -125,19 +124,19 @@ DG::writeFields( tk::real )
...
 {
-  auto d = m_disc[ thisIndex ].ckLocal();
-  Assert( d!=nullptr, "Discretization proxy's ckLocal() null" );
+  auto d = Disc();
 
   const auto ncomp = g_inputdeck.get< tag::component >().nprop();
 
   std::vector< std::vector< tk::real > >
-    diag( 3, std::vector< tk::real >( ncomp, 0.0 ) );
+    diag( 6, std::vector< tk::real >( ncomp, 0.0 ) );
```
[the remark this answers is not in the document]
jbakosi: You got it: d5dee471

src/Inciter/Diagnostics.C
jbakosi: You mean variable `i`? If so, it is a copy, so it does not really matter, right?

src/Inciter/Diagnostics.C
jbakosi: Sure, see d5dee471 .

src/Inciter/Diagnostics.C
jbakosi: Probably. I would do something like:
```
// Optionally collect diagnostics and send for aggregation across all workers
auto diag = g_inputdeck.get< tag::interval, tag::diag >();
if ( !((d.It()+1) % diag) ) {
```
What do you think?

src/Inciter/Diagnostics.C
jbakosi: Sure: 7f06ea79

jbakosi:
@rfbird and @adityakpandare re:
> I agree, at a glance it does appear that he's calculating |x|^2?

I have added some more meaningful comments in 7f06ea79 that clarify that at that point it is only a sum/max as the first step in computing the norms.

src/Inciter/Diagnostics.C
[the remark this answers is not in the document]
jbakosi: I have added some more meaningful comments in 7f06ea79 that clarify that at that point it is only a sum/max as the first step in computing the norms.

jbakosi:
Alright, @rfbird and @adityakpandare. Thanks for the review and the comments. I think with the exception of https://github.com/quinoacomputing/quinoa/pull/184#discussion_r162161219, I addressed all issues raised from https://github.com/quinoacomputing/quinoa/pull/184/commits/2734eb86226b8471945bbd1e7e194fdcf31fa8ae to https://github.com/quinoacomputing/quinoa/pull/184/commits/7f06ea7977f70a2155ae13c97f415bb3bc3ff24d.

Does anyone know how to force github NOT disappear my earlier comments? This is highly annoying...

src/Inciter/Scheme.h
jbakosi: Indeed. This is something I'm not very proud of with the work that went into the `Scheme`-stuff: everything is generic, except for the number of arguments passed.

src/Inciter/Discretization.C
jbakosi: In the end you are right. We should abstract the 4s and 3s away. Here they mean the number of vertices of the cell. I will put this on my list as that will touch a bunch of other places.

src/Inciter/Scheme.h
jbakosi: Now that I think about it more, this could possibly be fixed in some clever way. That `enable_if` is really only there to enable the compiler differentiate between the overloads of which one does not take a pointer at the end of the entry method argument list and one that does. Scheme provides both so client code is simpler. But now I'm thinking for better maintainability and readability, may be we should just require the client code to pass that final pointer always (which of course can be `nullptr`). What do you think?

jbakosi:
@adityakpandare: The sqrt is are In Transporter::Diagnostics(), the reduction target, at the very end of accumulation (across all PEs) of the sums for the various errors: https://github.com/quinoacomputing/quinoa/blob/lump/src/Inciter/Transporter.C#L641.

src/Inciter/DG.h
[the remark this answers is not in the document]
jbakosi: Yes, and also to ensure that the Assert always gets called (in DEBUG mode).

src/Inciter/Diagnostics.C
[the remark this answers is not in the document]
jbakosi: The `sqrt` is in `Transporter::Diagnostics()`, the reduction target, at the very end of accumulation (across all PEs) of the sums for the various errors: https://github.com/quinoacomputing/quinoa/blob/lump/src/Inciter/Transporter.C#L641.

src/Control/Inciter/Options/Scheme.h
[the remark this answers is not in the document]
jbakosi: Yeah, no good reason. I will just remove it.

src/Inciter/BoundaryConditions.C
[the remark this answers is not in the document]
jbakosi: Yeah, you are right. I have added the braces for the for loop.

A style guide is not a bad idea. It would also be good to enforce this automatically. Can this be done automatically with a customized clang-format style in a way that does not result in a huge difference compared to the current formatting?

src/Inciter/DG.C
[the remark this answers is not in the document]
jbakosi: You got it: d5dee471.

src/Inciter/Diagnostics.C
jbakosi: You mean variable i? If so, it is a copy, so it does not really matter, right?

src/Inciter/Diagnostics.C
jbakosi: Sure, see d5dee471.

src/Inciter/Diagnostics.C
jbakosi: Probably. I would do something like:
```
auto diag = g_inputdeck.get< tag::interval, tag::diag >();
if ( !((d.It()+1) % diag) ) {
...
}
```
What do you think?

src/Inciter/Diagnostics.C
jbakosi: Sure: 7f06ea79.

jbakosi: Also, I have added some more meaningful comments in 7f06ea79 that clarify that at that point it is only a sum/max as the first step in computing the norms.

src/Inciter/Discretization.C
jbakosi: In the end you are right. We should abstract the 4s and 3s away. Here they mean the number of vertices of the cell. I will put this on my list as that will touch a bunch of other places.

src/Inciter/Scheme.h
jbakosi: Indeed. This is something I'm not very proud of with the work that went into the Scheme-stuff: everything is generic, except for the number of arguments passed.

Now that I think about it more, this could possibly be fixed in some clever way. That enable_if is really only there to enable the compiler differentiate between the overloads of which one does not take a pointer at the end of the entry method argument list and one that does. `Scheme` provides both so client code is simpler. But now I'm thinking for better maintainability and readability, may be we should just require the client code to pass that final pointer always (which of course can be `nullptr`). What do you think?

src/Inciter/Diagnostics.C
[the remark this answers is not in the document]
jbakosi: I have added some more meaningful comments in 7f06ea79 that clarify that at that point it is only a sum/max as the first step in computing the norms.

src/Inciter/BoundaryConditions.C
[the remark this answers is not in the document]
rfbird: I don't know how to do this, but it may be worth looking into!

src/Inciter/Diagnostics.C
rfbird: Yea, as I said, not important and shouldn't make a difference the compiler, you just always seems very very good at fully qualifying things

src/Inciter/Diagnostics.C
rfbird: I like it more.

I'd personally be more explicit about the condition you're negating. At a glance when I see this i expect something like if `!(12 % 10) => if !(2)` 

If we're trying to see if the % == 0 or similar, I'd prefer to explicitly state it?

src/Inciter/Scheme.h
rfbird: I'm not sure I fully understand this. Do you mean we have two methods, one which takes a pointer, and one which does not. We should change the interface to always pass a pointer, which can be nullptr in the instance we don't actually need it? I'm OK with that, but don't immediately see that it's definitely "nicer"

src/Inciter/Diagnostics.C
jbakosi: Okay, sure. There is another place also where I have done this. I will fix this up tomorrow. Thx.

src/Inciter/Scheme.h
jbakosi: Yeah, we have two overloads for a single Charm++ entry method: one takes an extra optional argument which is a pointer to an options-struct and one for which these extra options are not passed, so Charm++ just uses the default. I wanted to mirror the Charm++ API with it. But if I remove the overload (and leave the one with the pointer to allow for the options), I believe the `enable_if` can go away. In that case client code will always have to pass a pointer, which can be `nullptr`.

Example client code:
```
m_scheme.discInsert< tag::elem >( ..., nullptr );  // passing the extra options-pointer as nullptr (this is how it wil have to be if we remove one overload, which removes the need for enable_if)
m_scheme.discInsert< tag::elem >( ... );    // not passing the extra options-pointer (works now)
```

src/Inciter/Diagnostics.C
jbakosi: Fixed: 26ac6821

jbakosi: Fixed: 26ac6821

PR #188, opened 2018-01-18
AMR error indicators

jbakosi:
Add two error indicators for AMR and the routines for computing solution gradients with unit tests.

Also add parsing grammar so that inciter can recognize the different error indicators configured by the user. Example input file block:

```
inciter
  amr
    initial uniform       # default: none
    error hessian       # or jump (default: jump)
  end
end
```

At first, I wanted to add a `gradient` type error indicator as well, but that does not guarantee boundedness [0...1], inclusive, so that is skipped for now. We can implement more indicators later.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/188)


src/Inciter/AMR/Error.C
```diff
@@ -0,0 +1,119 @@
...
+//! \param[in] c Scalar component to compute error of
+//! \param[in] coord Mesh node coordinates
+//! \param[in] inpoel Mesh element connectivity
+//! \param[in] esup Linked lists storing elements surrounding points, see
+//!    tk::genEsup()
+//! \param[in] err AMR Error indicator type
+//! \return Error indicator: a real number between [0...1] inclusive
+// *****************************************************************************
+{
+  if (err == inciter::ctr::AMRErrorType::JUMP)
```
rfbird: Not a big deal, but could we find a way to remove the branching/run time check here?

src/Inciter/AMR/Error.h
```diff
@@ -0,0 +1,55 @@
...
+
+ private:
+   using ncomp_t = kw::ncomp::info::expect::type;
+
+  public:
+    //! Compute error estimate for a scalar quantity
+    tk::real scalar( const tk::Fields& u,
+                     const std::pair< std::size_t, std::size_t >& edge,
+                     ncomp_t c,
+                     const std::array< std::vector< tk::real >, 3 >& coord,
```
rfbird: Do you think this would be cleaner if we used some type defs for some of these input types?

src/Mesh/Gradients.C
```diff
@@ -0,0 +1,165 @@
...
+std::array< tk::real, 3 >
+nodegrad( std::size_t node,
+          const std::array< std::vector< tk::real >, 3 >& coord,
+          const std::vector< std::size_t >& inpoel,
+          const std::pair< std::vector< std::size_t >,
+                           std::vector< std::size_t > >& esup,
+          const tk::Fields& U,
+          ncomp_t c )
+// *****************************************************************************
+//  Compute gradient at a mesh node
```
rfbird: Should we also settle on a style of consistent comments? I feel like most files have the comment before the method signature (my preferred way, it can also be consistent with headers then), but some have it between the definition and the body

src/Mesh/Gradients.C
```diff
@@ -0,0 +1,165 @@
...
+  const auto& x = coord[0];
+  const auto& y = coord[1];
+  const auto& z = coord[2];
+
+  // storage for gradient and volume at the mesh node
+  std::array< tk::real, 3 > g{{ 0.0, 0.0, 0.0 }};
+  tk::real vol = 0.0;
+
+  // loop over cells surrounding mesh node
+  for (auto k=esup.second[node]+1; k<=esup.second[node+1]; ++k) {
```
rfbird: Could we make this linked list indexing more transparent? It's not a huge problem, and I don't know how best to do it, but it seems a little obtuse

src/Mesh/Gradients.C
```diff
@@ -0,0 +1,165 @@
...
+     // access field data for scalar component c at nodes of element
+     auto u = U.extract( c, 0, N );
+
+     // compute nodal volume: every element contributes their volume / 4
+     vol += 5.0*J/120.0;
+
+     // compute gradient over element weighed by cell volume / 4 and sum to node
+     for (std::size_t j=0; j<3; ++j)
+       for (std::size_t i=0; i<4; ++i)
+         g[j] += grad[i][j] * u[i] * 5.0*J/120.0;
```
rfbird: Again I'd prefer braces for things which could easily get complexly nested

src/Inciter/AMR/Error.C
jbakosi: What do you mean? Should it be compile-time instead? Or some other way?

src/Mesh/Gradients.C
jbakosi: In headers doxygen wants the comments before and in the .C files we have the comments section after the signature and before the definition. This is consistent with that right?

src/Mesh/Gradients.C
jbakosi: I would also like that. Any suggestion? I cannot think of anything obvious right away.

src/Mesh/Gradients.C
jbakosi: Will do that. Thanks.

src/Inciter/AMR/Error.C
jbakosi: Okay, We could eliminate the runtime branching here if we trigger off the error indicator type from what the user configured in the input file. I didn't do that originally, because this way this `Error` class is independent of the user input (`g_inputdeck`), but I guess we it is okay to use it.

src/Mesh/Gradients.C
jbakosi: Ah, I think I just came up with a cool idea for this. Thanks for suggesting this! I think one should be able to write a container around `esup` and define iterators, `begin()`, `end()`, etc., so that a range-based `for` loop could be written around it by client code, and the loop variable would return `esup.first[k]`. _That_, in a nice generic fashion, would indeed render a LOT of such code less error-prone a lot nicer looking. I will see what I can do - unless you want to take a crack at it. ;-) Let me know.

jbakosi: Here is one way to get started: https://stackoverflow.com/a/31457319.

PR #192, opened 2018-01-22
Correct ordering of boundary-elements and boundary-face connectivities.

adityakpandare:
This commit has the changes required to get the correct ordering of  
boundary-face-node connectivities and boundary-elements. The
boundary-face-node connectivity vector "triinpoel" is read-in inside
Transporter::createPartitioner(). From here, it is passed into
Partitioner and thereby into Discretization. This vector is used to get
consistent face-numbers in the "inpofa" vector, which is the face-node
connectivity for all faces, boundary and internal, required for flux
computations in DG-type discretizations. The inpofa vector is then used
in Discretization to get the correct boundary-element ID's or
host-elements for the boundary faces, which we call "belem" here. The
belem array is required to compute the fluxes across boundary faces.
This kind of data-storage is necessary when loops over boundary-faces
are desired to be according to boundary-condition flags, or
side-set-IDs.


src/Mesh/DerivedData.C
```diff
@@ -802,6 +804,86 @@ genEsued( const std::vector< std::size_t >& inpoel,
   return std::make_pair( std::move(esued1), std::move(esued2) );
 }
 
+std::size_t
+genNbfacTet( std::size_t tnbfac,
+             const std::vector< std::size_t >& inpoel,
+             const std::vector< std::size_t >& t_triinpoel,
+             std::vector< std::size_t >& triinpoel )
+// *****************************************************************************
+//  Generate the number of boundary-faces on this chare/worker/mesh-partition.
```
jbakosi: I would try to phrase this (brief function title) in a more complete and more generic fashion, something like: _"Generate the number of faces and the triangle face connectivity for a chunk of a full mesh"_.

Then in `\detail`, it would be great to have a specific description on what is being done here. I'll attempt and you correct if I'm wrong: _"This function takes a mesh by its domain-element (tetrahedron-connectivity) in inpoel and a boundary-face (triangle) connectivity in t_triinpoel. Based on these two arrays, it searches for those faces of t_triinpoel that are also in inpoel and as a result it generates (1) the number of boundary faces shared with the mesh in inpoel and (2) the intersection of the triangle element connectivity whose faces are shared with inpoel. An example use case is where t_triinpoel contains the connectivity for the boundary of the full problem/mesh and inpoel contains the connectivity for only a chunk of an already partitioned mesh. This function then intersects t_triinpoel with inpoel and returns only those faces that share nodes with inpoel."_

Also, would it be better to name `t_triinpoel` as `triinpoel_complete` or `all_triinpoel`, something like that? And/or name `triinpoel` as `sub_triinpoel`?

src/Mesh/DerivedData.C
```diff
@@ -802,6 +804,86 @@ genEsued( const std::vector< std::size_t >& inpoel,
...
+
+  Assert( !inpoel.empty(), "Attempt to call genNbfacTet() on empty container" );
+  Assert( !t_triinpoel.empty(), "Attempt to call genNbfacTet() on empty container" );
+  Assert( t_triinpoel.size()/nnpf == tnbfac, 
+                  "Incorrect size of triinpoel in genNbfacTet()" );
+
+  auto nptet = inpoel;
+  auto nptri = t_triinpoel;
+
+  // reduced inpoel and t_triinpoel
```
jbakosi: "Reduced" in what sense? I think the the code itself, `unique`, is a good enough "documentation" here.

src/Mesh/DerivedData.C
```diff
@@ -802,6 +804,86 @@ genEsued( const std::vector< std::size_t >& inpoel,
...
+
+  // getting the nodes of the boundary-faces in this chunk
+  for (auto i : nptri)
+    if (snptet.find(i) != end(snptet))
+      nptri_chunk.push_back(i);
+
+  std::size_t tag, icoun;
+
+  // matching nodes in nptri_chunk with nodes in t_triinpoel to 
+  // get the number of faces in this chunk
```
jbakosi: ... and the triangle connectivity for those faces that are also in `inpoel`.

src/Mesh/DerivedData.C
```diff
@@ -1036,20 +1121,25 @@ std::vector< std::size_t >
...
 //  Generate derived data, points on faces for tetrahedra only
 //! \param[in] ntfac Total number of faces.
 //! \param[in] nbfac Number of boundary faces.
 //! \param[in] inpoel Element-node connectivity.
+//! \param[in] triinpoel Face-node connectivity.
 //! \param[in] esuelTet Elements surrounding elements.
 //! \return Elements surrounding faces.
-//! \details The unsigned integer map gives the elements to the left and to
+//! \details The unsigned integer vector gives the elements to the left and to
 //     the right of each face in the mesh.
```
jbakosi: This `\details` comment really relates to the return vector, correct? If so, let's just move it to after the description of `\return`.

src/Mesh/DerivedData.C
```diff
@@ -1099,7 +1183,89 @@ genInpofaTet( std::size_t ntfac,
...
+                               std::vector< std::size_t > >& esup )
+// *****************************************************************************
+//  Generate derived data, host elements for boundary faces
+//! \param[in] nbfac Number of boundary faces.
+//! \param[in] inpofa Face-node connectivity.
+//! \param[in] esup Elements surrounding points as linked lists, see tk::genEsup
+//! \return Host elements or boundary elements.
+//! \details The unsigned integer vector gives the elements to the left of
+//     each boundary face in the mesh.
+// *****************************************************************************
```
jbakosi: Same as in `genInpofaTet`?

src/Mesh/DerivedData.C
```diff
@@ -1099,7 +1183,89 @@ genInpofaTet( std::size_t ntfac,
...
   return inpofa;
 }
+        
+std::vector< std::size_t >
+genBelemTet( std::size_t nbfac,
+              const std::vector< std::size_t >& inpofa,
+              const std::pair< std::vector< std::size_t >,
+                               std::vector< std::size_t > >& esup )
+// *****************************************************************************
+//  Generate derived data, host elements for boundary faces
```
jbakosi: "Host" elements are? The domain element (tetrahedron) sharing a side/face with a boundary? If so, let's make it more clear for those who do not speak this language that well (yet).

src/UnitTest/tests/Mesh/TestDerivedData.h
```diff
@@ -2418,6 +2418,98 @@ void DerivedData_object::test< 57 >() {
   }
 }
 
+//! Generate and test number of boundary-faces for tet-only mesh
+template<> template<>
+void DerivedData_object::test< 65 >() {
+  set_test_name( "genNbfacTet for tetrahedra" );
```
jbakosi: Can you please give a nice, pronouncable, name for the users? This goes to the screen. E.g., _"boundary faces for tetrahedra"._

src/IO/ExodusIIMeshReader.C
```diff
@@ -377,6 +377,42 @@ ExodusIIMeshReader::readElements( const std::array< std::size_t, 2 >& ext,
   std::move( begin(inpoel), end(inpoel), std::back_inserter(conn) );
 }
 
+void
+ExodusIIMeshReader::readFaces( std::size_t nbfac,
+                               std::vector< std::size_t >& conn )
+// *****************************************************************************
+//  Read element connectivity of a number of mesh cells from ExodusII file
```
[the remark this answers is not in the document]
adityakpandare: Done.

src/Inciter/Discretization.C
```diff
@@ -42,9 +42,9 @@ Discretization::Discretization(
   const std::unordered_map< std::size_t, std::size_t >& filenodes,
   const tk::UnsMesh::EdgeNodes& edgenodes,
   int nchare,
-  std::size_t nbfac,
+  std::size_t tnbfac,
   const std::map< int, std::vector< std::size_t > >& bface,
-  const std::map< int, std::vector< std::size_t > >& belem ) :
+  const std::vector< std::size_t >& t_triinpoel ) :
```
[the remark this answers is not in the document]
adityakpandare: Yeah sure. Only reason I put "t_" in front of it is to clearly discriminate it from m_triinpoel, since the former is for the entire mesh, and the latter is for only that chunk. I've made the change though.

src/Inciter/Discretization.h
```diff
@@ -285,19 +286,22 @@ class Discretization : public CBase_Discretization {
...
     std::vector< int > m_esuel;
     //! Rotal number of faces
     std::size_t m_ntfac;
-    //! Wlement surrounding faces
-    std::vector< int > m_esuf;
     //! Face-node connectivity
     std::vector< std::size_t > m_inpofa;
+    //! Boundary element vector
+    std::vector< std::size_t > m_belem;
+    //! Wlement surrounding faces
```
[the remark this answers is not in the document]
adityakpandare: Oops!

src/Inciter/Transporter.C
```diff
@@ -234,21 +203,50 @@ Transporter::createPartitioner()
...
-  auto nbfac = er.readSidesetFaces( bface, belem );
+  std::map< int, std::vector< std::size_t > > bface;
+  auto nbfac = er.readSidesetFaces( bface );
+
+  std::vector< std::size_t > triinpoel;
+  const auto scheme = g_inputdeck.get< tag::selected, tag::scheme >();
+
+  if (nbfac<1 && scheme == ctr::SchemeType::DG)
+  {
+    CkPrintf( "ERROR: Boundary faces not specified using side-sets in ",
```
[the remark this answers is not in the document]
adityakpandare: Cool! Made the change.

src/Inciter/Transporter.C
```diff
@@ -234,21 +203,50 @@ Transporter::createPartitioner()
...
+    if (sidenodes.find(i) == end(sidenodes)) {
+      m_print.diag( "WARNING: Boundary conditions specified on side set " +
+        std::to_string(i) + " which does not exist in mesh file" );
+      break;
+    }
     if (bface.find(i) == end(bface)) {
       m_print.diag( "WARNING: Boundary conditions specified on side set face " +
         std::to_string(i) + " which does not exist in mesh file" );
       break;
     }
```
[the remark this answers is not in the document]
adityakpandare: Yeah, that's true. Getting rid of one of them.

src/Mesh/DerivedData.C
adityakpandare: I think it's better to leave triinpoel as it is, since all other connectivity arrays don't have a "_sub". However, I'll rename t_triinpoel as triinpoel_complete. Is that okay?

jbakosi: Sounds good.

src/Mesh/DerivedData.C
adityakpandare: Okay. Removed the comment.

adityakpandare:
Thanks for the corrections @jbakosi!

PR #194, opened 2018-01-26
Remove hard-coded number of args from Scheme

jbakosi:
This removes some overloads from `Scheme`. This makes it less general, but only removes overloads that have not been used. As a result, this code is more readable, less error prone, and more generic for the overloads that stay.

In particular, the removed overloads remove our ability to pass the optional `CkEntryOptions` Charm++ argument to the wrapped entry methods, but we don't use them right now. If/when we need to add the `CkEntryOptions` argument, we can just either add back the overload(s) (if we really want both) or simply add back the CkEntryOptions argument (without it being optional). Until then, we are left with more readable code.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/194)


jbakosi:
Hey @rfbird : we just tried reviewable.io on #198 and we like it. Please give it a try with this PR if you want.

PR #205, opened 2018-02-15
Add helper class for iterating over linked lists

jbakosi:
Inspired by @rfbird in https://github.com/quinoacomputing/quinoa/pull/188#pullrequestreview-95860305, add new helper class, `Around`, in `Mesh/Around.h` with unit tests.

This class helps simplifying client code for iterating on entries surrounding entries via linked lists (a pair of vectors) derived from unstructured mesh connectivity. It can be used for any linked lists of type `std::pair< std::vector< std::size_t >, std::vector< std::size_t > >`.

As a result, code, e.g., iterating through all elements surrounding node `p`, previously written as
```
for (auto i=esup.second[p]+1; i<=esup.second[p+1]; ++i) {
  points.push_back( esup.first[i] );
}
```
can now be written as
```
for (auto e : tk::Around(esup,p)) {
  points.push_back(e);
}
```
I think this latter is cleaner, easier to read, and less error-prone. Please suggest a better name for `Around`.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/205)

### DIFF
--- a/src/Main/UnitTest.C
+++ b/src/Main/UnitTest.C
@@ -90,6 +90,7 @@ const int MAX_TESTS_IN_GROUP = 80;
 
 #include "tests/Mesh/TestDerivedData.h"
 #include "tests/Mesh/TestReorder.h"
+#include "tests/Mesh/TestAround.h"
 
 #include "tests/RNG/TestRNG.h"
 #ifdef HAS_MKL
--- a/src/Mesh/Around.h
+++ b/src/Mesh/Around.h
@@ -1,0 +1,68 @@
+// *****************************************************************************
+/*!
+  \file      src/Mesh/Around.h
+  \copyright 2012-2015, J. Bakosi, 2016-2018, Los Alamos National Security, LLC.
+  \brief     Helper class for iterating through linked lists of derived data
+  \details   Helper class for iterating through every item in a linked list data
+    structure derived from unstructured mesh connectivity.
+  \see src/Mesh/DerivedData.h
+*/
+// *****************************************************************************
+#ifndef Around_h
+#define Around_h
+
+#include <vector>
+#include <map>
+#include <utility>
+#include <cstddef>
+#include "Types.h"
+#include "Fields.h"
+#include "UnsMesh.h"
+
+namespace tk {
+
+//! \brief Helper class simplifying client code for iterating on entries
+//!   surrounding entries via linked lists derived from unstructured mesh
+//!   connectivity
+class Around {
+
+  private:
+    //! Linked list type: T1: item list, T2: index list
+    using List =
+      std::pair< std::vector< std::size_t >, std::vector< std::size_t > >;
+
+    //! Const iterator to the item list in List::T1
+    using const_iterator = List::first_type::const_iterator;
+
+    //! Difference type for iterator/pointer arithmetics
+    using diff_type = List::first_type::difference_type;
+
+  public:
+    //! Constructor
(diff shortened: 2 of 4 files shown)


jbakosi:
The state is a reference and a long int and it is entirely in a header file, because I thought it will be a template. It is used on the fly as the unit tests show. I cannot think of any drawback or overhead.

Any better idea for a name instead of `Around`. (The logic is that it is always used to access entries _around_ (or surrounding) some entry, where _entry_ is point, face, element, etc.

adityakpandare:
I'm curious if we could do something for vectors with for example nnpf*ntfac entries? Many derived data structures follow this kind of structure, esuelTet, inpofa, triinpoel, etc. Or will it be an overkill?

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/205#-:-L5Q9apaNWoUYpFfiEoJ:b-w89zm0)*

adityakpandare:
This will make things very convenient! Thanks for adding this!

---

Reviewed 4 of 4 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/205#-:-L5QAgAbFc0l1Y7lO8Ps:bg675tw)*

jbakosi:
Since the number of nodes per faces, `nnpf`, is constant (for now), this is less of a concern, I guess. However,  I think current code such as
```
for (std::size_t e=0; e<nelem; ++e) {
    for (std::size_t ip=nfpe*e; ip<nfpe*(e+1); ++ip) {
       auto jelem = esuelTet[ip];
       use jelem
    }
}
```
could be replaced with
```
for (std::size_t e=0; e<nelem; ++e) {
  for (auto jelem : tk::Around(esuelTet,e)) {
    use jelem
  }
}
```

I think it does look simpler. Is that what you meant?

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/205#-:-L5QSdSu0fAao-kynLxd:buo2awk)*

adityakpandare:
Yes, that's what I meant. But is tk::Around able to handle that? Shouldn't there be an argument of "nfpe" somewhere? I thought I'd approved this. I'll approve it now.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/205#-:-L5QVhSO1jbc6U6npslB:b-yswk51)*

adityakpandare:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/205#-:-L5QWDb54ATX9hSrR95Z:bnfp4nl)*

jbakosi:
No,`tk::Around` cannot handle that in its current form because it takes a pair of vectors. Another similar helper class can be written to handle the case with a vector, which, in this case, is really a 2D array with one fixed (e.g., `nfpe`) dimension. And you are right, the constructor would have to take another argument, the size of the fixed dimension. So it would be something like:
```
for (std::size_t e=0; e<nelem; ++e) {
  for (auto jelem : tk::Around(esuelTet,e,4)) {
    use jelem
  }
}
```

Is that worth the effort?

PR #208, opened 2018-02-21
Re-enable progress indicators in inciter

jbakosi:
Progress indicators (`tk::Progress`) can be used for getting additional diagnostics info useful during large runs, enabled by the `-f` (_feedback_) command linear argument for inciter. This commit also takes out some stale, now unused, progress indicators that have been disabledt during a recent revamp of the discretization classes.

Another important change is to convert node and edge sets before sending them in `Partitioner::query` as the receiving side does not search, so a vector is sufficient. This results in 30% speedup, measured on 288 CPU cores with a 18M mesh with and without initial uniform mesh refinement (yielding a 146M mesh) - both 30% speedup.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/208)

### DIFF
--- a/src/Inciter/Transporter.h
+++ b/src/Inciter/Transporter.h
@@ -426,28 +426,14 @@ class Transporter : public CBase_Transporter {
 
     //! Non-reduction target for receiving progress report on flattening mesh
     void peflattened() { m_progReorder.inc<0>(); }
+    //! Non-reduction target for receiving progress report on node ID query
+    void pequery() { m_progReorder.inc<1>(); }
     //! Non-reduction target for receiving progress report on node ID mask
-    void pemask() { m_progReorder.inc<1>(); }
+    void pemask() { m_progReorder.inc<2>(); }
     //! Non-reduction target for receiving progress report on reordering mesh
-    void pereordered() { m_progReorder.inc<2>(); }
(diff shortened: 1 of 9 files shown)


src/Inciter/Transporter.h
```diff
@@ -426,28 +426,14 @@ class Transporter : public CBase_Transporter {
 
     //! Non-reduction target for receiving progress report on flattening mesh
     void peflattened() { m_progReorder.inc<0>(); }
+    //! Non-reduction target for receiving progress report on node ID query
+    void pequery() { m_progReorder.inc<1>(); }
     //! Non-reduction target for receiving progress report on node ID mask
-    void pemask() { m_progReorder.inc<1>(); }
```
rfbird: Is this another location we can try and replace magic numbers with some semantic content?

jbakosi: Yeah, you are right, I could do an enum or something. Noted, will fix.

jbakosi: And after pushing a new commit, my previous reply to this is, of course, gone! Hey @rfbird, I hate to be a pain, but let's not use github for code review anymore, reviewable.io is a lot nicer, and most importantly, preservers our discussion.

PR #212, opened 2018-02-23
Send only chare-boundary surface nodes/edges during inciter setup

jbakosi:
This commit reduces the volume of communication from domain to surface nodes and thus leads to a drastic improvement of the setup phase of inciter, especially with mesh refinement.

In particular, this commit puts in a preparatory step in `Partitioner::gather()`, which marks the beginning of the setup phase before global distributed mesh reordering can start.

This setup phase consists of a handshake, a two-step procedure, in which all PEs send a query to all other PEs with the node IDs of the mesh chunk they hold after partitioning. In the second step of this
handshake, as an answer to the query, each PE responds to all other PEs with the node IDs and the associated chare IDs they share with the PE the request came from. The two steps of this handshake are given by `Partitioner::query()` and `mask()`, respectively.

Before this commit the node IDs (and edge-end-point IDs in case of initial uniform mesh refinement was selected by the user) were sent during the query phase for the full mesh chunk domain. This commit
replaces the domain node IDs with only those node IDs that are on the chare-boundary surface, which are the only ones that are shared among chares and PEs. Thus this reduces the amount of data transferred during this two-step all-to-all communication step.

Tested with a 18M-cell mesh using 288 PEs this yields about a 10% improvement without mesh refinement (which does not send edge-end-point IDs), but leads to a serious speedup of **3.6x(!)** with mesh refinement. This is because the edge-end-point IDs contain a lot more data compared to node IDs.

Other interesting details:

  * For testing purposes, temporarily switched back to the old `Partitioner::refine()`, which is faster that the newer and more general one that is under construction. Note that the more general refine() is still there but commented.

  * Clarified the documentation on and abstracted away the ordering convention in `lpofa`, which is now reused in multiple places.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/212)


adityakpandare:
Wow! That's an excellent speedup!

---

Reviewed 3 of 3 files at r1.

---

*[src/Inciter/Partitioner.C, line 395 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/212#-L67y-p1HBAoz-KD675M:-L67y-p1HBAoz-KD675N:b-83sjoh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/96712b66fca1ffbcc885fb7c3a9454f566a2aa5d/src/Inciter/Partitioner.C#L395)):*
> ```
>   }
>   tk::unique( nodes );
> ```

Nice!

---

*[src/Mesh/DerivedData.h, line 31 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/212#-L67xQ8lkFGtX3O8C7tz:-L67xQ8lkFGtX3O8C7u-:b-vp6xny) ([raw file](https://github.com/quinoacomputing/quinoa/blob/96712b66fca1ffbcc885fb7c3a9454f566a2aa5d/src/Mesh/DerivedData.h#L31)):*
> ```objc
> //!   that not only these mappings are important, but also the order of the
> //!   nodes within the triplets as this specific order also defines the outwards
> //!   normal of each face.
> ```

This is a very clear explanation. Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/212#-:-L67wbDJrwYA91xuzy6N:b-iruneg)*

adityakpandare:
Reviewed 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/212)*

jbakosi:
@rfbird Sure, please use reviewable.io as that will nicely save our discussion for a PR.

`query()` and `mask()` I mention only to give background. The real stuff is in `gather()`.

PR #213, opened 2018-02-24
Serial implementation of linear Advection using DGP0

adityakpandare:
This commit includes implementation of DGP0 for advection equation in serial. The test is the advection of a 2D Gaussian function. At this time the advection fluxes are hard-coded in a very ‘non-general’ way in DG.C. The next commit will migrate this code to DGPDE, and use the DGAdvection and Problem classes. Another set of changes is that the Global-Exodus node IDs are done away with in the DG.C, and the FaceData class now works with the local renumbered node-IDs.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/213)

### DIFF
--- a/src/IO/ExodusIIMeshReader.C
+++ b/src/IO/ExodusIIMeshReader.C
@@ -379,13 +379,10 @@ ExodusIIMeshReader::readElements( const std::array< std::size_t, 2 >& ext,
 
 void
 ExodusIIMeshReader::readFaces( std::size_t nbfac,
-                               const std::vector< std::size_t >& node_map,
                                std::vector< std::size_t >& conn )
 // *****************************************************************************
 //  Read face connectivity of a number of boundary faces from ExodusII file
 //! \param[in] nbfac Number of boundary faces
-//! \param[in] node_map Vector mapping the local Exodus node-IDs to global
-//!            node-IDs
 //! \param[inout] conn Connectivity vector to push to
 //! \details This function reads-in the total number of boundary faces 
 //!   also called triangle-elements in the EXO2 file, and their connectivity.
@@ -406,7 +403,7 @@ ExodusIIMeshReader::readFaces( std::size_t nbfac,
     count = f*nnpf;
     for (std::size_t i=0; i<nnpf; ++i)
     {
-      conn.push_back( node_map[ l_triinpoel[count+i] ] );
+      conn.push_back( l_triinpoel[count+i] );
     }
   }
 }
--- a/src/IO/ExodusIIMeshReader.h
+++ b/src/IO/ExodusIIMeshReader.h
@@ -90,7 +90,6 @@ class ExodusIIMeshReader {
 
     //! Read face connectivity of a number boundary faces from file
     void readFaces( std::size_t nbfac,
-                    const std::vector< std::size_t >& nodemap,
                     std::vector< std::size_t >& conn );
 
     //! Read local to global node-ID map
--- a/src/Inciter/DG.C
+++ b/src/Inciter/DG.C
@@ -36,9 +36,14 @@ DG::DG( const CProxy_Discretization& disc,
         const FaceData& fd ) :
   m_itf( 0 ),
   m_disc( disc ),
-  m_u( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
+  m_fd( fd ),
+  m_nelem( m_disc[thisIndex].ckLocal()->Inpoel().size()/4 ),
+  m_u( m_nelem,
        g_inputdeck.get< tag::component >().nprop() ),
-  m_vol( 0.0 )
+  m_vol( 0.0 ),
+  m_lhs( m_nelem, 0.0 ),
+  m_rhs( m_nelem,
+         g_inputdeck.get< tag::component >().nprop() )
 // *****************************************************************************
 //  Constructor
 // *****************************************************************************
@@ -85,14 +90,18 @@ DG::setup( tk::real v )
   // Output fields metadata to output file
   d->writeElemMeta();
 
+  // Compute left-hand side of discrete PDEs
+  lhs();
+
   // zero initial solution vector
   // ...
 
-  std::size_t nnpe(4);
-  std::size_t nelem = d->Inpoel().size()/nnpe;
-
   // Set initial conditions for all PDEs
-  for (std::size_t e=0; e<nelem; ++e)
+  m_ax = 1.0;
+  m_ay = 1.0;
+  m_az = 0.0;
+
+  for (std::size_t e=0; e<m_nelem; ++e)
   {
     auto xcc = m_geoElem(e,1,0);
     auto ycc = m_geoElem(e,2,0);
@@ -118,7 +127,7 @@ DG::setup( tk::real v )
 void
 DG::dt()
 // *****************************************************************************
-// Comppute time step size
+// Compute time step size
 // *****************************************************************************
 {
   tk::real mindt = std::numeric_limits< tk::real >::max();
@@ -227,9 +236,115 @@ DG::out()
 }
 
 void
+DG::lhs()
+// *****************************************************************************
+// Compute left-hand side of discrete transport equations
+// *****************************************************************************
+{
+  for (std::size_t e=0; e<m_nelem; ++e)
+  {
+    m_lhs[e] = m_geoElem(e,0,0);
+  }
+}
+
+void
 DG::rhs()
 // *****************************************************************************
-// Compute right-hand side of transport equations
+// Compute right-hand side of discrete transport equations
+// *****************************************************************************
+{
+  auto& esuf = m_fd.Esuf();
+  auto& belem = m_fd.Belem();
+  auto& triinp = m_fd.Inpofa();
+
+  // initialize rhs as zero
+  for (std::size_t e=0; e<m_nelem; ++e)
+  {
+    m_rhs(e,0,0) = 0.0;
+  }
+
+  // compute internal surface flux integrals
+  for (auto f=m_fd.Nbfac(); f<m_fd.Ntfac(); ++f)
+  {
+    std::size_t el = static_cast< std::size_t >(esuf[2*f]);
+    std::size_t er = static_cast< std::size_t >(esuf[2*f+1]);
+
+    auto farea = m_geoFace(f,0,0);
+
+    std::vector< tk::real > fn { m_geoFace(f,1,0),
+                                 m_geoFace(f,2,0),
+                                 m_geoFace(f,3,0) };
+
+    // need to use tk::Fields::extract() here somehow
+    std::vector< tk::real > ul { m_u(el,0,0) };
+    std::vector< tk::real > ur { m_u(er,0,0) };
+
+    //--- upwind fluxes
+    auto flux = upwindFlux(ul, ur, fn);
+    
+    m_rhs(el,0,0) -= farea * flux[0];
+    m_rhs(er,0,0) += farea * flux[0];
+  }
+
+  // compute boundary surface flux integrals
+  for (std::size_t f=0; f<m_fd.Nbfac(); ++f)
+  {
+    std::size_t el = static_cast< std::size_t >(esuf[2*f]);
+
+    auto farea = m_geoFace(f,0,0);
+
+    //CkPrintf("(%d/%d) : %d  || ", f, m_fd.Nbfac(), el);
+    //CkPrintf("triinpo : %d,  %d,  %d \n", triinp[3*f], triinp[3*f+1], triinp[3*f+2] );
+
+    std::vector< tk::real > fn { m_geoFace(f,1,0),
+                                 m_geoFace(f,2,0),
+                                 m_geoFace(f,3,0) };
+
+    // need to use tk::Fields::extract() here somehow
+    std::vector< tk::real > ul { m_u(el,0,0) };
+    std::vector< tk::real > ur { 0.0 };
+
+    //--- upwind fluxes
+    auto flux = upwindFlux(ul, ur, fn);
+    
+    m_rhs(el,0,0) -= farea * flux[0];
+  }
+}
+
+std::vector< tk::real >
+DG::upwindFlux( std::vector< tk::real > ul,
+                std::vector< tk::real > ur,
+                std::vector< tk::real > fn )
+// *****************************************************************************
+// Riemann solver using upwind method
+//! \param[in] ul Left unknown/state vector
+//! \param[in] ur Right unknown/state vector
+//! \param[in] fn Face unit normal vector
+//! \return Riemann solution using upwind method
+// *****************************************************************************
+{
+    std::vector< tk::real > flux(ul.size(),0);
+
+    // wave speed
+    tk::real swave = m_ax*fn[0] + m_ay*fn[1] + m_az*fn[2];
+
+    // upwinding
+    if (swave > 0.0)
+    {
+      flux[0] = swave * ul[0];
+    }
+    else
+    {
+      flux[0] = swave * ur[0];
+    }
+
+    return flux;
+}
+
+void
+DG::tstep()
+// *****************************************************************************
+// Explicit time-stepping using forward Euler to discretize time-derivative
 // *****************************************************************************
 {
 }
@@ -246,8 +361,11 @@ DG::advance( tk::real newdt )
   // Set new time step size
   d->setdt( newdt );
 
-  // Compute rhs for next time step, solve/advance system, ...
+  // Compute rhs for next time step
   rhs();
+
+  // Advance solution/time-stepping
+  tstep();
 
   // Prepare for next time step
   next();
--- a/src/Inciter/DG.C
+++ b/src/Inciter/DG.C
@@ -40,6 +40,8 @@ DG::DG( const CProxy_Discretization& disc,
   m_nelem( m_disc[thisIndex].ckLocal()->Inpoel().size()/4 ),
   m_u( m_nelem,
        g_inputdeck.get< tag::component >().nprop() ),
+  m_un( m_nelem,
+        g_inputdeck.get< tag::component >().nprop() ),
   m_vol( 0.0 ),
   m_lhs( m_nelem, 0.0 ),
   m_rhs( m_nelem,
@@ -97,8 +99,8 @@ DG::setup( tk::real v )
   // ...
 
   // Set initial conditions for all PDEs
-  m_ax = 1.0;
-  m_ay = 1.0;
+  m_ax = 0.1;
+  m_ay = 0.1;
   m_az = 0.0;
 
   for (std::size_t e=0; e<m_nelem; ++e)
@@ -111,10 +113,10 @@ DG::setup( tk::real v )
     tk::real u = 1.0 * exp( -((xcc-0.25)*(xcc-0.25) 
                             + (ycc-0.25)*(ycc-0.25))/(2.0 * 0.005) );
     m_u(e,0,0) = u; // scalar
+    m_un(e,0,0) = u; // scalar
   }
 
   // Set initial conditions for all PDEs
-  std::cout << "DG::init: " << g_dgpde.size() << '\n';
   for (const auto& eq : g_dgpde) eq.initialize( d->Coord(), m_u, d->T() );
 
   // Output initial conditions to file (regardless of whether it was requested)
@@ -254,14 +256,10 @@ DG::rhs()
 // *****************************************************************************
 {
   auto& esuf = m_fd.Esuf();
-  auto& belem = m_fd.Belem();
-  auto& triinp = m_fd.Inpofa();
+  auto& bface = m_fd.Bface();
 
-  // initialize rhs as zero
-  for (std::size_t e=0; e<m_nelem; ++e)
-  {
-    m_rhs(e,0,0) = 0.0;
-  }
+  // set rhs to zero
+  m_rhs.fill(0.0);
 
   // compute internal surface flux integrals
   for (auto f=m_fd.Nbfac(); f<m_fd.Ntfac(); ++f)
@@ -287,27 +285,92 @@ DG::rhs()
   }
 
   // compute boundary surface flux integrals
-  for (std::size_t f=0; f<m_fd.Nbfac(); ++f)
+
+  // symmetry boundary condition
+  auto bc = bface.find(1);
+
+  if (bc != bface.end())
   {
-    std::size_t el = static_cast< std::size_t >(esuf[2*f]);
+    for (const auto& f : bc->second)
+    {
+      std::size_t el = static_cast< std::size_t >(esuf[2*f]);
 
-    auto farea = m_geoFace(f,0,0);
+      Assert( esuf[2*f+1] == -1,
+              "outside boundary element not -1" );
 
-    //CkPrintf("(%d/%d) : %d  || ", f, m_fd.Nbfac(), el);
-    //CkPrintf("triinpo : %d,  %d,  %d \n", triinp[3*f], triinp[3*f+1], triinp[3*f+2] );
+      auto farea = m_geoFace(f,0,0);
 
-    std::vector< tk::real > fn { m_geoFace(f,1,0),
-                                 m_geoFace(f,2,0),
-                                 m_geoFace(f,3,0) };
+      std::vector< tk::real > fn { m_geoFace(f,1,0),
+                                   m_geoFace(f,2,0),
+                                   m_geoFace(f,3,0) };
 
-    // need to use tk::Fields::extract() here somehow
-    std::vector< tk::real > ul { m_u(el,0,0) };
-    std::vector< tk::real > ur { 0.0 };
+      // need to use tk::Fields::extract() here?
+      std::vector< tk::real > ul { m_u(el,0,0) };
+      std::vector< tk::real > ur { ul[0] };
 
-    //--- upwind fluxes
-    auto flux = upwindFlux(ul, ur, fn);
-    
-    m_rhs(el,0,0) -= farea * flux[0];
+      //--- upwind fluxes
+      auto flux = upwindFlux(ul, ur, fn);
+      
+      m_rhs(el,0,0) -= farea * flux[0];
+    }
+  }
+
+  // inlet boundary condition
+  bc = bface.find(2);
+
+  if (bc != bface.end())
+  {
+    for (const auto& f : bc->second)
+    {
+      std::size_t el = static_cast< std::size_t >(esuf[2*f]);
+
+      Assert( esuf[2*f+1] == -1,
+              "outside boundary element not -1" );
+
+      auto farea = m_geoFace(f,0,0);
+
+      std::vector< tk::real > fn { m_geoFace(f,1,0),
+                                   m_geoFace(f,2,0),
+                                   m_geoFace(f,3,0) };
+
+      // need to use tk::Fields::extract() here?
+      std::vector< tk::real > ul { m_u(el,0,0) };
+      std::vector< tk::real > ur { 0.0 };
+
+      //--- upwind fluxes
+      auto flux = upwindFlux(ul, ur, fn);
+      
+      m_rhs(el,0,0) -= farea * flux[0];
+    }
+  }
+
+  // outlet boundary condition
+  bc = bface.find(3);
+
+  if (bc != bface.end())
+  {
+    for (const auto& f : bc->second)
+    {
+      std::size_t el = static_cast< std::size_t >(esuf[2*f]);
+
+      Assert( esuf[2*f+1] == -1,
+              "outside boundary element not -1" );
+
+      auto farea = m_geoFace(f,0,0);
+
+      std::vector< tk::real > fn { m_geoFace(f,1,0),
+                                   m_geoFace(f,2,0),
+                                   m_geoFace(f,3,0) };
+
+      // need to use tk::Fields::extract() here?
+      std::vector< tk::real > ul { m_u(el,0,0) };
+      std::vector< tk::real > ur { ul[0] };
+
+      //--- upwind fluxes
+      auto flux = upwindFlux(ul, ur, fn);
+      
+      m_rhs(el,0,0) -= farea * flux[0];
+    }
   }
 }
 
@@ -333,20 +396,33 @@ DG::upwindFlux( std::vector< tk::real > ul,
     {
       flux[0] = swave * ul[0];
     }
-    else
+    else if (swave < 0.0)
     {
       flux[0] = swave * ur[0];
+    }
+    else
+    {
+      flux[0] = 0.0;
     }
 
     return flux;
 }
 
 void
-DG::tstep()
+DG::tstep( tk::real dt )
 // *****************************************************************************
 // Explicit time-stepping using forward Euler to discretize time-derivative
 // *****************************************************************************
 {
+  for (std::size_t e=0; e<m_nelem; ++e)
+  {
+    m_u(e,0,0) = m_un(e,0,0) + dt/m_lhs[e] * m_rhs(e,0,0);
+  }
+
+  for (std::size_t e=0; e<m_nelem; ++e)
+  {
+    m_un(e,0,0) = m_u(e,0,0);
+  }
 }
 
 void
@@ -365,7 +441,7 @@ DG::advance( tk::real newdt )
   rhs();
 
   // Advance solution/time-stepping
-  tstep();
+  tstep( newdt );
 
   // Prepare for next time step
   next();
@@ -382,14 +458,14 @@ DG::next()
   // Output field data to file
   out();
   // Compute diagnostics, e.g., residuals
-  auto diag = diagnostics();
+  //auto diag = diagnostics();
   // Increase number of iterations and physical time
   d->next();
   // Output one-liner status report
   d->status();
 
   // Evaluate whether to continue with next step
-  if (!diag) eval();
+  /*if (!diag)*/ eval();
 }
 
 void
--- a/src/Inciter/DG.C
+++ b/src/Inciter/DG.C
@@ -39,12 +39,10 @@ DG::DG( const CProxy_Discretization& disc,
   m_fd( fd ),
   m_u( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
        g_inputdeck.get< tag::component >().nprop() ),
-  m_un( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
-        g_inputdeck.get< tag::component >().nprop() ),
+  m_un( m_u.nunk(), m_u.nprop() ),
   m_vol( 0.0 ),
   m_lhs( m_disc[thisIndex].ckLocal()->Inpoel().size()/4, 0.0 ),
-  m_rhs( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
-         g_inputdeck.get< tag::component >().nprop() )
+  m_rhs( m_u.nunk(), m_u.nprop() )
 // *****************************************************************************
 //  Constructor
 // *****************************************************************************
--- a/src/Inciter/DG.C
+++ b/src/Inciter/DG.C
@@ -37,14 +37,13 @@ DG::DG( const CProxy_Discretization& disc,
   m_itf( 0 ),
   m_disc( disc ),
   m_fd( fd ),
-  m_nelem( m_disc[thisIndex].ckLocal()->Inpoel().size()/4 ),
-  m_u( m_nelem,
+  m_u( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
        g_inputdeck.get< tag::component >().nprop() ),
-  m_un( m_nelem,
+  m_un( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
         g_inputdeck.get< tag::component >().nprop() ),
   m_vol( 0.0 ),
-  m_lhs( m_nelem, 0.0 ),
-  m_rhs( m_nelem,
+  m_lhs( m_disc[thisIndex].ckLocal()->Inpoel().size()/4, 0.0 ),
+  m_rhs( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
          g_inputdeck.get< tag::component >().nprop() )
 // *****************************************************************************
 //  Constructor
@@ -103,7 +102,9 @@ DG::setup( tk::real v )
   m_ay = 0.1;
   m_az = 0.0;
 
-  for (std::size_t e=0; e<m_nelem; ++e)
+  std::size_t nelem = m_u.nunk();
+
+  for (std::size_t e=0; e<nelem; ++e)
   {
     auto xcc = m_geoElem(e,1,0);
     auto ycc = m_geoElem(e,2,0);
@@ -243,7 +244,9 @@ DG::lhs()
 // Compute left-hand side of discrete transport equations
 // *****************************************************************************
 {
-  for (std::size_t e=0; e<m_nelem; ++e)
+  std::size_t nelem = m_u.nunk();
+
+  for (std::size_t e=0; e<nelem; ++e)
   {
     m_lhs[e] = m_geoElem(e,0,0);
   }
@@ -269,9 +272,9 @@ DG::rhs()
 
     auto farea = m_geoFace(f,0,0);
 
-    std::vector< tk::real > fn { m_geoFace(f,1,0),
-                                 m_geoFace(f,2,0),
-                                 m_geoFace(f,3,0) };
+    std::array< tk::real, 3 > fn {{ m_geoFace(f,1,0),
+                                    m_geoFace(f,2,0),
+                                    m_geoFace(f,3,0) }};
 
     // need to use tk::Fields::extract() here somehow
     std::vector< tk::real > ul { m_u(el,0,0) };
@@ -300,9 +303,9 @@ DG::rhs()
 
       auto farea = m_geoFace(f,0,0);
 
-      std::vector< tk::real > fn { m_geoFace(f,1,0),
-                                   m_geoFace(f,2,0),
-                                   m_geoFace(f,3,0) };
+      std::array< tk::real, 3 > fn {{ m_geoFace(f,1,0),
+                                      m_geoFace(f,2,0),
+                                      m_geoFace(f,3,0) }};
 
       // need to use tk::Fields::extract() here?
       std::vector< tk::real > ul { m_u(el,0,0) };
@@ -329,9 +332,9 @@ DG::rhs()
 
       auto farea = m_geoFace(f,0,0);
 
-      std::vector< tk::real > fn { m_geoFace(f,1,0),
-                                   m_geoFace(f,2,0),
-                                   m_geoFace(f,3,0) };
+      std::array< tk::real, 3 > fn {{ m_geoFace(f,1,0),
+                                      m_geoFace(f,2,0),
+                                      m_geoFace(f,3,0) }};
 
       // need to use tk::Fields::extract() here?
       std::vector< tk::real > ul { m_u(el,0,0) };
@@ -358,9 +361,9 @@ DG::rhs()
 
       auto farea = m_geoFace(f,0,0);
 
-      std::vector< tk::real > fn { m_geoFace(f,1,0),
-                                   m_geoFace(f,2,0),
-                                   m_geoFace(f,3,0) };
+      std::array< tk::real, 3 > fn {{ m_geoFace(f,1,0),
+                                      m_geoFace(f,2,0),
+                                      m_geoFace(f,3,0) }};
 
       // need to use tk::Fields::extract() here?
       std::vector< tk::real > ul { m_u(el,0,0) };
@@ -377,7 +380,7 @@ DG::rhs()
 std::vector< tk::real >
 DG::upwindFlux( std::vector< tk::real > ul,
                 std::vector< tk::real > ur,
-                std::vector< tk::real > fn )
+                std::array< tk::real, 3 > fn )
 // *****************************************************************************
 // Riemann solver using upwind method
 //! \param[in] ul Left unknown/state vector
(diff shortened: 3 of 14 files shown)


jbakosi:
Very nice work! See some comments for specifics.

---

Reviewed 14 of 14 files at r1.
Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.C, line 274 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HjBXSD8tgFqesDgaK:-L6HjBXSD8tgFqesDgaL:b1bq6mf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L274)):*
> ```
>     std::vector< tk::real > fn { m_geoFace(f,1,0),
>                                  m_geoFace(f,2,0),
>                                  m_geoFace(f,3,0) };
> ```

Since the size is always 3 (thus known at compile-time), let's use `std::array< tk::real, 3 >`.

---

*[src/Inciter/DG.C, line 277 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HkZQR-EtoQPUp4UsL:-L6HkZQR-EtoQPUp4UsM:b-ln7avm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L277)):*
> ```
> 
>     // need to use tk::Fields::extract() here somehow
>     std::vector< tk::real > ul { m_u(el,0,0) };
> ```

This could be
```
auto ul = m_u.extract(el);
```
But this vector holds a single scalar now. I think when we put this in `PDE`, we will make it more general anyway. So let's not spend time on this too much.

---

*[src/Inciter/DG.C, line 374 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HnVGcqn7rlxhg9Tys:-L6HnVGcqn7rlxhg9Tyt:bqkazby) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L374)):*
> ```
>       m_rhs(el,0,0) -= farea * flux[0];
>     }
>   }
> ```


Nice! I see now better what you meant when we discussed the boundaries and the need for a different loop body. This is good for now, I have some ideas how to nicely reuse the common code here. We will make this nicer when we move to the PDE classes.

---

*[src/Inciter/DG.C, line 380 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6Ho18iZQgWUvIMU2va:-L6Ho18iZQgWUvIMU2vb:b3yb2hz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L380)):*
> ```
> DG::upwindFlux( std::vector< tk::real > ul,
>                 std::vector< tk::real > ur,
>                 std::vector< tk::real > fn )
> ```

Nice Riemann solver! ;-) We will also make this nicer that allows different Riemann types of solvers in the future (if needed).

---

*[src/Inciter/DG.C, line 405 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HlCw82OmgJAowjrhT:-L6HlCw82OmgJAowjrhU:bq425fv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L405)):*
> ```
>     else
>     {
>       flux[0] = 0.0;
> ```

This is already covered by `std::vector< tk::real > flux(ul.size(),0);`

---

*[src/Inciter/DG.C, line 412 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HpYcrbsv7SSaBgp4N:-L6HpYcrbsv7SSaBgp4O:bi4te0n) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L412)):*
> ```
> 
> void
> DG::tstep( tk::real dt )
> ```

To make it consistent with other discretizations, let's call this `solve()`.

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6HfPVzfvJs0sTFnlg0:bsvh8bj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*
> ```
>   {
>     m_un(e,0,0) = m_u(e,0,0);
>   }
> ```

This could be done with
```
m_un = m_u;
```

Also, since `m_un == m_u`, if `m_lhs` is a `tk::Data`, we can use
```
m_u += 1.0/dt * m_lhs * m_rhs;
```
which hides both `for` loops. Correct?

Not yet:
```
m_u += m_lhs * m_rhs / dt;
```
that would require a new overload in `Data` for `operator/`, which I think could be done as well, but have not had the need for it so far. If we add it, it needs (unit) tests though, so I decided not bother with it for now.

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6HUMgk65kqtawQxpaf:bwgg41w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*
> ```objc
>     FaceData m_fd;
>     //! Total number of elements in this chunk
>     std::size_t m_nelem;
> ```

`m_nelem` could be eliminated if we just use `m_u.nunk()` instead. I realize that is harder to read, so locally when `m_nelem` is needed, I would do
```
  auto nelem = m_u.nunk();
  ... use nelem ...
```

---

*[src/Inciter/DG.h, line 107 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUvamR0nHpaU071x9:-L6HUvamR0nHpaU071xA:b-l9zq20) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L107)):*
> ```objc
>     tk::real m_az;
>     //! Left-hand side mass-matrix which is a diagonal matrix
>     std::vector< tk::real > m_lhs;
> ```

This is okay for now, but most likely `m_lhs` will have to be of type `tk::Field` when we switch to systems and the generic `PDE` machinery.

---

*[src/Inciter/Transporter.C, line 213 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L67Gm6nh6mG5Rs0FnvM:-L67Gm6nh6mG5Rs0FnvN:b1bcjz2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/Inciter/Transporter.C#L213)):*
> ```
> 
>   // Read local to global node-ID map from file
>   auto nodemap = er.readNodemap();
> ```

Whoa! I did not realize we could completely remove the exodus nodemap. Since we do not read this, it seems like we don't even need `ExodusIIMeshReader::readNodeMap()`. But let's keep that for now in case we do need it later.

See also the comments for `TestExodusIIMeshReader.h`. Since we will keep `ExodusIIMeshReader::readNodeMap()` but now can/will remove their use in the couple of unit tests in `TestExodusIIMeshReader.h`, can you please still add a unit test for it?

---

*[src/Mesh/DerivedData.C, line 828 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HXNEdIEZAaLnXjX0k:-L6HXNEdIEZAaLnXjX0l:bgimuqr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L828)):*
> ```
> //!   side set ids for the entire mesh.
> //! \param[in] lid Map of global renumbered node-IDs to local renumbered
> //!   node-IDs
> ```

I feel like some additional explanation here would make this even more understandable: "Mapping between the node indices used in the smaller inpoel connectivity (a subset of the entire triinpoel _complete connectivity), e.g., after mesh partitioning."

---

*[src/Mesh/DerivedData.C, line 909 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HayMxrxH3giYclXUx:-L6HayMxrxH3giYclXUy:b-nzjl7e) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L909)):*
> ```
>             triinpoel.push_back( nd->second );
>           }
>         }
> ```

Is `triinpoel_complete` always a superset of `triinpoel`? If so, `lid.find()` must always find the global id coming from `triinpoel_complete`. If that's case, we should use instead
```
triinpoel.push_back( tk::cref_find(lid,ip) );
```
instead of
```
auto nd = lid.find( ip );
if (nd != end(lid))
{
  triinpoel.push_back( nd->second );
}
```
in which `cref_find` will correctly always throw an exception (in DEBUG) if the key is not found, indicating an error (if I'm correct that `triinpoel_complete` is always a superset of `triinpoel`). (Using `find` silently swallows it.)

---

*[src/Mesh/DerivedData.C, line 1227 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HZDZB_xscGShNUA98:-L6HZDZB_xscGShNUA99:b9ohwyv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L1227)):*
> ```
>     inpofa[icoun+0] = triinpoel[icoun+2];
>     inpofa[icoun+1] = triinpoel[icoun+1];
>     inpofa[icoun+2] = triinpoel[icoun+0];
> ```

Was this wrong before?

---

*[src/UnitTest/tests/IO/TestExodusIIMeshReader.h, line 1553 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6H_cGo0pLzhaJ9pl1i:-L6H_cGpYxtMxpOjL68E:byv6d38) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/UnitTest/tests/IO/TestExodusIIMeshReader.h#L1553)):*
> ```objc
>   // Read boundary face-node connectivity
>   std::vector< std::size_t > triinpoel;
>   auto nodemap = er.readNodemap();
> ```

Since `nodemap` is gone, this is not necessary.

---

*[src/UnitTest/tests/IO/TestExodusIIMeshReader.h, line 1602 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6H_kM7scWG7oMtcoGM:-L6H_kM7scWG7oMtcoGN:bijdsfs) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/UnitTest/tests/IO/TestExodusIIMeshReader.h#L1602)):*
> ```objc
>     std::string infile( REGRESSION_DIR "/meshconv/gmsh_output/box_24_ss1.exo" );
>     tk::ExodusIIMeshReader er( infile );
>     auto nodemap = er.readNodemap();
> ```

Since nodemap is gone, this is not necessary.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-:-L6HpvDGBL4G76URzeCj:b-8m3w81)*

adityakpandare:
Thanks!

---

Review status: all files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.C, line 274 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HjBXSD8tgFqesDgaK:-L6IC6lWEI31NHUnHDI5:b-l2v4ii) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L274)):*

Since the size is always 3 (thus known at compile-time), let's use `std::array< tk::real, 3 >`.
</blockquote></details>

That's right. Done.

---

*[src/Inciter/DG.C, line 277 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HkZQR-EtoQPUp4UsL:-L6IC1HJLrZiS7Y6gZYw:bedzbu6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L277)):*

This could be
```
auto ul = m_u.extract(el);
```
But this vector holds a single scalar now. I think when we put this in `PDE`, we will make it more general anyway. So let's not spend time on this too much.
</blockquote></details>

Okay we'll do this in the next PR.

---

*[src/Inciter/DG.C, line 374 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HnVGcqn7rlxhg9Tys:-L6ICFyC5x67_OPMiw3m:b45c54l) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L374)):*

Nice! I see now better what you meant when we discussed the boundaries and the need for a different loop body. This is good for now, I have some ideas how to nicely reuse the common code here. We will make this nicer when we move to the PDE classes.
</blockquote></details>

Yeah I was going to bring it up with you, so that we can still reuse most of the code.

---

*[src/Inciter/DG.C, line 380 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6Ho18iZQgWUvIMU2va:-L6ICRxzUbUm6LEcEDzX:b1ttwm0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L380)):*

Nice Riemann solver! ;-) We will also make this nicer that allows different Riemann types of solvers in the future (if needed).
</blockquote></details>

Yes. Once we get to Euler equations, we'll have more choice. Even this can be written better actually. Let me clean it up.

---

*[src/Inciter/DG.C, line 405 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HlCw82OmgJAowjrhT:-L6ICdabZlfFzWt5pHLR:bss4a2b) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L405)):*

This is already covered by `std::vector< tk::real > flux(ul.size(),0);`
</blockquote></details>

Removed.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-:-L6IC9ulzWtaDWmN4zt8:b-ucfknh)*

adityakpandare:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.C, line 412 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HpYcrbsv7SSaBgp4N:-L6IY1XHlhTnlgH-97VY:bb6un1j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L412)):*

To make it consistent with other discretizations, let's call this `solve()`.
</blockquote></details>

Done

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6IYKp5LsYxgfTor_rc:btkv9f) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

This could be done with
```
m_un = m_u;
```

Also, since `m_un == m_u`, if `m_lhs` is a `tk::Data`, we can use
```
m_u += 1.0/dt * m_lhs * m_rhs;
```
which hides both `for` loops. Correct?

Not yet:
```
m_u += m_lhs * m_rhs / dt;
```
that would require a new overload in `Data` for `operator/`, which I think could be done as well, but have not had the need for it so far. If we add it, it needs (unit) tests though, so I decided not bother with it for now.
</blockquote></details>

Made the first change. However, we're going to change this to RK3 soon. So I think we should keep the other for-loop. We'll need it. We'll have another for-loop going around it for the RK-stages too. Is that ok?

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6I_qcci3lcBvCOK0q2:bq5elg9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*

`m_nelem` could be eliminated if we just use `m_u.nunk()` instead. I realize that is harder to read, so locally when `m_nelem` is needed, I would do
```
  auto nelem = m_u.nunk();
  ... use nelem ...
```
</blockquote></details>

Actually, we are going to need nelem multiple times in multiple functions. Do you still think we should get rid of it?

---

*[src/Inciter/DG.h, line 107 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUvamR0nHpaU071x9:-L6Ia0yz2K1Ylid8KVJe:b-837i3v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L107)):*

This is okay for now, but most likely `m_lhs` will have to be of type `tk::Field` when we switch to systems and the generic `PDE` machinery.
</blockquote></details>

Okay.

---

*[src/Inciter/Transporter.C, line 213 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L67Gm6nh6mG5Rs0FnvM:-L6Ia5ks87pWJwArmYdh:b-sbzuip) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/Inciter/Transporter.C#L213)):*

Whoa! I did not realize we could completely remove the exodus nodemap. Since we do not read this, it seems like we don't even need `ExodusIIMeshReader::readNodeMap()`. But let's keep that for now in case we do need it later.

See also the comments for `TestExodusIIMeshReader.h`. Since we will keep `ExodusIIMeshReader::readNodeMap()` but now can/will remove their use in the couple of unit tests in `TestExodusIIMeshReader.h`, can you please still add a unit test for it?
</blockquote></details>

Yeah! way cleaner now right?! Yes, I think we can keep that function, although it's unused.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[src/UnitTest/tests/IO/TestExodusIIMeshReader.h, line 1553 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6H_cGo0pLzhaJ9pl1i:-L6IajxeNA7Wp3bJF87T:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/UnitTest/tests/IO/TestExodusIIMeshReader.h#L1553)):*

Since `nodemap` is gone, this is not necessary.
</blockquote></details>

Done.

---

*[src/UnitTest/tests/IO/TestExodusIIMeshReader.h, line 1602 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6H_kM7scWG7oMtcoGM:-L6Iaj-FFts7FhX0FxN-:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/UnitTest/tests/IO/TestExodusIIMeshReader.h#L1602)):*

Since nodemap is gone, this is not necessary.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[src/Inciter/Transporter.C, line 213 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L67Gm6nh6mG5Rs0FnvM:-L6Ib-YWkWQ2AUcpBlo9:b-c5iwt7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/Inciter/Transporter.C#L213)):*

Yeah! way cleaner now right?! Yes, I think we can keep that function, although it's unused.
</blockquote></details>

Yes I'll add a unit test for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[src/Mesh/DerivedData.C, line 828 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HXNEdIEZAaLnXjX0k:-L6IcO5JKKssdVuYQIFT:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L828)):*

I feel like some additional explanation here would make this even more understandable: "Mapping between the node indices used in the smaller inpoel connectivity (a subset of the entire triinpoel _complete connectivity), e.g., after mesh partitioning."
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6Ib7hkqwgfrjqNIS2o:b-4qqyd6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Made the first change. However, we're going to change this to RK3 soon. So I think we should keep the other for-loop. We'll need it. We'll have another for-loop going around it for the RK-stages too. Is that ok?
</blockquote></details>

We can still do
```
m_u = m_un + m_lhs * m_rhs / dt;
```
and eliminate the element loop, right?

I think the RK-loop should be at a higher level so we can overlap communication and computation (of the various tasks, e.g., rhs, within a time step stage) across time step stages. In other words, a single `rhs()` and `solve()` pair will be only for a single RK-stage and the RK "for loop" will not really be a loop within `solve()`, but across calling `advance()` ->  `next()` -> `advance()`, etc.

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6IdLaK7pIs_VFxd9tv:b-19gwpr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*

Actually, we are going to need nelem multiple times in multiple functions. Do you still think we should get rid of it?
</blockquote></details>

Yes, because `m_u.nunk()` will be available everywhere in `DG`, right?

---

*[src/Inciter/Transporter.C, line 213 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L67Gm6nh6mG5Rs0FnvM:-L6IddEmsWotNg0WZKRR:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a97811e90d036ae2f24534b03ff32131fcca3e80/src/Inciter/Transporter.C#L213)):*

Yes I'll add a unit test for this.
</blockquote></details>

Thanks!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/Mesh/DerivedData.C, line 909 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HayMxrxH3giYclXUx:-L6IdtOqcwV72VYfdtyo:b2pbbs2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L909)):*

Is `triinpoel_complete` always a superset of `triinpoel`? If so, `lid.find()` must always find the global id coming from `triinpoel_complete`. If that's case, we should use instead
```
triinpoel.push_back( tk::cref_find(lid,ip) );
```
instead of
```
auto nd = lid.find( ip );
if (nd != end(lid))
{
  triinpoel.push_back( nd->second );
}
```
in which `cref_find` will correctly always throw an exception (in DEBUG) if the key is not found, indicating an error (if I'm correct that `triinpoel_complete` is always a superset of `triinpoel`). (Using `find` silently swallows it.)
</blockquote></details>

Yes, triinpoel_complete is always a superset of triinpoel. I'll make this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6IeYFsNBAwE6QPmtMV:bkxox2k) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*

Yes, because `m_u.nunk()` will be available everywhere in `DG`, right?
</blockquote></details>

Yes it will be. Okay I'll make this change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/Mesh/DerivedData.C, line 909 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HayMxrxH3giYclXUx:-L6Ieq3Gkz6cu9BI5zY1:b-ut7qva) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L909)):*

Yes, triinpoel_complete is always a superset of triinpoel. I'll make this change.
</blockquote></details>

Thanks, that will lead to better error messages later, especially important in async/parallel.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6Ieecb-7HFEyqvfoW9:b-tw7gn2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*

Yes it will be. Okay I'll make this change.
</blockquote></details>

However, now that I think about this more, would m_u.nunk() always be equal to nelem? For DG for example, we will have multiple DOF's of the unknown per element. For P1, we will have 4 unknowns per equation per element. In this case is nelem = m_u.nunk() always true?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.h, line 91 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HUMgk65kqtawQxpae:-L6IfE2o4esQC6WKN3GI:bjrqh00) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.h#L91)):*

However, now that I think about this more, would m_u.nunk() always be equal to nelem? For DG for example, we will have multiple DOF's of the unknown per element. For P1, we will have 4 unknowns per equation per element. In this case is nelem = m_u.nunk() always true?
</blockquote></details>

Sure, not only for P1, but also for systems `m_u.nunk()` will not equal to `nelem`. But that `nelem` will always be computable (at compile time) from the dimensions of `m_u`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6Il0lg1Y7TmdkX2CnU:b-jpvzsx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

We can still do
```
m_u = m_un + m_lhs * m_rhs / dt;
```
and eliminate the element loop, right?

I think the RK-loop should be at a higher level so we can overlap communication and computation (of the various tasks, e.g., rhs, within a time step stage) across time step stages. In other words, a single `rhs()` and `solve()` pair will be only for a single RK-stage and the RK "for loop" will not really be a loop within `solve()`, but across calling `advance()` ->  `next()` -> `advance()`, etc.
</blockquote></details>

Above, I meant, we can still do
```
m_u += 1.0/dt * m_lhs * m_rhs;
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Mesh/DerivedData.C, line 1227 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HZDZB_xscGShNUA98:-L6IkFBwKIiWMc1C-Ezm:b-le87kf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L1227)):*

Was this wrong before?
</blockquote></details>

Yes, this was flipped before. So the normal was inward pointing. Thus the change even in TestDerivedData.h. This had escaped me earlier when I generated the data-structures for the faces. Realized this when the solution was diverging in a particular way.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Mesh/DerivedData.C, line 1227 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HZDZB_xscGShNUA98:-L6ImLxDKIN5EwL_3LCP:b-a6tnfo) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Mesh/DerivedData.C#L1227)):*

Yes, this was flipped before. So the normal was inward pointing. Thus the change even in TestDerivedData.h. This had escaped me earlier when I generated the data-structures for the faces. Realized this when the solution was diverging in a particular way.
</blockquote></details>

Cool! Nice catch!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6InDuqd28REn5jPmNI:bk0zdr4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Above, I meant, we can still do
```
m_u += 1.0/dt * m_lhs * m_rhs;
```
</blockquote></details>

Actually, we'll need m_u = m_un + 1.0/dt * m_lhs * m_rhs. This is again because of the upcoming RK3.
The form of RK3 is something like:

rk-stage-i
{
   m_u = k1*m_un + k2*m_u + 1.0/dt * m_lhs * k3 * m_rhs
}

I might have to discuss more about the parallelism that you're talking about here, so that I can understand you correctly.

But, yes we can remove the element-loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6IoZNBjIXxd9e5H-uV:b3gkx40) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Actually, we'll need m_u = m_un + 1.0/dt * m_lhs * m_rhs. This is again because of the upcoming RK3.
The form of RK3 is something like:

rk-stage-i
{
   m_u = k1*m_un + k2*m_u + 1.0/dt * m_lhs * k3 * m_rhs
}

I might have to discuss more about the parallelism that you're talking about here, so that I can understand you correctly.

But, yes we can remove the element-loop.
</blockquote></details>

Oh wait. We need 1/m_lhs[e]. How do we do that without the element loop?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6IpGHwnUywtvTtPpaC:bkhsz16) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Oh wait. We need 1/m_lhs[e]. How do we do that without the element loop?
</blockquote></details>

Yes, sorry, my screw-up, I meant:
```
m_u = m_un + 1.0/dt * m_lhs * m_rhs;
```

But why do we need to _dividde by_ the lhs? And yes, those RK-constants will be okay, because we will pass them as arguments to `rhs()` and `solve()`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6IpicIC4pgxWM9rKn7:b-c42rf6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Yes, sorry, my screw-up, I meant:
```
m_u = m_un + 1.0/dt * m_lhs * m_rhs;
```

But why do we need to _dividde by_ the lhs? And yes, those RK-constants will be okay, because we will pass them as arguments to `rhs()` and `solve()`.
</blockquote></details>

Because the lhs is the volume/mass matrix:

m_lhs * (m_u - m_un)/dt  =  m_rhs

m_u = m_un + dt/m_lhs * m_rhs

The difference for DG is that this m_lhs is a block diagonal matrix, of the size ndof * ndof where ndof are the number of DOFs per element, which would depend on the discretization scheme. For P0 it is 1x1.

Now, We could invert it beforehand too and then multiply. But whatever we do, the inversion of the ndof * ndof sized matrix is necessary.

Does this make sense?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6IphZOTOjWQmH71dWx:b-cnhqb5) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Because the lhs is the volume/mass matrix:

m_lhs * (m_u - m_un)/dt  =  m_rhs

m_u = m_un + dt/m_lhs * m_rhs

The difference for DG is that this m_lhs is a block diagonal matrix, of the size ndof * ndof where ndof are the number of DOFs per element, which would depend on the discretization scheme. For P0 it is 1x1.

Now, We could invert it beforehand too and then multiply. But whatever we do, the inversion of the ndof * ndof sized matrix is necessary.

Does this make sense?

</blockquote></details>

Yes, totally. We divide by lhs in CG as well, either by inverting the matrix or by the lumped diagonal.

Also, I see now. Once again, I confused things by butchering your originally correct expression. Sorry. ;-) How about:
```
m_u = m_un + dt*m_rhs/m_lhs;
```
This should compile.

The reason I screwed up and confused myself is because `operator*` between a `tk::Data` and a scalar _is_ defined for the scalar being on either side. However, `operator/` _is not_ defined between a `tk::Data` and a scalar at all. (This can be done, for the scalar being on either side, but needs unit tests so I don't want to bother with it right now.) However, `operator/` _is_ defined for when both operands are `tk::Data` where lhs and rhs are of the same type and it means element-by-element division. I should really write an `operator/` with the scalar on the right and left.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J1ZMo8duukUrbs6Z8:b-chk555) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Yes, totally. We divide by lhs in CG as well, either by inverting the matrix or by the lumped diagonal.

Also, I see now. Once again, I confused things by butchering your originally correct expression. Sorry. ;-) How about:
```
m_u = m_un + dt*m_rhs/m_lhs;
```
This should compile.

The reason I screwed up and confused myself is because `operator*` between a `tk::Data` and a scalar _is_ defined for the scalar being on either side. However, `operator/` _is not_ defined between a `tk::Data` and a scalar at all. (This can be done, for the scalar being on either side, but needs unit tests so I don't want to bother with it right now.) However, `operator/` _is_ defined for when both operands are `tk::Data` where lhs and rhs are of the same type and it means element-by-element division. I should really write an `operator/` with the scalar on the right and left.
</blockquote></details>

That's okay Jozsef. We can do this later. However, I cannot compile the above change.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J2xzYpwgAI2zTuJnl:bz0yzw1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

That's okay Jozsef. We can do this later. However, I cannot compile the above change.
</blockquote></details>

I think the reason is the type of m_rhs and m_lhs are different. m_lhs (dim = nelem) does not depend on the number of equations in the system we're solving, whereas m_rhs (dim = neqn * nelem) does. Am I right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J4NZ_4YPEhOYDw9Ss:ba095hn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

I think the reason is the type of m_rhs and m_lhs are different. m_lhs (dim = nelem) does not depend on the number of equations in the system we're solving, whereas m_rhs (dim = neqn * nelem) does. Am I right?
</blockquote></details>

Yeah, but right now we have a single scala eq, right? So if both rhs and lhs are tk::Data this should work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J4eF8m104AelocRZZ:b5o0ren) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Yeah, but right now we have a single scala eq, right? So if both rhs and lhs are tk::Data this should work.
</blockquote></details>

Okay, this does not really matter for now. This will be more important for when we move this stuff in the PDE classes.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J4cdxj9f0NWkg7Uuk:brd94et) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Okay, this does not really matter for now. This will be more important for when we move this stuff in the PDE classes.
</blockquote></details>

But m_lhs isn't tk::Data. Its a std::vector< tk::real >. I thought we don't want to change that until we move all this to PDE. Also, is it okay to do m_rhs/m_lhs, considering their dimensions are going to be different in general?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J7VswJ48Fde4mcWGd:bfsgw14) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

But m_lhs isn't tk::Data. Its a std::vector< tk::real >. I thought we don't want to change that until we move all this to PDE. Also, is it okay to do m_rhs/m_lhs, considering their dimensions are going to be different in general?
</blockquote></details>

No, that would not compile. Either way is fine though. Either convert lhs to `tk::Data` or leave it and leave the loop.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

adityakpandare:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J7yTohhPf29Xli28B:b-q0c61y) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

No, that would not compile. Either way is fine though. Either convert lhs to `tk::Data` or leave it and leave the loop.
</blockquote></details>

Okay. Let's address it in the upcoming PR when we try to move everything to PDEs.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:

---

*[src/Inciter/DG.C, line 425 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6HfPVzfvJs0sTFnlg-:-L6J8GRX9OQdq0yAfqjG:brxp9sq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6fb7164a7fe9fab4f6e9d13305c53dcf076fc74b/src/Inciter/DG.C#L425)):*

Okay. Let's address it in the upcoming PR when we try to move everything to PDEs.
</blockquote></details>

Sounds good.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
:lgtm:

---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/Inciter/DG.C, line 42 at r2](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6JLQR3u6RopPZHWKn9:-L6JLQR3u6RopPZHWKnA:bgsv78j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5eec3c010cc4071c8bab3a289bdd4c815ca5f701/src/Inciter/DG.C#L42)):*
> ```
>   m_u( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
>        g_inputdeck.get< tag::component >().nprop() ),
>   m_un( m_disc[thisIndex].ckLocal()->Inpoel().size()/4,
> ```

To reduce code, this could also be
```
m_un( m_u.nunk(), m_u.nprop() ),
```
and the same for m_`rhs`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-:-L6JM9bw6Ipuu7Sqs16Q:bnfp4nl)*

adityakpandare:
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

*[src/Inciter/DG.C, line 42 at r2](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-L6JLQR3u6RopPZHWKn9:-L6JO6AyIkXZqRYbm1az:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5eec3c010cc4071c8bab3a289bdd4c815ca5f701/src/Inciter/DG.C#L42)):*

To reduce code, this could also be
```
m_un( m_u.nunk(), m_u.nprop() ),
```
and the same for m_`rhs`.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

jbakosi:
:lgtm:

---

Review status: 13 of 14 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213#-:-L6JxT7q1IsfvALRb5s-:bnfp4nl)*

jbakosi:
Reviewed 1 of 1 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/213)*

PR #214, opened 2018-02-26
Read mesh node coordinates one by one in inciter

jbakosi:
_Note_: this is to be merged to branch `boundary` instead of `develop`, to show only the diff relevant to this single step towards several ones improving on the setup for mesh reordering in `Partitioner`.

Somewhat counterintuitively, reading mesh node coordinates one by one does not make reading the coordinates slower but it allows inciter to use a lot less memory when reading the mesh node coordinates in parallel.

_Background_: Since in a parallel run reading the mesh is distributed by elements, i.e., PE0 reads approximately elements 0...n, PE1 reads approximately elements n+1...2n, etc., node coordinates of all node IDs associated to elements that are read must be read in.  However, there is no reason to assume that the node IDs of the elements read are nicely contiguously numbered (and experience shows that they are usually not). Thus so far inciter has computed the extents, i.e., the minimum and maximum, of the global node IDs and read in all nodes between them, in the hope that reading in multiple node coordinates will always be faster than reading them one by one. This however, is not very good since there is still a lot of nodes that get read in such a way, because the nodes are so arbitrarily badly ordered in the file compared to the consecutively read elements. As a result, most PEs read most of the coordinates, which does not scale. Thus this commit replaces this extent-based node reading strategy with a more memory-efficient one, which reads only those nodes that are absolutely necessary on a given PE, reading node coordinates one by one but with global node IDs in increasing order, which other than that though, are arbitrary.

The symptom of this problem came when a 144M-cell mesh was fine on 25 compute nodes (each with 128GB RAM), but the nodes of a 288M-cell mesh could not be read on 25 nodes and not even on 70. As a result of this commit, a 288M-cell mesh can now be read fine on 25 nodes (with or without initial uniform mesh refinement, which of course, uses additional memory).

### DIFF
--- a/src/IO/ExodusIIMeshReader.C
+++ b/src/IO/ExodusIIMeshReader.C
@@ -140,24 +140,20 @@ ExodusIIMeshReader::readAllNodes( UnsMesh& mesh ) const
 }
 
 std::array< std::vector< tk::real >, 3 >
-ExodusIIMeshReader::readNodes( const std::array< std::size_t, 2 >& ext ) const
+ExodusIIMeshReader::readNodes( const std::vector< std::size_t >& gid ) const
 // *****************************************************************************
 //  Read coordinates of a number of mesh nodes from ExodusII file
-//! \param[in] ext Extents of element ids whose connectivity to read, both
-//!   inclusive
+//! \param[in] gid Node IDs whose coordinates to read
 //! \return Mesh node coordinates
 // *****************************************************************************
 {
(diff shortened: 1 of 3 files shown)


adityakpandare:
Nice!

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/214#-:-L6Irrtpo-dCcV1aD5OV:b-83sjoh)*

PR #216, opened 2018-03-02
Compute chare-communication map across faces among chares

jbakosi:
Note that this appears to work, but correctness is not yet established.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/216)


adityakpandare:
*[src/Inciter/DG.C, line 69 at r2](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6cQpkFpFxZeCh-0bTc:-L6cQpkFpFxZeCh-0bTd:br6excv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0f4df6bd22b75e7e0e3b44581856c3a99e57dba2/src/Inciter/DG.C#L69)):*
> ```
>       int counter = 0;
>       for (std::size_t en=0; en<4; ++en) {
>         auto i = n.second.find( d->Inpoel()[ e*4+en ] );
> ```

I have a question about Msum. If I'm correct, Msum uses global node-IDs right? If this is the case, the following line:
```c++
        auto i = n.second.find( d->Inpoel()[ e*4+en ] );
```
we should replace by:
```c++
        auto i = n.second.find(  d->Gid()[ d->Inpoel()[ e*4+en ] ] );
```

since Inpoel uses local IDs, and we need to map it back to the global IDs so that we can find them correctly in `msum_set`. Does that make sense? I have tried this, and atleast this gives the same number of `m_sum_el` at the print statement in `DG::adj()`. Will check whether the element-IDs are correct now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
I'm testing the m_msum_el currently. There's a possible bug that I've found. I got tipped off by the fact that for 2 partitions, the m_msum_el should contain equal number of element-IDs for both the chares; at this wasn't happening when I run the code. Could you take a look at the comment below, where I've explained it in further detail? Please let me know if it makes sense.

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-:-L6cWmA6Ow9CurSUSjx_:b-m3f766)*

jbakosi:
Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/Inciter/DG.C, line 69 at r2](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6cQpkFpFxZeCh-0bTc:-L6dKJAxu0hWoKYUrCFJ:bsnhjuo) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0f4df6bd22b75e7e0e3b44581856c3a99e57dba2/src/Inciter/DG.C#L69)):*

I have a question about Msum. If I'm correct, Msum uses global node-IDs right? If this is the case, the following line:
```c++
        auto i = n.second.find( d->Inpoel()[ e*4+en ] );
```
we should replace by:
```c++
        auto i = n.second.find(  d->Gid()[ d->Inpoel()[ e*4+en ] ] );
```

since Inpoel uses local IDs, and we need to map it back to the global IDs so that we can find them correctly in `msum_set`. Does that make sense? I have tried this, and atleast this gives the same number of `m_sum_el` at the print statement in `DG::adj()`. Will check whether the element-IDs are correct now.

</blockquote></details>

That makes perfect sense. Thanks for catching this, ye I have missed this. The face that the number of tet ids are equal on the two chares, is promising.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

jbakosi:
So using the gids, I don't get the same number of tet ids on 2PEs. Have you tested this further?

---

Review status: 2 of 4 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-:-L6dNH3MZjBJb6Rj_MB_:blkc4a2)*

adityakpandare:
Reviewed 1 of 4 files at r1, 3 of 9 files at r3.
Review status: 4 of 10 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 1 of 9 files at r3.
Review status: 5 of 10 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 1 of 9 files at r3.
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/UnitTest/tests/Base/TestData.h, line 1715 at r3](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6r2isPLc1HFrr_X3Oy:-L6r2isPLc1HFrr_X3Oz:b-83sjoh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/267fe58acda4eda25f8bbd6027069c201f321bab/src/UnitTest/tests/Base/TestData.h#L1715)):*
> ```objc
>   // tk::Data::enlarge() unimplemented with EqCompUnk data layout
> }
> ```

Nice!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/UnitTest/tests/Base/TestData.h, line 1715 at r3](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6r2isPLc1HFrr_X3Oy:-L6r5pXflW2UGPiDX0e5:b-kcu8r) ([raw file](https://github.com/quinoacomputing/quinoa/blob/267fe58acda4eda25f8bbd6027069c201f321bab/src/UnitTest/tests/Base/TestData.h#L1715)):*

Nice!
</blockquote></details>

`Data::enlarge()` may be needed, but if so, there must be unit-tests right way at such a low level.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Review status: 6 of 10 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/UnitTest/tests/Base/TestData.h, line 1715 at r3](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6r2isPLc1HFrr_X3Oy:-L6r65k5cK5X8cjQrByG:bejbz9x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/267fe58acda4eda25f8bbd6027069c201f321bab/src/UnitTest/tests/Base/TestData.h#L1715)):*

`Data::enlarge()` may be needed, but if so, there must be unit-tests right way at such a low level.
</blockquote></details>

"may NOT be needed"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Review status: 6 of 10 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/UnitTest/tests/Base/TestData.h, line 1715 at r3](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L6r2isPLc1HFrr_X3Oy:-L6rB74eyL2uikfwaeoK:bld2edh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/267fe58acda4eda25f8bbd6027069c201f321bab/src/UnitTest/tests/Base/TestData.h#L1715)):*

"may NOT be needed"
</blockquote></details>

That's true. I'm still getting used to unit-testing. This helps me understand unit-testing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 3 of 9 files at r3, 13 of 14 files at r4.
Review status: 22 of 23 files reviewed at latest revision, 1 unresolved discussion, some commit checks broke.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 1 of 5 files at r6, 4 of 6 files at r9.
Review status: 10 of 12 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/Inciter/DG.C, line 247 at r9](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L7Ujmqo_GsWm_bxh0Ec:-L7UjmqpkuIeUcgGjydG:b-mo4phz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5b7a3bfeabdf564d0e8d93a41079464f8e7fd7e8/src/Inciter/DG.C#L247)):*
> ```
> 
> std::unordered_map< int, std::unordered_set< std::size_t > >
> DG::msumset( const std::vector< std::size_t >& inpofa ) const
> ```

We won't be needing the inpofa here now, correct? Since msum_set doesn't hold the physical boundary faces like we thought before...

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 2 of 6 files at r9.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
That's great! Can't wait to test P0 in parallel with this!

---

Reviewed 1 of 3 files at r11, 1 of 5 files at r12.
Review status: 9 of 13 files reviewed at latest revision, all discussions resolved.

---

*[src/Inciter/DG.h, line 18 at r12](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L7e2h6r2NY-M8y74vbB:-L7e2h6r2NY-M8y74vbC:b-gwfqkm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/055100d68bef5f625205378eb6221a5e976f5b95/src/Inciter/DG.h#L18)):*
> ```objc
>     utilizes the structured dagger (SDAG) Charm++ functionality. The high-level
>     overview of the algorithm structure and how it interfaces with Charm++ is
>     discussed in the Charm++ interface file src/Inciter/diagcg.ci.
> ```

Did you mean dg.ci instead on diagcg.ci?

---

*[src/Mesh/UnsMesh.h, line 75 at r12](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L7e1nCl-fchTcafdeJI:-L7e1nCl-fchTcafdeJJ:bp4t5v0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/055100d68bef5f625205378eb6221a5e976f5b95/src/Mesh/UnsMesh.h#L75)):*
> ```objc
>     //! Unique set of node ID triplets representing unique tetrahedron faces
>     using FaceSet = std::unordered_set< Face, FaceHasher, FaceEq >;
> ```

Nice move shifting this to UnsMesh!

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-:-L6dOKiz2Vmj2ocTp4V6:b-1v9vmu)*

adityakpandare:
Reviewed 2 of 5 files at r12.
Review status: 11 of 13 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Review status: 11 of 13 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/Inciter/DG.h, line 18 at r12](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L7e2h6r2NY-M8y74vbB:-L7e6egDMGpqVi-a5NxG:b-mj12o8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/055100d68bef5f625205378eb6221a5e976f5b95/src/Inciter/DG.h#L18)):*

Did you mean dg.ci instead on diagcg.ci?
</blockquote></details>

You bet! Will fix it, thx.

---

*[src/Mesh/UnsMesh.h, line 75 at r12](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-L7e1nCl-fchTcafdeJI:-L7e6lT6wqoCM007oko3:b-68t90z) ([raw file](https://github.com/quinoacomputing/quinoa/blob/055100d68bef5f625205378eb6221a5e976f5b95/src/Mesh/UnsMesh.h#L75)):*

Nice move shifting this to UnsMesh!
</blockquote></details>

Yeah, that has always been the intention, but less typing during development. ;-)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 2 of 2 files at r13.
Review status: 13 of 14 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 1 of 5 files at r12.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 16 of 18 files at r19.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
I think we're good to merge.

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/216#-:-L8HexNWPIZV6PA7PP5V:b-sj13my)*

jbakosi:
Famous last words:
> I think we're good to merge.

;-)

adityakpandare:
;-)

Well, we're definitely closer now! :sweat_smile:

---







*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-:-L8mUs4s4vUVfHjgqep0:b6erbot)*

adityakpandare:
Reviewed 5 of 7 files at r28.
Review status: 15 of 23 files reviewed at latest revision, all discussions resolved, some commit checks broke.

---

*[src/Inciter/FaceData.C, line 78 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oZyqE8UhSkXS-1PHh:-L8oZyqE8UhSkXS-1PHi:b-g1u078) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/FaceData.C#L78)):*
> ```
>   return std::accumulate( m_bface.cbegin(), m_bface.cend(), std::size_t(0),
>            []( std::size_t acc, const decltype(m_bface)::value_type& b )
>               { return acc + b.second.size(); } );
> ```

I looked up cppreference.com for more explanation about this, but I wasn't able to understand much of what was there. I'll probably talk to you more about this the next time we do a call/hangouts. I do get its purpose though. However, why do we not want nbfac to be passed in from ``Partitioner::createWorkers()``?

---

*[src/Inciter/Partitioner.C, line 1333 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oc2Ft4gTzNFA5ejbI:-L8oc2Ft4gTzNFA5ejbJ:b-392w1w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1333)):*
> ```
>     // Generate map associating new(value) to file(key) node ids for this chare
>     decltype(m_linnodes) newnodes;
>     const auto& chfilenodes = tk::cref_find( m_chfilenodes, cid );
> ```

Any specific reason why we got rid of searching in m_linnodes, and are looking in m_chfilenodes instead?

---

*[src/Inciter/Partitioner.C, line 1337 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8od7vv1pg9v-o240ZN:-L8od7vv1pg9v-o240ZO:bf61apn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1337)):*
> ```
> 
>     // Generate face set for this chare for potentially faster searches
>     tk::UnsMesh::FaceSet faceset;
> ```

Will this be very memory intensive (or am I being too careful)?

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Reviewed 2 of 7 files at r28.
Review status: 15 of 23 files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

*[src/Inciter/FaceData.C, line 78 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oZyqE8UhSkXS-1PHh:-L8oq4ey6z8QGUB5M6dp:bwh4w7e) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/FaceData.C#L78)):*

I looked up cppreference.com for more explanation about this, but I wasn't able to understand much of what was there. I'll probably talk to you more about this the next time we do a call/hangouts. I do get its purpose though. However, why do we not want nbfac to be passed in from ``Partitioner::createWorkers()``?
</blockquote></details>

There is no need to store it, because it can be computed as the sum of the sizes of the maps in `m_bface`, which is what the above `std::accumulate` does. It basically takes a pair of iterators, to the beginning and to the end of of a container (`m_bface`), the initial value of an accumulator, 0, and a lambda, an inline function that takes two arguments: `acc` the current (running) value of the accumulator, and `b`, which serves as iterating through `m_bface` by stepping through its pairs of int and std::vector (as a range-for would do over the map), and the body adds `b.second.size()` to the accumulator and returns it for the next iteration. The result is the sum of the sizes of the inner vectors.

I could have written it as a double-loop as well, but I just did a very similar `std::accumulate` call in `DG` previously, so I just adopted that one here. There, in `DG::comfac`,  the reason for using `std::accumulate` was because that can be written inside of an `Assert` in a single instruction, which then completely gets ignored in RELEASE. With a loop that would be harder to do.

For these algorithms, the most helpful for me is to look at the _Possible implementation_.

---

*[src/Inciter/Partitioner.C, line 1333 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oc2Ft4gTzNFA5ejbI:-L8otIEF8GugdSJvAPVX:bwdvdze) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1333)):*

Any specific reason why we got rid of searching in m_linnodes, and are looking in m_chfilenodes instead?
</blockquote></details>

It is the same data as was before. Before it was called `chlinnodes`, now I call it `newnodes`. I think this somewhat a better name because for a file id (key) it gives the new node id (value). (It is generated the same way as before as `for (auto i : chinpoel) newnodes[ tk::cref_find(chfilenodes,i) ] = i;`

---

*[src/Inciter/Partitioner.C, line 1337 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8od7vv1pg9v-o240ZN:-L8ouJC38QsfxFGsk4ik:b-xzl7sk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1337)):*

Will this be very memory intensive (or am I being too careful)?
</blockquote></details>

I don't think so. The performance problem with the earlier version was that for every physical boundary face we had an inner loop that checked every face in the partition. That was the linear search we also talked about, so this was expected. (That does not scale because it has `O(NxM)` complexity, where`N` is the number of physical boundary faces and `M` is the number of all faces. The product, and thus the computational cost grows faster than linear as the size of the mesh grows.) Now this linear search is replaced with a `O(const)`-complexity search using a hash-map in `FaceSet`. The resulting algorithm has four searches but all four with hash maps, so none of their complexity depend on the sizes of the data, which should scale great with increasing mesh size. All of this, however, only scales great if the hash maps/sets are relatively large. For small data, i.e., less than a few hundred (or even a thousand) a linear search through vectors will perform best because that is guaranteed to be contiguous in memory so it incurs less cache misses than hash maps and sets. But for larger data, the better-complexity algorithm will win.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

adityakpandare:
Reviewed 4 of 10 files at r20, 1 of 3 files at r25, 2 of 7 files at r28, 2 of 2 files at r29.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/Inciter/FaceData.C, line 78 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oZyqE8UhSkXS-1PHh:-L8qZFWzA4NxL4kuDMlv:bh4nrph) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/FaceData.C#L78)):*

There is no need to store it, because it can be computed as the sum of the sizes of the maps in `m_bface`, which is what the above `std::accumulate` does. It basically takes a pair of iterators, to the beginning and to the end of of a container (`m_bface`), the initial value of an accumulator, 0, and a lambda, an inline function that takes two arguments: `acc` the current (running) value of the accumulator, and `b`, which serves as iterating through `m_bface` by stepping through its pairs of int and std::vector (as a range-for would do over the map), and the body adds `b.second.size()` to the accumulator and returns it for the next iteration. The result is the sum of the sizes of the inner vectors.

I could have written it as a double-loop as well, but I just did a very similar `std::accumulate` call in `DG` previously, so I just adopted that one here. There, in `DG::comfac`,  the reason for using `std::accumulate` was because that can be written inside of an `Assert` in a single instruction, which then completely gets ignored in RELEASE. With a loop that would be harder to do.

For these algorithms, the most helpful for me is to look at the _Possible implementation_.
</blockquote></details>

Nice! Thanks for the explanation. That's a great idea for an Assert btw; very cleanly removes the debug code from release-builds!

---

*[src/Inciter/Partitioner.C, line 1333 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8oc2Ft4gTzNFA5ejbI:-L8qZbkB4hWec9THtBZk:b-7v9qox) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1333)):*

It is the same data as was before. Before it was called `chlinnodes`, now I call it `newnodes`. I think this somewhat a better name because for a file id (key) it gives the new node id (value). (It is generated the same way as before as `for (auto i : chinpoel) newnodes[ tk::cref_find(chfilenodes,i) ] = i;`
</blockquote></details>

My bad. I thought we were previously using m_linnodes. Never mind.

---

*[src/Inciter/Partitioner.C, line 1337 at r28](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8od7vv1pg9v-o240ZN:-L8qZsVz3cge156EFGTs:bbb71ps) ([raw file](https://github.com/quinoacomputing/quinoa/blob/fe76dcf65224b756b02661bac503b7aa9bee6cc1/src/Inciter/Partitioner.C#L1337)):*

I don't think so. The performance problem with the earlier version was that for every physical boundary face we had an inner loop that checked every face in the partition. That was the linear search we also talked about, so this was expected. (That does not scale because it has `O(NxM)` complexity, where`N` is the number of physical boundary faces and `M` is the number of all faces. The product, and thus the computational cost grows faster than linear as the size of the mesh grows.) Now this linear search is replaced with a `O(const)`-complexity search using a hash-map in `FaceSet`. The resulting algorithm has four searches but all four with hash maps, so none of their complexity depend on the sizes of the data, which should scale great with increasing mesh size. All of this, however, only scales great if the hash maps/sets are relatively large. For small data, i.e., less than a few hundred (or even a thousand) a linear search through vectors will perform best because that is guaranteed to be contiguous in memory so it incurs less cache misses than hash maps and sets. But for larger data, the better-complexity algorithm will win.
</blockquote></details>

That's true. It definitely has gotten faster. :+1: Thanks again for the explanation.

---

*[src/Inciter/Partitioner.C, line 1364 at r29](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216#-L8qaRG64_dJp8k3Xq8P:-L8qaRG64_dJp8k3Xq8Q:b-m31yol) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2e6dd15d8b95e8d5f6bab394f5b0be189f0fbc/src/Inciter/Partitioner.C#L1364)):*
> ```
>             for (std::size_t i=0; i<3; ++i) chtriinpoel.push_back( t[i] );
>             // store physical boundary face id associated to sideset id
>             chbface[ ss.first ].push_back( cnt++ );
> ```

Looks good! I can see how this is a huge improvement in terms of computation costs from the previous loop-inside-loop implementation. :+1:

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Reviewed 1 of 4 files at r1, 1 of 9 files at r3, 7 of 18 files at r19, 3 of 10 files at r20, 1 of 3 files at r25, 4 of 7 files at r28.
Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

jbakosi:
Reviewed 5 of 5 files at r30.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/216)*

PR #217, opened 2018-03-05
Put in signal handlers for various signals to all executables

jbakosi:
This is so we have a chance at debugging asynclogic (and other) bugs happening randomly. Most importantly, catching `SIGSEGV`, so we hopefully get a trace for a segfault.

Note also, that the default behavior of attempting to generate a trace is no changed from only in DEBUG to independent of the `CMAKE_BUILD_TYPE`.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/217)


src/Base/Exception.C
```diff
@@ -96,14 +92,14 @@ Exception::saveTrace() noexcept
 //!   Requires stdio.h, execinfo.h.
 // *****************************************************************************
 {
+#ifndef HOST_OS_ALPINE
```
junghans: I would really like to see a comment why ALPINE doesn't have it....

jbakosi: It does, but segfaults on calling `backtrace()`, and I have not clue how to fix it. I tried playing with the buffer length it takes its 2nd argument, as well as putting the buffer in global scope, making it static, etc. I just cannot fix it. So on alpine we cannot call it.

jbakosi: Also, https://www.freshports.org/devel/libexecinfo says it's quick'n'dirty, so it is probably broken.

junghans: Still a comment about alpine OS would be nice

junghans: I know the commit message has it, but I mean right here in the code.

jbakosi: Ah, okay, I see. I will put in a comment there then. Thx.

PR #218, opened 2018-03-07
Make DG BCs more general and reuse code

jbakosi:
Also ties BC specification to user input based on input control file. See also commit messages. Still missing is making this a little more robust based on crap user input, by, e.g., parse-time error checking - left for another day.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/218)


adityakpandare:
Reviewed 10 of 11 files at r1.
Review status: 10 of 11 files reviewed at latest revision, 3 unresolved discussions.

---

*[regression/inciter/transport/GaussHump/gauss_hump.q, line 34 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71aS1dAARNg1Haf3Ut:-L71aS1dAARNg1Haf3Uu:b-a7zyf3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/regression/inciter/transport/GaussHump/gauss_hump.q#L34)):*
> ```
>     end
>   end
> ```

Whoa!! This is awesome! I've seen something similar in OpenFoam. 

---

*[src/Control/Inciter/InputDeck/Grammar.h, line 446 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71bLDR-7GdOL4mAjDA:-L71bLDSh941E2Iw5JCd:b-ndrdge) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/Control/Inciter/InputDeck/Grammar.h#L446)):*
> ```objc
>                            bc< kw::bc_sym, tag::compflow, tag::bcsym >,
>                            bc< kw::bc_inlet, tag::compflow, tag::bcinlet >,
>                            bc< kw::bc_outlet, tag::compflow, tag::bcoutlet > >,
> ```

So this makes sure that when we write code for compressible Euler equations, these keywords will be valid?

---

*[src/Control/Inciter/Options/BC.h, line 52 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71bjmU1woBm9PkaS62:-L71bjmU1woBm9PkaS63:b-m0l45y) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/Control/Inciter/Options/BC.h#L52)):*
> ```objc
>           { kw::bc_outlet::string(), BCType::OUTLET } } ) {}
> };
> ```

Could you explain to me what this part does? Does this have to do with the communication of the solution at the chare-boundary faces?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218)*

adityakpandare:
That's made the code very clean now. I really like the way we don't have repeating for-loops for the BCs. Thank you!

---

Reviewed 1 of 11 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-:-L71fd9G66-7ottkkD8U:b-757g1t)*

adityakpandare:
Thanks for your responses Jozsef. I think you're good to merge this into develop.

---

Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[src/Control/Inciter/InputDeck/Grammar.h, line 446 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71bLDR-7GdOL4mAjDA:-L75YHPnuFEvHNh5VvaZ:b-90g5i7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/Control/Inciter/InputDeck/Grammar.h#L446)):*

Yep! This is already valid, and it should be used very similarly as in `DGTransport.h`.

There is/was also another way of setting BCs in the input file but we have not tested/used that recently, which allows specifying, e.g., an externally specified value for a inhomogeneous Dirichlet or Neumann BCs, with something like this:
```
    bc_dirichlet
            # setid  comp  val
      sideset     2     2   0.0 end  # inflow: velx = 0
      sideset     2     3   0.0 end  #         vely = 0
      sideset     2     4   1.0 end  #         velz = 1

      sideset     3     2   0.0 end  # outflow:velx = 0
      sideset     3     3   0.0 end  #         vely = 0

      sideset     8     2   0.0 end  # cyl:    velx = 0
      sideset     8     3   0.0 end  #         vely = 0
      sideset     8     4   0.0 end  #         velz = 0
  end
```
This works with the current grammar, i.e., it is parsed correctly, but I'm not sure if we still have the code to use this and to set BCs based on this, but we should probably revive/improve it sometime.

One can catch rather complex errors during parsing, i.e., very early on, which is what I meant by not being very robust yet, but it can be.
</blockquote></details>

This is quite nice to have.

---

*[src/Control/Inciter/Options/BC.h, line 52 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71bjmU1woBm9PkaS62:-L75Yh-1nbC9pvjwxNNs:bnpc4ni) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/Control/Inciter/Options/BC.h#L52)):*

No, if I understand your question correctly, I don't think so.

This is a class that specializes and initializes a generic base, `tk::Toggle`, used to model a switch with multiple states. This particular one has 3 states (so far). The important feature is to associate keywords (used for parsing) to typesafe enums (which cannot be confused at compile-time with other enums, so they are unique to a given Option switch) and back. The input file parser uses these options/switches in a generic fashion: parses a keyword, then sets a particular state for a given option using various checks during both compile-time and parsing/run-time. For example, if a particular option is being parsed, e.g., `BC`, the allowed states are checked that they can only be selected from the available ones listed here, and error messages are (can be) generated accordingly. Also, once the state is chosen for an option, we don't store the string or the keyword (that would mean any string or keyword could be used to set as a valid option), instead we store the enum, which is not only very small (usually 8 bit) but also typesafe, so attempting to write grammar code that would accept assigning a state for and option that isn't compatible with it the compiler will not allow it. (The error message are not very pleasant at this time, but this should improve with concepts in C++20 and after.) All this state parsed from the command line and from the control file is stored in `g_inputdeck` which is then made available to all PEs and assumed read-only after it has been filled by the parser. This is a hierarchical tagged tuple, capable of storing an arbitrary hierarchy of arbitrary types grouped in arbitrary categories internally, that can be accessed by compile-time tags. Access, is resolved at compile-time, so it is immediate, no searches, loops,  and typesafe, and access errors generate compile-time errors, unless the access is to a runtime-allocated data structure, e.g., a vector.
</blockquote></details>

Thanks a lot for the explanation. I'm not sure I understand all of it, but I do understand a lot of it; and it also answers my question.

---

*[src/PDE/Transport/DGTransport.h, line 336 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71e57oTnkTNNUj-OTn:-L75Yu_hlDo8hIi-kOiz:b-olgkiv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/PDE/Transport/DGTransport.h#L336)):*

No particular reason. I figured it is a single argument from which we get both the centroid coordinates and the face normals as well. It's (maybe) a little less typing for the function signature -- we can totally switch this back to whatever way this is usually done. Do you want me to switch to, e.g., the following?
```
upwindFlux( std::size_t f, tk::real xc, tk::real yc, tk::real zc,
            const std::array< tk::real, 3 >& fn,
            const std::array< std::vector< tk::real >, 2 >& u ) const
```
</blockquote></details>

Umm.. Let it be the way it is right now. It's not costing any extra memory/computation right. I'm leaning towards extracting geoFace for the face f, and send that one alone to upwindFlux(). But I don't want to waste your time, since this is just something I'm used to. So let's leave this the way it is right now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-:-L75ZPkAZW5rKscyTbNM:balwrau)*

jbakosi:
Thanks for the extra set of eyes!

---

Review status: :shipit: all files reviewed at latest revision, all discussions resolved, all commit checks successful.

---

*[src/Control/Inciter/Options/BC.h, line 52 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71bjmU1woBm9PkaS62:-L75amNrzjbSGzcMwIYB:b7c4i97) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/Control/Inciter/Options/BC.h#L52)):*

Thanks a lot for the explanation. I'm not sure I understand all of it, but I do understand a lot of it; and it also answers my question.
</blockquote></details>

Sure. We can also talk more about this stuff later if needed.

---

*[src/PDE/Transport/DGTransport.h, line 336 at r1](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-L71e57oTnkTNNUj-OTn:-L75b0zssbWz-W4OEo9b:b-jp2051) ([raw file](https://github.com/quinoacomputing/quinoa/blob/70f0db020a50c700b9e303a84ae93539569f16be/src/PDE/Transport/DGTransport.h#L336)):*

Umm.. Let it be the way it is right now. It's not costing any extra memory/computation right. I'm leaning towards extracting geoFace for the face f, and send that one alone to upwindFlux(). But I don't want to waste your time, since this is just something I'm used to. So let's leave this the way it is right now.
</blockquote></details>

Okay. This is pretty simple to modify later if we decide that we like it some other way. (I do also kind of like sending just the geoface(f), but that would involve some copies, so let's not bother.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/quinoacomputing/quinoa/218#-:-L75agJfuErbf9zvdodq:b-jmqgas)*

PR #231, opened 2018-04-17
Simplify migration of variant

jbakosi:
Thanks to @nilsdeppe for improving on this.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/231)

### DIFF
--- a/src/Base/PUPUtil.h
+++ b/src/Base/PUPUtil.h
@@ -15,9 +15,8 @@
 #include <array>
 
 #include "NoWarning/optional.h"
-
+#include "NoWarning/variant.h"
 #include "NoWarning/pup_stl.h"
-
 #include "CharmUtil.h"
 
 //! Extensions to Charm++'s Pack/Unpack routines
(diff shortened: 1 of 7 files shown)


rfbird:
Reviewed 1 of 7 files at r1.
Review status: 1 of 7 files reviewed at latest revision, all discussions resolved.

---

*[src/Base/PUPUtil.h, line 210 at r1](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231#-LAK9KFm6CATSeSYBGLO:-LAK9KFm6CATSeSYBGLP:b4w9gut) ([raw file](https://github.com/quinoacomputing/quinoa/blob/57f1bf7aca5fe81f62002e8387822d9754d5fd25/src/Base/PUPUtil.h#L210)):*
> ```objc
> //! \param[in] var boost::variant< Ts... > of arbitrary types to pack/unpack
> template <class T, class... Ts>
> char pup_helper(int& index, const int send_index, PUP::er& p,
> ```

What's the deal with char here (vs saying, casting at a later stage)? Also, does it return '0' no matter what?

---

*[src/UnitTest/tests/Base/MigratedTypes.h, line 45 at r1](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231#-LAK9ekT0gEgeim3_GVm:-LAK9ekT0gEgeim3_GVn:b-8hntze) ([raw file](https://github.com/quinoacomputing/quinoa/blob/57f1bf7aca5fe81f62002e8387822d9754d5fd25/src/UnitTest/tests/Base/MigratedTypes.h#L45)):*
> ```objc
>                                              tag::age,   int,
>                                              tag::email, std::string >;
> using Variant = boost::variant< int, double >;
> ```

Again, I know you don't mind it, but a change going from generic variant to specifically boost variant seems like it could be an odd decision to me 

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231)*

rfbird:
Reviewed 6 of 7 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231)*

jbakosi:

---

*[src/Base/PUPUtil.h, line 210 at r1](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231#-LAK9KFm6CATSeSYBGLO:-LAL9Em59jndy4sQavl2:b-561paj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/57f1bf7aca5fe81f62002e8387822d9754d5fd25/src/Base/PUPUtil.h#L210)):*

What's the deal with char here (vs saying, casting at a later stage)? Also, does it return '0' no matter what?
</blockquote></details>

That `char` is not really used or significant here. Since`std::initializer_list` cannot be templated with a `void`, `pup_helper` is chosen to return a `char`. `std::initializer_list` is only used to write a compile-time loop that calls `pup_helper` at each iteration which only does anything if the location of the type matches the type/value held by the variant puped and that something is: get the value for sending and set the value for receiving . @nilsdeppe can probably explain this better though or correct me if I'm wrong, since he wrote it. ;-) Ingenious, isn't it?

---

*[src/UnitTest/tests/Base/MigratedTypes.h, line 45 at r1](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231#-LAK9ekT0gEgeim3_GVm:-LAL58FBBVu4oDjK9IHJ:bdff3mn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/57f1bf7aca5fe81f62002e8387822d9754d5fd25/src/UnitTest/tests/Base/MigratedTypes.h#L45)):*

Again, I know you don't mind it, but a change going from generic variant to specifically boost variant seems like it could be an odd decision to me 
</blockquote></details>

Yeah. `std::variant` is C++17, hence boost. I think switching from boost to std at some point will be easy.

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231)*

adityakpandare:
Reviewed 7 of 7 files at r1.

---

*[src/Inciter/SchemeBase.h, line 143 at r1](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231#-LATOsoc1U5joFvRO6BF:-LATOsoc1U5joFvRO6BG:bd66zvt) ([raw file](https://github.com/quinoacomputing/quinoa/blob/57f1bf7aca5fe81f62002e8387822d9754d5fd25/src/Inciter/SchemeBase.h#L143)):*
> ```objc
>     //! \param[in,out] p Charm++'s PUP::er serializer object reference
>     void pup( PUP::er &p ) {
>       p | proxy;
> ```

This is where you've gotten rid of the tuple ```v```. Thank's for explaining this to me!

---


*Comments from [Reviewable](https://beta.reviewable.io/reviews/quinoacomputing/quinoa/231)*

nilsdeppe:
@jbakosi is right about the `char` thing. The `std::initializer_list` cannot be filled with `void` (this might be different in C++17, I haven't tested exactly what regular void enables). I'm also in favor of using Boost::variant rather than rolling one yourself, and it looks like tk::variant was just a wrapper around boost::variant that significantly increased memory used and complexity. so :+1: from me :)

jbakosi:
I think @rfbird did not mean pulling one's own variant only nicer way of abstracting away the use of `boost::` vs `std::variant`. Right now `boost::variant` is somewhat scattered around the code, not too many places though. Abstracting that away with a single `typedef` would make it easier to switch in the future, but I don't care enough about it. ;-)

`tk::Variant` was only a PUP helper.

Thanks, again, @nilsdeppe !

nilsdeppe:
Ah, I see. Depending on the STL implementation you may be better off with Boost version later on :) This is true if you don't need `shared_ptr` to be thread safe. You can make the boost version non-threadsafe, which is faster, but you can't do that for the STL version. I'm not sure if any such issues will show up for variant.

You're welcome! :)

PR #235, opened 2018-04-23
Distributed-memory parallel, asynchronous initial mesh refinement before partitioning

jbakosi:
Distributed-memory parallel, asynchronous initial mesh refinement before partitioning
========================
Example user configuration
---------------------------
Setting up 3 equation systems: compressible flow (5 components, single material), and 2 transport equation systems each with 3 scalar equations. Each system must configure a _dependent variable_ (`depvar`), which must be unique among the systems. This `depvar` is then used to configure _refinement variables_ (`refvar`) in the `amr...end` block. Specifying the refinement variables is done by a letter (a previously configured depvar), and an optional number, denoting the position of the scalar variable relative to the first depvar in its equation system. Omitting the number means the first depvar.

```shell
inciter
  ...
  compflow
    ...
    depvar c
    ...
  end

  transport
    ...
    ncomp 3
    depvar t
    ...
  end

  transport
    ...
    ncomp 3
    depvar h
    ...

  end

  amr
    # Configure an arbitrary list of initial refinement steps: only 'uniform' and 'ic'
    # are currently implemented: 'uniform' means uniform refinement, 'ic'
    # means refinement based on the numerical error as the initial condition is
    # represented on the mesh given.  Note that the order matters, the
    # refinement steps are applied in the order specified.
    initial uniform
    initial uniform
    initial ic
    initial uniform
    initial ic
    initial uniform
    initial ic
    initial ic
    initial uniform
    #initial ic
    #initial ic
    #initial uniform
    # refine initial mesh based on the union of errors from scalar components
    # c0, t2, and h1
    refvar c t2 h1 end
    # use the 'jump' error indicator for computing errors
    error jump
  end
  ...
end
```

High-level description of the algorithm
-------------------------
To perform a _single_ step of potentially multiple initial refinement steps (e.g., as configured above), each PE performs the following steps:

* `Partitioner::Partitioner` reads in the mesh in a distributed parallel fashion (both connectivity and coordinates). If the initial mesh refinement list is empty, skip to `Partitioner::partchare`.

* `Partitioner::partref`, calls the mesh partitioner (currently Zoltan2) and performs partitioning of the mesh. The number of partitions here always equals the number of PEs. The result is an array that assigns a PE (i.e., MPI rank) to each mesh element on each PE. Since there is no guarantee that the cells read on a PE will be assigned to the same PE, a reshuffling step needs to ensure that the cells are communicated to their target PEs.

* `Partitioner::DistributePE`, performs distribution of mesh cells to their target PEs after partitioning to PE parts. This is done after a categorization step, `Partitioner::categorize()`, which assigns cells (given by the 4 global node IDs of their vertices) to target PEs. `Partitioner::distributePE()` then distributes the cells among PEs: it keeps those cells that got assigned to the PE the cell happens to be on, and packs up those that are not on the PE and sends them to their target PE. The receive-side is `addPeMesh()`. Note that both the mesh connectivity as well as the node coordinates are sent, this latter as a map that associates the coordinates to global node IDs. The receive side stores the mesh connectivity as well as the coordinates, and sends an acknowledgment back to the sender that the message has been received. Collecting all these receipts are used to trigger a global reduction to the single chare `Transporter`, to its `reductiontarget` entry method `refdistributed()`, which is then used to detect global completion of this mesh shuffling step. This is a global synchronization step. The next step is collecting unique edges on PE boundaries.

* `Partitioner::bndEdges` generates a unique set of boundary edges of the mesh chunk on every PE. These are edges that are potentially shared along PE boundaries (but also contain the edges on physical boundaries - it is easier to generate the union of the physical and PE boundaries based on the derived data structure, `esuelTet`, we have easy access to at this point). Note that potentially multiple PEs can share the same edge. An edge is uniquely defined by the global node IDs of its end-points, see `tk::UnsMesh::Edge`. The unique set of boundary edges are then sent to _all_ other PEs. This is an all-to-all communication step, by calling `thisProxy.addBndEdges()`, in a broadcast fashion on all PEs. This is potentially a communication-expensive step, which sends a set of edges (node ID pairs), but appears necessary, since at this point on each PE we have a new mesh chunk and know nothing about what other PEs a given boundary edge can be shared with. Once the receiving side has heard from all other PEs (including itself), we continue (on each PE separately, thus overlapping among PEs), by calling `Partitioner::refine()`.

* `Partitioner::refine`, does a single initial mesh refinement step. It first converts the node coordinate map to a flat vector which then will be modified by the refiner. Mesh refinement is done by calling `errorRefine()` for `initial ic`, and for `uniformRefine()` for `initial uniform`.

* _Mesh refinement and updating the mesh_. The actual mesh refinement is performed by the AMR library, instantiated by its `AMR::mesh_adapter_t` for both `ic` or `uniform` mesh refinements. The error-based refinement, simply hard-codes the initial conditions for continuous Galerkin at this time. The same mesh update is followed by both types of initial refinement steps.

* `Partitioner::updateMesh`, updates the necessary data structures after a mesh refinement step:
  - The new mesh connectivity is obtained from the refiner.
  - Coordinates are generated for each newly added node that has been added to an edge by the refiner. The new coordinates are simply appended in `Partitioner::m_coord`, which effectively assigns a new local ID to the added node.
  - Local IDs for newly added nodes are added to `m_lid` and `m_gid`, which are global->local and local->global node index maps.
  - The new node coordinates are also added to the coordinate map, which associates node coordinates to global node IDs.
  - The coordinates are also added to a map (`m_edgenode`) that associates coordinates to global node IDs, and associates all this to edges, given by old global (parent) node IDs.
  - A new global ID is assigned to the newly added node by combining two hashes obtained by `std::hash`, appropriately specialized to the ID type and combined with `boost::hash_combine`, see `tk::UnsMesh::EdgeHash`.
  - Finally, the mesh connectivity (given by global node IDs) is updated using  the above data.
  - All of this happens independently on each PE and the hash guarantees unique global node IDs independently on each PE, even if multiple PEs assign to the same physical edge, since the hash is generated based on the parent global node IDs, which are assumed to be unique to start with. (Uniqueness is verified later on for the price of some extra communication.)

* _Partitioner::addRefBndEdges_. After the mesh and associated data structures have been updated after refinement, the newly added nodes to only the boundary edges are communicated to the PE the edge is shared with. This is done by collecting all (and only those) boundary edges that a given PE shares with another PE and the edge (given by the unique global parent node IDs) and associated newly assigned global ID and coordinates are sent to target PEs (in a point-to-point fashion). The receiving side stores this information (buffers it up for later verification). A similar acknowledgment step followed by a global reduction as before are used to ensure global completion of this step, resulting in a call to `Transporter::matched`. This is a second global synchronization point during a single step of initial mesh refinement. `Transporter::matched` then initiates `Partitioner::nextref()`.

* `Partitioner::nextref`, does two things:

  1. Ensures that the same globally unique node ID has been assigned by multiple PEs to the same physical PE-boundary edge. This is done by comparing what a PE assigned to a given edge to the (same physical) edge that was received from another PE that shares the edge. Not only do we compare the new node IDs but also the coordinates of the newly added node. Note that neither comparing the ID nor the coordinates would be strictly necessary (but this helps ensuring correctness). The price of this is some extra communication when calling `Partitioner::addRefBndEdges()`.

  2. Those edges that are shared among multiple PEs and at least one PE assigned a new node to, all other PEs that share the same edge must correct their part of the mesh in the vicinity of the edge. These edges are collected in `nextref()`, and another step of refinement is done in `Partitioner::correctRefine()`, tagging only those edges that need correction. This, of course, happens independently on each PE. The result is that if an edge was refined by a PE among potentially multiple PEs that share the edge, all PEs that share the edge refine the edge. In other words, we opt for more refinement and thus smaller numerical error in the vicinity of a refined edge.

* When the correction step above is done (if needed), a single step of the potentially multiple steps in initial mesh refinement is complete and we evaluate if we need to do another one. If so, `Partitioner::partref()` is called again and the same series of steps are repeated. If there is no more refinement steps, `Partitioner::finishref()` is called, which invokes a global reduction that does a global sum on the (current/final) number of mesh elements across all PEs. This is targeted at the reductiontarget `Transporter::refined()`, which continues by issuing a broadcast to `Partitioner::partchare()`, which partitions the mesh into `nchare` chares.

* `Partitioner::partchare()`, which used to be called simply `Partitioner::partition()`, calls the mesh partitioner one last time (or the first time if no initial mesh refinement has been performed), and divides the mesh into potentially more partitions than the number of PEs.

* `Partitioner::distributePE()`. The partitioning step to chares above is followed by a similar mesh reshuffling step as was done as part of a single step of initial mesh refinement, but using `Partitioner::distributeCh()` instead of `Partitioner::distributePE()`, with the receiving side, `Paritioner::addChMesh()` (as opposed to `Paritioner::addPeMesh()`). `DistributeCh()` is slightly more complex than `DistributePE()`, because the former has to work with potentially multiple chares on each PE (and with more mesh chunks categorized by chares), but in principle its job is the same: retain the elements targeted to a given PE, and package up those mesh cells that are targeted to other PEs. The data to be sent is the mesh connectivity with global node IDs, node IDs, and their coordinates, all this data categorized by chares.

* `Partitioner::flatten()`. The reshuffling step above, after a global synchronization in the reductiontarget `Transporter::distributed()` is followed by a broadcast to `Partitioner::flatten()`, whose job is the same as before: prepare the data structures for global distributed mesh node reordering. From this point on the functionality is the same as before, but simplified because edges now are not needed, since the mesh refinement has already been done.

Another difference is how the node coordinates are passed to the workers, `Discretization`. This data is now passed to the workers, via the `discInsert()` call, which parameterizes the `Discretization` constructor under the hood. Thus `Discretization` now does not have to read the node coordinates.

Also re-enabled regression tests that were previously disabled, updated the parallel ones with new baselines, and added a new one that exercises multiple levels uniform refinement.

Todo (see #236)
-------
1. Non-uniform refinement produces non-positive Jacobians even in serial. (This is probably an easy fix for @rfbird in the AMR object.)
2. All this code will be abstracted away to another class, instead of it being in `Partitioner`.
3. The `IC`/error-based initial refinement currently still hard-codes the field obtained from the initialization for continuous a Galerkin PDE. This will be done in a discretization-independent fashion so it also works for a discontinuous Galerkin discretization.
4. All this should be made to work for a union of all the scalar `depvars` given in the  `refrvar...end` block.
5. Setting Dirichlet boundary conditions using continuous Galerkin (`DiagCG` and `MatCG`) discretizations, does not yet work if the mesh is refined: newly added nodes will have to inherit the BCs from their parent's edge.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/235)


rfbird:
Reviewed 26 of 35 files at r1, 76 of 80 files at r3, 2 of 2 files at r4, 3 of 3 files at r5.

---

*[src/Base/ContainerUtil.h, line 38 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LAtcM6bD8-pVrflKo0p:-LAtcM6c0DWjeQRAkYxo:b-rk6d4w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2535cf08a1e8f488d8d9178aebdcf091461578af/src/Base/ContainerUtil.h#L38)):*
> ```objc
> template< class Container >
> Container
> cunique( const Container& src )
> ```

I'm a bit worried this shortening of copy to c is going to conlfict with things like const in cref_find

---

*[src/Control/Inciter/InputDeck/Grammar.h, line 252 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMKMoZ72qAO1CBwF4K:-LBMKMoZ72qAO1CBwF4L:b-mxr7i2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Control/Inciter/InputDeck/Grammar.h#L252)):*
> ```objc
>   };
> 
>   //! Rule used to trigger action
> ```

What's the deal with this repeated comment?

---

*[src/Inciter/MeshReader.C, line 36 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMLmkV5WZ2_86Z4Kg-:-LBMLmkV5WZ2_86Z4Kg0:b-dbez36) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/MeshReader.C#L36)):*
> ```
> 
>   // Compute extents of element IDs of our mesh chunk to read
>   auto npes = static_cast< std::size_t >( n );
> ```

Is it weird to static cast into an auto? Half explicit, half deduced? Seems like we're explicitly sending ints (presumably to avoid the 64bit size_t send?) 

---

*[src/Inciter/AMR/mesh_adapter.C, line 85 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMOb-_7pk2QX7FVRFk:-LBMOb-_7pk2QX7FVRFl:bl0alyi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/AMR/mesh_adapter.C#L85)):*
> ```
> 
>     void mesh_adapter_t::error_refinement( const std::vector< edge_t >& edge,
>                                            const std::vector< real_t >& crit )
> ```

Is crit here criteria? I'm really not convinced saving those 4 characters is worth it ^_^ x

---

*[src/Mesh/Reorder.C, line 194 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMP94Z-09CccdBf9X5:-LBMP94Z-09CccdBf9X6:bbxirdn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Mesh/Reorder.C#L194)):*
> ```
> 
> bool
> positiveJacobians( const std::vector< std::size_t >& inpoel,
> ```

Don't we already have a Jacobian function somewhere that this could call?

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/235)*

jbakosi:
Review status: all files reviewed at latest revision, 5 unresolved discussions, some commit checks failed.

---

*[src/Base/ContainerUtil.h, line 38 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LAtcM6bD8-pVrflKo0p:-LBQgpHN-vkivPPAWNfz:bkj2uyh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2535cf08a1e8f488d8d9178aebdcf091461578af/src/Base/ContainerUtil.h#L38)):*

I'm a bit worried this shortening of copy to c is going to conlfict with things like const in cref_find
</blockquote></details>

Agreed. Will fix it.

---

*[src/Control/Inciter/InputDeck/Grammar.h, line 252 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMKMoZ72qAO1CBwF4K:-LBQhp3LAShACQMycEK1:bfkgcd8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Control/Inciter/InputDeck/Grammar.h#L252)):*

What's the deal with this repeated comment?
</blockquote></details>

I don't understand. That comment always refers to an empty struct inheriting from `pegtl::success`, and is always followed by a specialization of the given template struct. This is a peculiarity of PEGTL v2.0 to which we updated a few months ago.

---

*[src/Inciter/MeshReader.C, line 36 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMLmkV5WZ2_86Z4Kg-:-LBQjJmW9ZRFrcAyvemr:b-gemb92) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/MeshReader.C#L36)):*

Is it weird to static cast into an auto? Half explicit, half deduced? Seems like we're explicitly sending ints (presumably to avoid the 64bit size_t send?) 
</blockquote></details>

No. The indices (here: cell IDs) are always of type `std::size_t` but ckMyPe() and NumPes() (from Charm++) both return `int`, so we have to cast and I deemed to be the least obtrusive/confusing to just simply cast both at the beginning of the arithmetic involving them that follows. As to why not
```
std::size_t npe = n;
```
that will cause a compiler warning on implicit conversion (and loosing the sign), and why not
```
std::size_t npe = static_cast< std::size_t >( n );
```
because it is longer than 
```
auto npes = static_cast< std::size_t >( n );
```
and more error prone, because it repeats the type which must be kept consistent.

---

*[src/Inciter/AMR/mesh_adapter.C, line 85 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMOb-_7pk2QX7FVRFk:-LBQlL_q4o1jpEH6wPLo:b-c5hnqb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/AMR/mesh_adapter.C#L85)):*

Is crit here criteria? I'm really not convinced saving those 4 characters is worth it ^_^ x
</blockquote></details>

Yeah. ;-) Will fix it.

---

*[src/Mesh/Reorder.C, line 194 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMP94Z-09CccdBf9X5:-LBQlR2R1-o7uCuNljMc:bw7evjm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Mesh/Reorder.C#L194)):*

Don't we already have a Jacobian function somewhere that this could call?
</blockquote></details>

I don't believe so, but we probably could. (Also this goes through the whole mesh.)

I keep thinking about abstracting this away, because it appears at many places, indeed, but I have not found a good way to avoid unnecessary copies of function arguments _and_ manage to keep the results from the function available, which sometimes are not all needed. We can revisit this though sometimes.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/235)*

jbakosi:
Review status: 77 of 104 files reviewed at latest revision, 5 unresolved discussions.

---

*[src/Base/ContainerUtil.h, line 38 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LAtcM6bD8-pVrflKo0p:-LBbrCPY6zyRU8423YYt:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2535cf08a1e8f488d8d9178aebdcf091461578af/src/Base/ContainerUtil.h#L38)):*

Agreed. Will fix it.
</blockquote></details>

Done.

---

*[src/Control/Inciter/InputDeck/Grammar.h, line 252 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMKMoZ72qAO1CBwF4K:-LBbrVemBUI4HvVz3S6p:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Control/Inciter/InputDeck/Grammar.h#L252)):*

I don't understand. That comment always refers to an empty struct inheriting from `pegtl::success`, and is always followed by a specialization of the given template struct. This is a peculiarity of PEGTL v2.0 to which we updated a few months ago.
</blockquote></details>

Done.

---

*[src/Inciter/MeshReader.C, line 36 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMLmkV5WZ2_86Z4Kg-:-LBbrPquAO-Ma1JYedEg:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/MeshReader.C#L36)):*

No. The indices (here: cell IDs) are always of type `std::size_t` but ckMyPe() and NumPes() (from Charm++) both return `int`, so we have to cast and I deemed to be the least obtrusive/confusing to just simply cast both at the beginning of the arithmetic involving them that follows. As to why not
```
std::size_t npe = n;
```
that will cause a compiler warning on implicit conversion (and loosing the sign), and why not
```
std::size_t npe = static_cast< std::size_t >( n );
```
because it is longer than 
```
auto npes = static_cast< std::size_t >( n );
```
and more error prone, because it repeats the type which must be kept consistent.
</blockquote></details>

Done.

---

*[src/Inciter/AMR/mesh_adapter.C, line 85 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMOb-_7pk2QX7FVRFk:-LBbqvlIAsU_HLEfAOEd:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Inciter/AMR/mesh_adapter.C#L85)):*

Yeah. ;-) Will fix it.
</blockquote></details>

Done.

---

*[src/Mesh/Reorder.C, line 194 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/235#-LBMP94Z-09CccdBf9X5:-LBbrTJJ21oEURsKyLGX:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4f2c9307cbf55848d82d5aff5e5160e899e12514/src/Mesh/Reorder.C#L194)):*

I don't believe so, but we probably could. (Also this goes through the whole mesh.)

I keep thinking about abstracting this away, because it appears at many places, indeed, but I have not found a good way to avoid unnecessary copies of function arguments _and_ manage to keep the results from the function available, which sometimes are not all needed. We can revisit this though sometimes.
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/235)*

PR #238, opened 2018-04-26
Abstract away flux function for DG

jbakosi:
Lax-Friedrichs (LF) vs HLLC for now, but potentially more in the future. Note that HLLC is the default but LF should also be recognized and should work, but not tested. Does it make sense to add a regression test, just like vortical_flow_dg but exercising LF?

Also put in some minor cleanup in `DGTransport.h` and for parsing discretization parameters. Please see commit messages.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/238)


jbakosi:
Sure. I could add a LF regression test, but could you please do this? (You are more familiar with what is a correct solution and what might be a difference in the solutions between HLLC and LF.) Should we add a LF-version for Sod as we ll as vortical flow?

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[src/Control/Inciter/Types.h, line 61 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-LB6m9RW1xMCpR6fgqjd:-LB7DxiBC247FzqZQIxj:b-woyscw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c2c3651b61f120837b663b7ba974aa584c8d268d/src/Control/Inciter/Types.h#L61)):*

I like this new arrangement that the discretization scheme is in the tuple, and not by it's own as before.
</blockquote></details>

It was in a tuple before too, but now the discretization-related parameters are grouped together.

---

*[src/PDE/CompFlow/DGCompFlow.h, line 158 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-LB6o8df2R3U1IIk6M1d:-LB7Egoa1vhzdcDGgwvd:b-s1k66x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c2c3651b61f120837b663b7ba974aa584c8d268d/src/PDE/CompFlow/DGCompFlow.h#L158)):*

This is really awesome! One question: Why don't we put the repeating code with the ```intIntegral``` and ```bndIntegral``` outside the if-conditions, so that we wouldn't have to repeat it?
</blockquote></details>

We can't. The type `Flux` is defined for a given branch of the if-test (only). Any code outside of the branches will not see the definition of `Flux`. (An the definition of `Flux` cannot be outside, because then it can only be defined one way or another at compile-time, so we cannot switch between two definitions based on a parameter (what the user configured) at runtime.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-:-LB7DFfHDDuJYJ5WkmiW:bs508rl)*

adityakpandare:
Yes I'll do this. I think one for vortical flow is sufficient, unless you want me to put one for Sod in too. 

---

---

*[src/PDE/CompFlow/DGCompFlow.h, line 158 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-LB6o8df2R3U1IIk6M1d:-LB8EOU-9RXLVBQd_oCQ:b-6lw87w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c2c3651b61f120837b663b7ba974aa584c8d268d/src/PDE/CompFlow/DGCompFlow.h#L158)):*

We can't. The type `Flux` is defined for a given branch of the if-test (only). Any code outside of the branches will not see the definition of `Flux`. (An the definition of `Flux` cannot be outside, because then it can only be defined one way or another at compile-time, so we cannot switch between two definitions based on a parameter (what the user configured) at runtime.
</blockquote></details>

That makes sense. Thanks for explaining.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-:-LB8EVJr06QnYpko0doR:b-prqlcu)*

jbakosi:
Okay thanks. Let's put in a test with LF for vortical flow only.

---

---



*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-:-LB8f9kM0SHVqfc9G3_1:bkuz50t)*

adityakpandare:
This is great! It looks very similar to the DGPDE.h. I have not completely understood it yet, but I'll understand it as I use it more. I'll push the LF regression test now.

---

Reviewed 3 of 4 files at r3.
Review status: 30 of 31 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/PDE/CompFlow/DGCompFlow.h, line 197 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-LBGvMW_-ea4n6ZPidqb:-LBGvMW_-ea4n6ZPidqc:bdm0wrf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/36d6a2e9ddbbfa0f82080d34b8549e99fe6b0f97/src/PDE/CompFlow/DGCompFlow.h#L197)):*
> ```objc
> 
>         auto flux =
>           m_riemann.flux( f, geoFace, {{U.extract(el), U.extract(er)}} );
> ```

This is awesome! :+1: It has become very clean now. Thanks for doing this.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/238#-:-LBGwVkzAr7pzMKDxsas:bxue2m7)*

PR #242, opened 2018-05-03
Switch reduction of mesh statistics from arrays to individual reals

jbakosi:
This is prompted by the following assert triggered very rarely by CI:
```
terminating with uncaught exception of type tk::Exception: Size of max(stat) must be 2
[18:01:18] :     [Step 3/3] >>> Exception at /tmp/TeamCity-4/work/821a9fd6f64749d9/src/Inciter/Transporter.C:567: void inciter::Transporter::maxstat(tk::real *, std::size_t)
```
`Transporter::maxstat` is a reduction target receiving contributions using a built-in reduction type, so it is puzzling why the size of the incoming array is incorrect.

This commit is an attempt to remedy this problem by:

1. switching to `std::vector` on the contribute side and use Charm++'s new `contribute` API that takes a `std::vector`, which then figures out its size, see Charm++ manual _4.6 Reductions on Chare Arrays_, and

2. switching to a list of reals on the receiving side in the reduction targets, which though eliminates the array size, it does not necessarily eliminate the error, which is still unknown at this point, see also the discussion at https://lists.cs.illinois.edu/lists/arc/charm/2018-05/msg00017.html.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/242)


adityakpandare:
Yeah I saw the email exchanges you had with Sam White about this. Thanks for noticing and fixing this!

---

Reviewed 1 of 4 files at r1.
Review status: 1 of 4 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---

*[src/Inciter/Discretization.C, line 276 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-LButaME-orKDEANpob2:-LButaME-orKDEANpob3:b7obmqy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/67d5cf191a8fffaa3c2574a1ce53a7c18fcdf59d/src/Inciter/Discretization.C#L276)):*
> ```
>   std::vector< tk::real > min{ MAX, MAX };
>   std::vector< tk::real > max{ MIN, MIN };
>   std::vector< tk::real > sum{ 0.0, 0.0, 0.0, 0.0 };
> ```

So we don't really know whether switching to ```std::vectors``` helps here, correct? In that case, I was wondering if making this change makes sense. Just curious. ;-)

---

*[src/Inciter/Transporter.C, line 536 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-LButvEu82lNtZw_5FO7:-LButvEu82lNtZw_5FO8:bd0txgq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/16166bb2753786581f31656e91b41dca62568fbb/src/Inciter/Transporter.C#L536)):*
> ```
> 
> void
> Transporter::minstat( tk::real* d, std::size_t n )
> ```

Didn't Sam say that ```void reductionTargetfn( int n, double d[n] )``` was also okay?

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-:-LButNeq3Y1DLBD98NIw:b-ths884)*

adityakpandare:
Reviewed 2 of 4 files at r1.
Review status: 2 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/242)*

adityakpandare:
No problem at all. Looks good to merge! 

---

Reviewed 2 of 2 files at r2.

---

*[src/Inciter/Discretization.C, line 276 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-LButaME-orKDEANpob2:-LBvzuLxCIGeaoVBYZtN:b-en1jof) ([raw file](https://github.com/quinoacomputing/quinoa/blob/67d5cf191a8fffaa3c2574a1ce53a7c18fcdf59d/src/Inciter/Discretization.C#L276)):*

You are right, we have no idea if this will fix it. But I noticed that Charm++ can now take a `std::vector` as an argument to `contribute`, so I switched to that, because: (1) this is safer than the previous C-style array (as passed to `contribute`), and (2) `std::vector` knows its size so whatever caused the problem before, hopefully goes away(?). (Or we also risk sweeping the problem under the rug. Either way, I like it that now Charm++ determines the number of the entries we are collecting. I also tried passing `std::array`, but there is no API for that at this point.)
</blockquote></details>

Okay, that makes sense. Thanks for explaining!

---

*[src/Inciter/Transporter.C, line 536 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-LButvEu82lNtZw_5FO7:-LBvzliU7kjj8IdrdwOR:beljne2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/16166bb2753786581f31656e91b41dca62568fbb/src/Inciter/Transporter.C#L536)):*

Yeah but I  prefer spelling simple values out individually whenever there are just a handful instead of using C-style arrays.
</blockquote></details>

That's true too. Makes code simpler.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/242#-:-LBw-6HY1QxuhqBkuGC2:b3gigy)*

Issue #233, opened 2018-04-20
Optimize inclusion of git sha in executables


jbakosi:
Currently, the git sha is queried by cmake and placed into a header which then gets included into all executables via `Main/QuinoaConfig.h`. Unfortunately, a new git commit (or stash) operation triggers rebuilding half the code, which is sub-optimal. A way to avoid this would be to remove it from the header and write a small free function that returns the sha when queried, which if lives in a `.C` file and its own shared object, avoids triggering rebuilding half the code, which would increase productivity, especially on laptops.

jbakosi:
Done, see #244.

PR #244, opened 2018-05-05
Move configuration variables imported from cmake into shared library

jbakosi:
This is a solution for #233. Configuration values imported from cmake are now separated into compile-time and run-time macros and variables, respectively. The former are imported via a `src/Main/QuinoaConfig.h`, while the latter are via `src/Main/QuinoaConfig.C`. We also build a shared library, `Config`/`quinoa_config`, from this latter, containing simple free functions that return strings at run-time providing the imported variables.

Since only the shared library contains the git commit sha1, the build system now only triggers the rebuild of this shared library and the relink of all executables if a git operation changes the commit sha1, which is a lot faster than previously, where `QunioaConfig.h` was included in many places and triggered rebuilds of more than half the compilation units.

@junghans, @rfbird, @adityakpandare: Can you please test this, let me know what you think, and how it can be done better?

Also, @junghans: Do you think ccache now would be more useful?

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/244)


adityakpandare:
I just tested this on my Macbook. This works great for me! Thanks for making this change!

---

Reviewed 18 of 18 files at r1.

---



*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/244#-:-LBw-JVr4FNLhPfQPu71:b-tkht4l)*

PR #248, opened 2018-05-30
Mark user-defined edges for initial mesh refinement

jbakosi:
Usage in input file within the `inciter ... end` block:
```
  amr

    initref
      1 2 3 4
    end

  end
```
The above config block tags edges `{1,2}` and `{3,4}` for initial mesh refinement. Basic error checking is done during parsing, which trips on odd size of the above list.

Note: merge to branch _dgref_ so that only the small diff, related to this capability, is visible.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/248)

### DIFF
--- a/src/Control/CommonGrammar.h
+++ b/src/Control/CommonGrammar.h
@@ -116,6 +116,7 @@ namespace grm {
     WRONGSIZE,          //!< Size of parameter vector incorrect
     HYDROTIMESCALES,    //!< Missing required hydrotimescales vector
     HYDROPRODUCTIONS,   //!< Missing required hydroproductions vector
+    INITREFODD,         //!< AMR initref vector size is odd (must be even)
     CHARMARG };         //!< Argument inteded for the Charm++ runtime system
 
   //! Associate parser errors to error messages
@@ -268,6 +269,10 @@ namespace grm {
       "Specification of a 'hydrotimescales' vector missing." },
     { MsgKey::HYDROPRODUCTIONS, "Error in the preceding line or block. "
       "Specification of a 'hydroproductions' vector missing." },
+    { MsgKey::INITREFODD, "Error in the preceding line or block. "
+      "The number of edge-nodes, marking edges as pairs of nodes, used for "
+      "explicit tagging of edges for initial mesh refineoment is odd (it must "
+      "be even)." },
     { MsgKey::CHARMARG, "Arguments starting with '+' are assumed to be inteded "
       "for the Charm++ runtime system. Did you forget to prefix the command "
       "line with charmrun? If this warning persists even after running with "
--- a/src/Control/Inciter/InputDeck/Grammar.h
+++ b/src/Control/Inciter/InputDeck/Grammar.h
@@ -292,6 +292,24 @@ namespace grm {
     }
   };
 
+  //! Rule used to trigger action
+  struct check_amr_errors : pegtl::success {};
+  //! \brief Set defaults and do error checking on the transport equation block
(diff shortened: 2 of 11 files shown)


rfbird:

---

*[src/Inciter/Partitioner.C, line 122 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/248#-LDrZz138SeywqvA9SkV:-LDrZz144aOLNG1rN3n0:b-ww6mgh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ba20ac592a5a916110cdf5bf2509daa89f1b7cf5/src/Inciter/Partitioner.C#L122)):*
> ```
>   std::reverse( begin(m_initref), end(m_initref) );
> 
>   if ( !g_inputdeck.get< tag::amr, tag::init >().empty() ||
> ```

Is the condition here a bit backwards? Doing a negation of an empty vs checking if it exists? I can't decide, testing on empty is pretty standard and idiomatic, but `(not x) || (not y)` seems like it may be better as a check in the affirmative 

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/248)*

jbakosi:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/Inciter/Partitioner.C, line 122 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/248#-LDrZz138SeywqvA9SkV:-LDrbSlXB7hkywc7uONk:b5f67xt) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ba20ac592a5a916110cdf5bf2509daa89f1b7cf5/src/Inciter/Partitioner.C#L122)):*

Is the condition here a bit backwards? Doing a negation of an empty vs checking if it exists? I can't decide, testing on empty is pretty standard and idiomatic, but `(not x) || (not y)` seems like it may be better as a check in the affirmative 
</blockquote></details>

`.empty()` is faster than `.size()`. I'm not sure what you mean by "checking on exists". They are both `std::vector`, so they definitely exist, because they are from the tuple that is filled during parsing, so they are always there, but they may have zero size, if the user did not configure amr or when the user did not specify edges explicitly, respectively.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/248)*

jbakosi:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[src/Inciter/Partitioner.C, line 122 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/248#-LDrZz138SeywqvA9SkV:-LDrcIoJ30FnPEW2M6Q7:bc4dc6h) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ba20ac592a5a916110cdf5bf2509daa89f1b7cf5/src/Inciter/Partitioner.C#L122)):*

`.empty()` is faster than `.size()`. I'm not sure what you mean by "checking on exists". They are both `std::vector`, so they definitely exist, because they are from the tuple that is filled during parsing, so they are always there, but they may have zero size, if the user did not configure amr or when the user did not specify edges explicitly, respectively.
</blockquote></details>

The thinking behind this is that the user can EITHER configure a list of multiple refinements (consisting of unfirom or ic-based steps, following each other arbitrarily) OR explicitly specify a list of edges OR BOTH. So either of these three should rout to AMR, i.e., all these three options (list, edges, or both) can be used.

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/248)*

PR #249, opened 2018-06-11
Read node sets directly as well

jbakosi:
Recently we have switched to generating node lists associated to side sets based on faces and face connectivities. This is still the right thing to do, but we have partially-wrongly-saved large meshes that we still would like to use with CG and for those this commit puts in code to read the side set node lists directly from the ExodusII file if face connectivity (and triangle elements) are not available from the file. This is how we used to do this but this approach is only sufficient for node-based discretizations, e.g., CG.

Also put in a regression test exercising diagcg with a mesh that does not have triangles.

Note that for some reason matcg trips on the boundary conditions not being correct with this, which needs further investigation. (If the mesh is correctly saved, i.e., both triangles and tets, the matcg regression tests pass, but for some reason, only in parallel matcg fails with the mesh added, unitcube_60k.exo.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/249)


adityakpandare:
:lgtm:

---

Reviewed 17 of 17 files at r1.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/249#-:-LEklkSbEU1yZgiUhFTg:bnfp4nl)*

PR #250, opened 2018-06-13
Make meshconv generate triangle block if only tet exist

jbakosi:
This enables `meshconv`, using exo as both input and output, to detect if a triangle element block does not exist and generate one. It also transfer the side sets from the input exodus file.

There is a bunch of related changes:

1. Remove an inciter regression test that was put in to use a mesh that does not have triangle element block. This is now required. For DG, it is not okay to have no triangle element block or node side sets. For CG, it is okay to not configure a side set for a problem. (Some inciter regression tests use meshes without side sets, so the absence of side sets and/or triangle block in the input file is not made an error, only for DG.)

2. Modify a bunch of `meshconv` regression tests. These were necessary as `meshconv` now adds a triangle element block if it does not find one and the output file diffs if there is a new element block.

3. Changed the order of triangle and tet element blocks as `ExodusIIMeshWriter` saves them. This was needed to transfer the side set tet ids from the input and output mesh when adding a triangle element
block. As a result, more meshconv regression tests had to updated. Visually verified the new gold standard meshes to be correct. 

4. Fixed a couple of bugs in ExodusIIMeshReader:

  * `m_nel` was not cleared between potentially successive calls to `readElemBlockIDs()`

  * `readFaces()` was attempting to read as many triangle elements as many boundary faces are found assigned to side sets. These are not necessarily always the same. The correct number of triangle elements to read is the max number of triangle elements in the file.

5. `ExodusIIMeshWriter` now write side sets, but only if a `tk::UnsMesh` object is passed to it. This is how meshconv uses the mesh writer, but not how inciter uses it. If side sets are needed that takes a side sets themselves, a new API can be added. This is consistent with how inciter reads and writes meshes and field data into exodus files: it writes different components independently as (1) that scales better in parallel and (2) it is more convenient to read/write components when they are needed/ready instead of creating a mesh object with everything inside.

6. Remove the piece of code that reads the node lists associated to side sets if no triangle element block is found in the input file: this is in two places: `Partitioner::createWorkers()` and `Transoprter::createPartitioner()`. For `DiagCG/MatCG` if BCs are desired, sidesets are required, which are obtained from triangle element connectivity and side sets instead of attempting to read the node lists. DG always requires side sets as well as triangle element connectivity. 

7. Enabled some commented out exception catches, that were erroneously left commented some time 
 ago.

8. Since `readSidesetFaces()` slightly changes, some unit tests also had to be updated.


adityakpandare:
Sorry for the really late review! Thanks for putting this in. I have a few questions about the ```ExodusIIMeshReader/Writer``` part.

---

Reviewed 36 of 36 files at r1.
Review status: all files reviewed, 6 unresolved discussions (waiting on @jbakosi)

---

*[src/Inciter/DG.C, line 230 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI0c1VEeSxPM--NI3v:-LFI0c1W1GV-J14qZKUh:b-tvot4b) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/Inciter/DG.C#L230)):*
> ```
>   }
> 
> std::cout << "adj: " << s[0] << ", " << s[1] << ", " << s[2] << '\n';
> ```

Should we be having this here?

---

*[src/IO/ExodusIIMeshReader.C, line 386 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI5GsyC6boZWsuqSci:-LFI5GsyC6boZWsuqScj:bafool7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L386)):*
> ```
> // *****************************************************************************
> {
>   auto nbfac = readSidesetFaces( mesh.sidetet(), mesh.faceid() );
> ```

I might've misunderstood this, but doesn't ```Transporter``` already call ```readSidesetFaces()```? And when the ```ExodusIIMeshReader``` object ```er``` is created in ```Transporter```, wouldn't this be called too? So we'd be double-calling ```readSidesetFace()```?

---

*[src/IO/ExodusIIMeshReader.C, line 412 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI1wp8AYSJc9ZJb73X:-LFI1wp8AYSJc9ZJb73Y:b-s021u) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L412)):*
> ```
>   // Read triangle boundary-face connectivity (all the triangle element block)
>   std::vector< std::size_t > l_triinpoel;
>   readElements( {{0,nelem(tk::ExoElemType::TRI)-1}}, tk::ExoElemType::TRI,
> ```

Why was this change from ```nbfac-1``` needed?

---

*[src/IO/ExodusIIMeshReader.C, line 571 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI7lvW4yDNSBhEGKHr:-LFI7lvW4yDNSBhEGKHs:b-qmrih4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L571)):*
> ```
>       // Read in face and element list for side set i
>       ErrChk( ex_get_set( m_inFile, EX_SIDE_SET, i, tbelem.data(),
>                           faces.data() ) == 0,
> ```

Could you explain what exactly ```faces``` stores? I'm slightly confused. I thought ```tbelem``` contains the boundary face list. The 4th argument is something called "side_set_side_list" according to the exo2 API doc: https://gsjaardema.github.io/seacas/html/exodusII_8h.html#ac96dfd540bec6ecfdfafaa90fdc25276.

---

*[src/IO/ExodusIIMeshWriter.C, line 242 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI9EH8E1-6OuyeI0wB:-LFI9EH8E1-6OuyeI0wC:b-nobksd) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshWriter.C#L242)):*
> ```
>     // tet (1...4) indicating which face is aligned with the side set.
>     ErrChk( ex_put_set( m_outFile, EX_SIDE_SET, s.first, sidetet.data(),
>                         tk::cref_find( mesh.faceid(), s.first ).data() ) == 0,
> ```

So basically the ```faces``` in ```ExodusIIMeshReader::readSidesetFaces()``` is only used here, correct?

---

*[src/IO/MeshFactory.C, line 169 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI9YQ6DaLQJhP-PVyv:-LFI9YQ6DaLQJhP-PVyw:b-icj0jx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/MeshFactory.C#L169)):*
> ```
>           triinpoel.push_back( inpoel[ mark+tk::lpofa[f][0] ] );
>           triinpoel.push_back( inpoel[ mark+tk::lpofa[f][1] ] );
>           triinpoel.push_back( inpoel[ mark+tk::lpofa[f][2] ] );
> ```

I'm not entirely sure, but we might have to reverse the order of the second index of ```lpofa```

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-:-LFI9gaW1SFU2uDaPOnw:b-cgxwkc)*

adityakpandare:
Looks good!

---

Reviewed 1 of 1 files at r2.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---

*[src/IO/ExodusIIMeshReader.C, line 386 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI5GsyC6boZWsuqSci:-LFIUyuo-eqAo79fDXGM:b-6lw8j3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L386)):*

No. `readAllSideSets()` is only called _within_ `ExodusIIMeshReader`. `Transporter` only calls `readSidesetFaces()`, i.e., without a `UnsMesh` object being created first. This is how we do it in `inciter`. `readAllSidesets()` is called by `Meshconv`.
</blockquote></details>

That makes sense. Thanks for explaining!

---

*[src/IO/ExodusIIMeshReader.C, line 412 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI1wp8AYSJc9ZJb73X:-LFIRd5vE4z1FAX89tjI:b-9l4hrb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L412)):*

Because the number of boundary faces (returned by `readSidesetFaces()` is not necessarily the same as the number of faces in the triangle element block in the Exodus file. This can happen when multiple side sets refer to the same faces, which all of the Hexpress-generated meshes do: they have side set 1 collecting all faces of the boundary then also specifying side sets 2-7 assigning faces of a cube. For those meshes `nbfac` is larger than the number of faces in the triangle element block. This is more general.
</blockquote></details>

I get it now. Thanks for the explanation.

---

*[src/IO/ExodusIIMeshReader.C, line 571 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI7lvW4yDNSBhEGKHr:-LFISJha0RVPBpEyPSF2:bs9gl0b) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshReader.C#L571)):*

Yeah. `faces` here stores the sides of faces used to define which face of a tetrahedron is aligned with a triangle element (this is a relative index between 1...4) and is associated to a side set id. This is also documented for `UnsMesh::m_faceid` to which `MeshConv` reads this into. This is called _"side set side list"_ by Exodus.

The other data structure Exodus assigns to a side set is the list of element (tet) ids adjacent to a side set. This is _"side set element list"_ in Exodus. I also documented this for `UnsMesh::m_sidetet`.
</blockquote></details>

Nice! Seems like ```m_faceid``` could be useful. Thanks!

---

*[src/IO/ExodusIIMeshWriter.C, line 242 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI9EH8E1-6OuyeI0wB:-LFITA6g4WeDE3kO-n8h:b7x9ihp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/ExodusIIMeshWriter.C#L242)):*

That's correct. We don't need it in `inciter` (so far anyway), because we only work with the tets adjacent to the side sets and generate the face list and their connectivity (in `bface` and `triiinpoel`) ourselves.

For writing side sets (as meshconv needs to put this into the output mesh), this is needed to write both the _side set element list_ and _side set side list_ associated to a side set in the output mesh file. This latter is why I switched the order of writing the tetrahedra element block followed by writing the triangle element block in `ExodusIIMeshWriter::writeElements()`: this way `Meshconv` can just copy over both the _side set element list_ (`UnsMesh::sidetet()`) as well as the _side set side list_ (`UnsMesh::faceid()`) data read from the input file. With the opposite order, the automatically-generated element map would have to be changed when we add the triangle element block to a file that does not have this yet. Note that `ExodusIIMeshWriter::writeSidesets()` does not write an element map. It could, but by writing the tetrahedron element block first, the element mapping does not change in the file and so the tet ids and their relative face ids (1..4) associated to a side set can remain the same without remapping.
</blockquote></details>

Okay. Thanks!

---

*[src/IO/MeshFactory.C, line 169 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-LFI9YQ6DaLQJhP-PVyv:-LFIRIq5Cf0VxvJb6Pc1:brxp9sd) ([raw file](https://github.com/quinoacomputing/quinoa/blob/61512c7a92aa1e650a96192a100dd6fadc1401a9/src/IO/MeshFactory.C#L169)):*

Yeah, we will see how this works out. This could also be our tool to in the future to "fix" meshes suitable for inciter.
</blockquote></details>

Sounds good!

---


*Comments from [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/250#-:-LFIZxW85mVHk2HGOOuO:b-vs1prh)*

PR #253, opened 2018-07-06
Interface with Omega_h mesh reader for inciter

jbakosi:
This adds the capability to read [Omega_h](https://github.com/ibaned/omega_h)'s `osh` meshes in inciter. Omega_h is a new TPL and is off by default. Can read meshes in serial and parallel (regression-tested) but no side sets, thus no boundary conditions, yet.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/253)


jbakosi:
FYI: @ibaned. Thanks for all the help and for developing Omega_h.

Issue #35, opened 2015-07-02
Explore build without MPI


jbakosi:

jbakosi:
There are two aspects of this:

1. Using MPI third-party libraries possibly without MPI (e.g., OpenMPI, etc.) but with Charm++'s AMPI. Not sure if this is possible but would be a good. See https://lists.cs.illinois.edu/lists/arc/charm/2016-03/msg00049.html.

2. Not using the MPI wrappers for the build, but let cmake pick up whatever it finds as default system-wide. This already works fine, see d4add0a.

PR #255, opened 2018-07-18
Initial steps towards replacing MPI with AMPI

jbakosi:
## Background
This goes back to #35. In a nutshell, the idea is to build Charm++ as well as AMPI on a native Charm++ communication layer instead of MPI, then build all of our MPI TPLs on top of AMPI. Benefits are
- use Charm++'s native communication layer (with likely better performance) instead of MPI,
- no need for MPI (as a dependency), and, perhaps most importantly,
- use advanced features of AMPI, e.g., load balancing and thread migration, _even_ within an MPI-only TPL function call, e.g., hypre.

## Changes
This PR, in conjunction with https://github.com/quinoacomputing/cmake-modules/pull/22 and https://github.com/quinoacomputing/quinoa-tpl/pull/28, augments our build system to build Charm++'s AMPI using a non-MPI communication layer and successfully builds all of our MPI TPLs on top of AMPI.

All of this is optional and can be triggered by passing `the -DENABLE_AMPI=true` to our TPL cmake command.

## Status
This is tested only with the gnu compilers and on a single linux workstation so far, but using various Charm++ non-MPI communication layers. For a detailed status see https://github.com/quinoacomputing/quinoa-tpl/blob/ampi/CMakeLists.txt#L122.

## Future work
Overall, this partial progress, because Charm++/MPI interoperation using AMPI is not a completed feature yet in the way described above. See also @stwhite91's insights, and an experimental (but immediate) way forward, at https://lists.cs.illinois.edu/lists/arc/charm/2018-07/msg00011.html.

Thus the purpose of this PR(s) is to get ready for when this feature is completed in Charm++ and/or experiment with it (and not to break the AMPI build) in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/255)


cmake/trilinos_ampi.patch
```diff
@@ -0,0 +1,74 @@
...
+diff --git a/packages/zoltan/src/coloring/color_test.c b/packages/zoltan/src/coloring/color_test.c
+index 465a6a97ed..47fca6fa94 100644
+--- a/packages/zoltan/src/coloring/color_test.c
++++ b/packages/zoltan/src/coloring/color_test.c
+@@ -59,7 +59,7 @@ extern "C" {
+ #include "g2l_hash.h"
+ #include "params_const.h"
+ #include "zz_util_const.h"
+-#include "graph.h"
++#include "../graph/graph.h"
```
junghans: Should you just drop that part of the patch by setting an include path?

jbakosi: It's not that simple. This collides with a Charm++ include file that is named the same way. Both include directories should be on the list of the include paths and I think the order in which they appear will decide what gets included, but the question is for what compilation unit? In this case, both are in a TPL that is out of our control. I figured, this is the safest solution, but they have not merged this PR yet: https://github.com/trilinos/Trilinos/pull/3122

stwhite91: We can also rename the graph.h in Charm++. That's too generic a name for us to be putting in our include/. I opened a redmine issue for this here: https://charm.cs.illinois.edu/redmine/issues/1948

jbakosi: Thanks Sam, that would be great.

PR #259, opened 2018-08-10
Migration

jbakosi:
This PR enables the 3 _inciter_ regression tests with migration that @adityakpandare has set up previously. The code uses Charm++'s `AtSync()`  with `RandCentLB`, see the section on load balancing of the [Charm++ manual](http://charm.cs.illinois.edu/manuals/html/charm++/7.html) for more detail, which randomly assigns PEs to chares that call `AtSync()`.

Compared to the original way how the 3 regression tests were setup, they are all modified to
 * take less time steps,
 * user higher virtualization, and, of course,
 * now they use `AtSync()` migration instead of _periodic load balancing_ (that was triggered previously by the `+LBPeriod` Charm++ command line argument).

This is so
 * the tests take less time to run and evaluate (cmake grep can be slow on a large-screen-output regression test), and
 * to more heavily stress-test migration (due to larger virtualization).

This PR also puts in a capability to collect chare states during a run. It is intended for debugging and off by default. See 710a46fa for more details on how to enable and use this feature.

At this point `DG` and `DiagCG` can do migration, `MatCG` is not yet ready.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/259)


jbakosi:
Hey @rfbird! I believe this is now _as good as it gets_ and according to my current best knowledge of how migration is supposed to be used with bound chare arrays and `AtSync()`. So please take a look when you get a chance and let's merge this in.

Actually, the merge will probably conflict, because `Refiner` is not `pup`able yet, nevertheless, this could now use a code review and I'm ready to merge it in.

As it is now, running the regression tests that exercise migration in an infinite loop ran for hours for me, and when it broke, it broke with a problem coming out of _pthreads_, and not due to quiescence or a problem on our side. So, fingers crossed, this is _as good as it gets_ for now.

rfbird:
OK, awesome. More to come as I see the rest of the files

jbakosi:
Hey @rfbird! See https://github.com/quinoacomputing/quinoa/pull/259/commits/059312850a8111c68df4cf71f0aa4d84fb8a5067. Thanks for forcing me to do it! Once this passes, I think this is ready for a merge.

PR #266, opened 2018-10-03
Amr enhancements

rfbird:
Replacing algorithm with possible faster implementation.

Profiling to follow..


rfbird:
Marked the comments as done now, just to signal that I'm happy for this to be merged now (as per our meeting).

If the future I still have some cleaning up to do, but this is much better than it was

Issue #261, opened 2018-08-31
SMP


jbakosi:
Notes towards enabling quinoa build and run on Charm++ SMP build:

SMP todo
-------------
- [x] Try `SMP`-build of Charm++ and use tasks_per_node = nodes_per_node-1 for communication and -2 for OS noise.
- [x] Keep mesh reader and partitioner in nodegroup, move reshuffling to array (tried different ways: will not do it: not clear it would be beneficial vs added code, performance seems already good on 36-core (2socket) workstation `+p36` vs `+ppn 18+pemap 1-17,19-35 +commap 0,18`, the latter is 2x faster)
- [x] Fix `unittest`'s MPI suite in SMP mode: Revamped unittest to use nodegroups instead of invoking the MPI suite via Charm++-MPI interopeation and this almost works. Still need to
  - [x] Fix the MPI test _"ErrChkMPI macro doesn't throw all true"_, which currently fails even with `+ppn 1`.
  - [x] Port those unit tests that formerly were not part of the MPI suite but now with `+ppn N>1` invoke MPI calls because they want to run parallel. These are tests that exercise MPI-only libraries, such as  ExodusII and Omega_H. _Solution:_ call these tests from the _MPIRunner_ nodegroup and only call the test from a single rank. (This should yield the previous, serial behavior.)
- [x] Move outfile writer to nodegroup in inciter. (This is not strictly necessary for SMP to work. For performance reasons, however, it might be worth it. I defer this for now for some other time and after a profiling to see how much it is needed.) _Update: Done, using a group instead of a nodegroup, exercising asynchronous writes, see dbcb0ac2, bb91ba13, and 2e783569._
- [x] Fix `MatCG` with `+ppn N>1` (`MatCG` now works with `Solver` as a nodegroup in non-SMP mode, next is to try it in SMP mode and witn `ppn N>1`). Did work towards it, does not work and hard to debug likely due to the different parallel semantics of nodegroups compared to groups, won't fix, see https://github.com/quinoacomputing/quinoa/pull/279/commits/86db6418e0d004e7f775e5d492bb9a3c104853e6.
- [x] Add `+ppn N>1` regression tests. WIP: augmenting the build system (cmake/ctest) to run the full regression test suite exercising SMP mode for parallel tests. This is done and all tests pass in non-SMP mode with f60a9f69.
- [x] Add more `npe != ppn` regression tests for SMP mode.
- [x] Augment CI with SMP mode.

Performance todo
-------------
[...]

PR #279, opened 2018-11-27
SMP

jbakosi:
This is work towards being able to run with Charm++ built in SMP mode. See Sec. [1.4 Machine Model](http://charm.cs.illinois.edu/manuals/html/charm++/1.html#SECTION00140000000000000000) of the [Charm++ documentation](http://charm.cs.illinois.edu/manuals/html/charm++/manual.html).

So far we have only used Charm++ in non-SMP mode and the commits in this PR port the code to SMP mode. At this point, the port appears to be successful in that the code builds and all tests pass both in non-SMP as well as in SMP mode. The next steps are to exercise SMP mode by adding regression tests with `+ppn N>1`, and add CI build matrix entries with Charm++ built in SMP mode so that both the currently used non-SMP mode as well as SMP mode can be automatically tested. For more details, see #261.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/279)


jbakosi:
### Closing notes

* Augmented all CI with SMP builds for all compilers.

* cmake now detects and outputs if we Charm++ is built in SMP mode and if Charm++ is built with randomized queues:
```
-- Charm++ built with randomized message queues
-- Charm++ built in non-SMP mode
```

* Another way to tell if an executable is running in SMP mode (besides Charm++'s info) _and_ using threads is the new screen output on the PE layout. The screen output below shows that Charm++ is running in SMP mode and that we are running 2 logical (compute) nodes (i.e., two MPI ranks), and each spawns 17 worker threads + a thread designated for communication only, thus the total number of PEs used is 36.

```
./charmrun +p 34 --bind-to none Main/unittest -v -q +ppn 17

Running on 2 processors:  --bind-to none Main/unittest -v -q +ppn 17
charmrun>  /usr/bin/setarch x86_64 -R  mpirun -np 2  --bind-to none Main/unittest -v -q +ppn 17
Charm++> Running on MPI version: 3.1
Charm++> level of thread support used: MPI_THREAD_FUNNELED (desired: MPI_THREAD_FUNNELED)
Charm++> Running in SMP mode: 2 processes, 17 worker threads (PEs) + 1 comm threads per process, 34 PEs total
Charm++> The comm. thread both sends and receives messages
Converse/Charm++ Commit ID:
CharmLB> Load balancer assumes all CPUs are same.
Charm++> Running on 1 hosts (2 sockets x 18 cores x 1 PUs = 36-way SMP)
Charm++> cpu topology info is gathered in 0.090 seconds.
...
                                 < ENVIRONMENT >
                                 ------ o ------                                            
...                                               
 * Run-time environment:                       
 -----------------------                       
...
     Number of processing elements            : 34 (2x17)
...
```

* Adding a regression tests now take an optional argument that specifies Charm++'s `+ppn` argument.

```
add_regression_test( <test_name> <executable>
￼ #                      [NUMPES n]
￼ #                      [INPUTFILES file1 file2 ...]
￼ #                      [ARGS arg1 arg2 ...]
￼ #                      [LABELS label1 label2 ...]
```

* There is some discussion on how SMP mode expects the number of cores/threads with logical nodes in `cmake/add_regression_test.cmake`.

* There are more regression tests in SMP mode, specifically exercising SMP mode. Thus working with an SMP build is more thorough and it takes longer to run the regression tests. (It is much faster, ~2X, to run the unit tests.)

* Removed `script/run_tests.sh`. It became wasteful to maintain it, hard and confusing to use. Here is how I run the unit-. and regression tests  routinely on my 36-core workstation (line 1: non-SMP, line 2: SMP)
```
./charmrun +p 36 Main/unittest -v -q && ctest -j 36 --output-on-failure -LE extreme
./charmrun +p 34 --bind-to none Main/unittest -v -q +ppn 17 && ctest -j 36 --output-on-failure -LE extreme
```

* The internals of running exodiff in SMP mode is more involved because rank ids may not be the same for a baseline saved with non-SMP mode and there is new cmake code to handle that. See details in `cmake/test_runner.cmake`.

* Note that if you use `mpirun` as your runner, the `-n` argument means _run this many copies of my program_, and not _use this many CPUs_. Using `charmrun` with `+p` still specifies the number of PEs to use in both non-SMP and SMP mode, but in SMP mode the `+p` arg must be a multiple of `+ppn`.

* `MatCG` is purged. R.I.P. It was a good learning experience and we learned how _not_ to do things.

* Parallel mesh read is now chunk per compute node.

* Large field data output is now asynchronous and the writes are serialized per compute node. This should also help performance.

* Inciter's schemes' asynchronous logic (within and before) time step has changed and some functions have been renamed.

* There is a new algorithm for distributing the mesh and boundary condition information after mesh partitioning, which also performs better. We tested this in SMP mode at O(10K) compute cores and the improvement is huge. Big thanks to @nncarlson for the idea and the for discussion with that algorithm.

* The unittest executable got a complete overhaul: it now runs fully in Charm++ land not with Charm++/MPI interopration. This new way no longer requires explicit MPI code and runs the all unit tests within a single harness.

PR #287, opened 2019-01-28
added capability to send intermediate nodes at char boundaries -- nee…

rfbird:
I think this fixes the parallel bug we've been seeing, with the changes being pretty much exclusively in `Refiner.C`. 

We discussed most of the details on the phone, but the key changes are:

1) We had that rogue `if` that ignored locks
2) We had that `continue` instead of `break`in the AMR library which meant in a corner case the wrong face could be 1:4'd
3) Sending nodes which are "intermediates" across the networking, and locking a lot more. This definitely needs some thought into given into it, perhaps something to talk about next meeting, but from my quick testing seems to be giving valid results at least

Can you look this over and possibly see if it's working too?

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/287)


jbakosi:
Ah! Outstanding! I will take a look at this shortly! Thanks for tracking this down!

PR #288, opened 2019-02-11
Communicate limited initial solution, and globally sync this initialization

adityakpandare:
Initial solution limiting has to come with a few synchronization points (global-reductions) where the limited solution has to be communicated to neighbor-chares before time-stepping starts. These communications and global-reduction have been added in this set of commits. Updated the baselines accordingly. This should help pass the problematic cyl_advect_dgp1_u0.90_pe4 test case. Let's keep an eye on Azure to make sure this actually does pass. Thanks @jbakosi for your help in debugging and fixing this issue.


jbakosi:
Hey guys. I have updated the [CLA](http://www.clahub.com/agreements/quinoacomputing/quinoa), which is new mostly due to LANL's management transition but the wording is also slightly different. Please read and sign. Thx.

PR #289, opened 2019-02-12
Fixes based on cppcheck

jbakosi:
I have fixed a bunch of code based on cppcheck's suggestions. This PR touches a lot of code but the mods are very small and local.

I also ignore and suppress some of cppcheck's suggestions. Most of these ignore the suggestion to use an STL algorithm instead of a range-based for loop. While I think this is a good idea in general (and I did convert quite a few to algorithms), it can lead to more verbose, and in IMO, less readable code, mostly because C++11  does not yet support generic lambdas (whose arguments can be `auto`), but range-based for loops do. (Generic lambdas come in C++14 for which we are not quite ready yet.)

Also updated relevant cmake code to generate two, instead of one, cppcheck targets: `cppcheck` and `cppcheck-xml`. Previously, we only had the latter, but with a name of the former. The rationale of having two targets is that the first one outputs its findings to the screen (and so it is easy and quicker to work with) and we can capture the non-zero exit code from the internal cmake command in the shell correctly, which will be good for tripping CI on cppcheck's findings. On the other hand, the latter, `cppcheck-xml`, will be good for generating the nice(r) html output at, e.g., https://quinoacomputing.github.io/Debug/cppcheck/index.html. However, since this latter target runs several shell commands within the same cmake target, the exit code from cppcheck (the first command) never gets to the shell, so cppcheck's findings don't generate a non-zero exit code, which then cannot be used for tripping CI. (Eventually, to trip CI based on cppcheck, we probably have to run both cppcheck targets, one which can produce an error code to the shell, and one that can generate a html output. I don't really like this, but I see no better solution for this. Cmake gurus welcome to suggest something better.)

So these code changes are based on cppcheck's current findings, but I have not touched the AMR directory, for which the fixes don't appear too bad at all. We will not enable tripping CI based on cppcheck just yet.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/289)


jbakosi:
Also, here is an interesting discussion on `std::for_each` vs. range-for: 

https://www.reddit.com/r/cpp/comments/appu15/why_you_should_use_stdfor_each_over_rangebased/?st=JS2N3RPH&sh=d656509a

Interesting that it starts with advising to prefer the former followed by all/most responses that disagree, including [a performance test](https://www.reddit.com/r/cpp/comments/appu15/why_you_should_use_stdfor_each_over_rangebased/egc6dvk/), showing substantial favor towards the range-for wrt. memory usage as well as performance using two compilers.

src/Control/Walker/CmdLine/CmdLine.h
```diff
@@ -83,6 +83,7 @@ class CmdLine : public tk::Control<
     //!   contains a copy of the command-line stack, the command-line stack must
     //!   be possible to be instantiated without passing the ctrinfo map,
     //!   otherwise it would be a mutual dependency.
+    // cppcheck-suppress noExplicitConstructor
```
rfbird: I wonder if we can have some standard comment that the parsers still recognize that quickly tells human it's an annotation? 

`//@@ cppcheck-suppress noExplicitConstructor`

After thinking a bit more, I guess given a lot of the compiler ones are pragmas it's harder

src/DiffEq/MixDirichletCoeffPolicy.h
```diff
@@ -139,13 +139,13 @@ class MixDirichletHomCoeffConst {
...
       for (ncomp_t c=0; c<ncomp; ++c) sumY += Y[c];
 
-      // Y|Kc
-      std::vector< tk::real > YK( ncomp, 0.0 );
-      for (ncomp_t c=0; c<ncomp; ++c) {
-        YK[c] = sumY - lookup( mean(depvar,c), moments );
-        //std::cout << "YK: " << YK[c] << ' ';
-      }
-      //std::cout << std::endl;
+//      // Y|Kc
```
rfbird: Do we want to delete this? Is this something you're still working on (I may have seen that chain elsewhere?)

src/DiffEq/MixMassFractionBetaCoeffPolicy.h
```diff
@@ -186,7 +186,7 @@ class MixMassFracBetaCoeffDecay {
         tk::real v = tk::ctr::lookup( tk::ctr::variance(depvar,c), moments );
 
         if (m<1.0e-8 || m>1.0-1.0e-8) m = 0.5;
-        if (v<1.0e-8 && v>1.0-1.0e-8) v = 0.5;
+        if (v<1.0e-8 || v>1.0-1.0e-8) v = 0.5;
```
rfbird: This seems like a semantic change? Is this a bug fix snook in?

Also given the numbers of a `-` in, i.e. `1e-8` do you think it's better to leverage spacing here:

` if (v<1.0e-8 || v>1.0-1.0e-8) v = 0.5;`

vs

` if ( v < 1.0e-8 || v > 1.0 - 1.0e-8 ) v = 0.5;`

Finally, as ever, I love braces {} in if's

rfbird: `if ( v<1.0e-8 || v>1.0 - 1.0e-8 ) v = 0.5;` 

may even be better middle ground

src/IO/ExodusIIMeshReader.C
```diff
@@ -620,7 +619,7 @@ ExodusIIMeshReader::readSidesetNodes()
       tk::unique( nodes );
       // Store 0-based node ID list as std::size_t vector instead of ints
       auto& list = side[ i ];
-      for (auto&& n : nodes) list.push_back( static_cast<std::size_t>(n-1) );
+      for (auto n : nodes) list.push_back( static_cast<std::size_t>(n-1) );
```
rfbird: Very happy to see `&&` die!

src/IO/ExodusIIMeshReader.h
```diff
@@ -142,9 +144,12 @@ class ExodusIIMeshReader {
     }
 
     //! Copy constructor: in terms of copy assignment
+    // cppcheck-suppress uninitMemberVar
     ExodusIIMeshReader( const ExodusIIMeshReader& x ) { operator=(x); }
 
     //! Move assignment
+    // cppcheck-suppress operatorEqMissingReturnStatement
+    // cppcheck-suppress operatorEqVarError
```
rfbird: It would be nice if there was a way to combine these into one line

src/Inciter/ALECG.C
```diff
@@ -310,8 +311,9 @@ ALECG::rhs()
     comrhs_complete();
   else // send contributions of rhs to chare-boundary nodes to fellow chares
     for (const auto& n : d->Msum()) {
-      std::vector< std::vector< tk::real > > r;
-      for (auto i : n.second) r.push_back( m_rhs[ tk::cref_find(d->Lid(),i) ] );
+      std::vector< std::vector< tk::real > > r( n.second.size() );
+      std::size_t j = 0;
```
rfbird: again we see a very similar code pattern here

src/PDE/Integrate/Volume.C
```diff
@@ -321,13 +321,12 @@ tk::volIntP2( ncomp_t system,
...
     {
-      // Continue to calculate the derivatives of the basis functions dB/dx for DG(P2)
+      // Calculate derivatives of the basis functions dB/dx for DG(P2)
       auto db5dxi1 = 12.0 * coordgp[0][igp] + 6.0 * coordgp[1][igp]
                    +  6.0 * coordgp[2][igp] - 6.0;
       auto db5dxi2 =  6.0 * coordgp[0][igp] + 2.0 * coordgp[1][igp]
                    +  2.0 * coordgp[2][igp] - 2.0;
-      auto db5dxi3 =  6.0 * coordgp[0][igp] + 2.0 * coordgp[1][igp]
-                   +  2.0 * coordgp[2][igp] - 2.0;
+      auto db5dxi3 = db5dxi2;   // = 6.0 * xi1 + 2.0 * xi2 + 2.0 * xi3 - 2.0
```
rfbird: I have such mixed feelings about this, do we think it's the right way to handle this?

On one side, explicitly donating the third term is equal to the second, and still referencing the third term later seems sensible...

On the other, a line that just is `terse_variable_a = terse_variable_b` (with some actual very terse variable) looks like a problem waiting to happen

PR #293, opened 2019-02-15
Dgp2 vol

WeizhaoLi2018:
Integrate srcInt, volInt and bndInt into a compact function form
Add more functions for volume integration in Basis.C

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/293)


jbakosi:
Since you commented out `coord` and `inpoel`, clang errors out as they are documented but does not find them among the function arguments: 
```
/home/quinoa/src/PDE/Transport/DGTransport.h:227:20: error: parameter 'inpoel' not found in the function declaration [-Werror,-Wdocumentation]
    //! \param[in] inpoel Element-node connectivity
                   ^~~~~~
/home/quinoa/src/PDE/Transport/DGTransport.h:228:20: error: parameter 'coord' not found in the function declaration [-Werror,-Wdocumentation]
    //! \param[in] coord Array of nodal coordinates
                   ^~~~~
```
See `dg::Transport::dt()` to see how you can avoid this temporarily.

Hey @adityakpandare : Will we ever need `inpoel` and `coord` for `fieldOutput` in DG? Currently, neither transport or compflow use it but rely on `geoElem` instead. If we will not need it, let's remove them both from `fieldOutput`'s argument list (compflow, transport, and DGPDE).

jbakosi:
@WeizhaoLi2018 you don't have to wait for travis, it will probably fail anyway as some builds run out of the allotted time. (We are considering turning off travis soon.)

PR #296, opened 2019-03-13
Parallel AMR

jbakosi:
It is time to review this once again and merge as it is growing pretty big already. Both DG and DiagCG appear to work with during-time-stepping mesh refinement in serial and parallel, with and without overdecomposition, with and without migration. Also added a bunch of regression tests.

I will also take a look at this diff and might push some changes if I find something. Thanks in advance for taking a look.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/296)


jbakosi:
Adding @adityakpandare as a reviewer, because a lot of this touches code that he is very familiar with and since this is just one of those monster PRs again. Please feel free to selectively ignore and only comment on what makes reasonable sense to review. Sorry guys for the monster.

And please @rfbird , sign the CLA. Thanks.

PR #334, opened 2019-07-03
Enable quiescence detection for migration tests

jbakosi:
This is possible if `+LBPeriod` is small. Tested it locally in infinite loop -- appears to work fine, so attempting to enable it.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/334)

### DIFF
--- a/cmake/add_regression_test.cmake
+++ b/cmake/add_regression_test.cmake
@@ -263,11 +263,8 @@ function(ADD_REGRESSION_TEST test_name executable)
   # Construct and echo configuration for test being added
   set(msg "Add regression test ${test_name} for ${executable}")
 
-  # Run all non-migrating regression tests with quiescence detection
-  list(FIND TEST_LABELS migration i)
-  if (${i} EQUAL -1)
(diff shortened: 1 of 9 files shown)


jbakosi:
Additional background info from @epmikida:

> Do keep in mind that QD working with LB is also dependent on LBPeriod. I found out during my digging that the reason QD sometimes incorrectly triggers with LB due to how periodic load balancing mixes with AtSync load balancing in the current LB framework. If all AtSync chares reach the AtSync point *before* the LBPeriod (which can be specified on the command line in seconds), then the load balancing framework will not do anything until the LBPeriod mechanism triggers. This mechanism is timer based, and not visible to QD. Which is why at this point, QD will trigger. If, however, the LBPeriod is reached before the chares hit the AtSync then QD will work just fine. You may want to set a small LBPeriod with the `+LBPeriod` command line argument. The new LB framework does not use a timer to make sure all PEs participate in AtSync, which is why QD is better supported in the new framework.

PR #342, opened 2019-08-14
P2 adaptive

WeizhaoLi2018:
This PR includes adding DGP2 to p-adaptive DG scheme and also with some related regression testcases update

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/342)


jbakosi:
Hi @WeizhaoLi2018 . I suspect the azure builds fail because I have updated the docker images that they are using to be compatible (and require) C++17. I think if you merge in the C++17 branch, it will pass. You can try that (fixing potential merge conflicts) or just wait for the C++17 branch to be merged first - it is close.

jbakosi:
This now seems to be passing on azure.

jbakosi:
Sure, but let us allow @adityakpandare take a look as well.

src/Control/Inciter/InputDeck/Grammar.hpp
```diff
@@ -229,6 +229,15 @@ namespace grm {
...
           Message< Stack, ERROR, MsgKey::BC_EMPTY >( stack, in );
+
+      // Put in default farfield pressure if not specified by user
+      // if outlet BC is configured for this compflow system
+      auto& bcoutlet = stack.template get< tag::param, eq, tag::bcoutlet >();
+      if (!bcoutlet.empty() || bcoutlet.size() != neq.get< eq >()) {
+        auto& fp =
+          stack.template get< tag::param, eq, tag::farfield_pressure >();
+        if (fp.size() != bcoutlet.size()) fp.push_back( 1.0 );
+      }
```
jbakosi: This block is executed when a `compflow ... end` input file block has finished parsing. What we do here is to set the default farfield pressure. The logic is that if outlet bc is specified, there should be a farfield pressure as well. If this is not given, we set it to 1.0, if given, the given value is used.

We set the defaults this way because the farfield pressure value is not stored inside the `CompFlowPDEParameters` tuple in `Control/Inciter/Types.hpp`. Its type is a `std::vector` because there can be potentially multiple `compflow ... end` blocks in a single input file and we need to be able to address each individually. This makes not much sense for a compflow block, but compflow is just another PDE system type in the parser/grammar, which by design allows multiple coupled PDE systems, e.g., flow (governing conservation of mass momentum and energy) + combustion (governing scalars only). This is why the logic says
```
      if (!bcoutlet.empty() || bcoutlet.size() != neq.get< eq >()) {
```
which detects if outlet bc is configured by the user at all for this compflow block (not empty) or if this is not the first one (and the previous one(s) potentially have outlet bcs), this most recently parsed compflow block does not have outlet bc. Then inside this if block, we detect if the farfield pressure is given. If not, we assign it.

src/Control/Inciter/InputDeck/Grammar.hpp
```diff
@@ -581,14 +590,37 @@ namespace deck {
...
                                     param > {};
 
+  //! put in PDE parameter for equation matching keyword
+  template< typename eq, typename keyword, typename p,
+            class kw_type = tk::grm::number >
+  struct parameter :
+         tk::grm::process< use< keyword >,
+                           tk::grm::Store_back< tag::param, eq, p >,
+                           kw_type > {};
+
```
jbakosi: I had to move this higher in the file so the new `subsonic_bc` struct below can _see_ it.

Sorry for this confusion here: there _is_ a `parameter` in `tk::grm` (in `Control/CommonGrammar.hpp`), and also another one here.

src/Control/Inciter/InputDeck/Grammar.hpp
```diff
@@ -581,14 +590,37 @@ namespace deck {
...
+                                        param > > > {};
+
+  //! Farfield boundary conditions block
+  template< class keyword, class eq, class param >
+  struct subsonic_bc :
+         pegtl::if_must<
+           tk::grm::readkw< typename use< keyword >::pegtl_string >,
+           tk::grm::block<
+             use< kw::end >,
+             parameter< eq, kw::farfield_pressure, tag::farfield_pressure >,
```
jbakosi: I have created this separate `subsonic_bc` block that takes the extra `farfield_pressure` input line, because if we had kept it inside the more generic `bc` struct above, then _all_ PDE types that use _any_ BC must have a storage for farfield, and this makes no sense for Transport. This allows having a storage for `farfield_pressure` in `CompFlowPDEParameters` and nowhere else, because `subsonic_bc` is only given in the `compflow` struct grammar block, see below.

src/Control/Inciter/InputDeck/Grammar.hpp
```diff
@@ -733,7 +757,9 @@ namespace deck {
                            bc< kw::bc_dirichlet, tag::compflow, tag::bcdir >,
                            bc< kw::bc_sym, tag::compflow, tag::bcsym >,
                            bc< kw::bc_inlet, tag::compflow, tag::bcinlet >,
-                           bc< kw::bc_outlet, tag::compflow, tag::bcsubsonicoutlet >,
+                           subsonic_bc< kw::bc_outlet,
+                                        tag::compflow,
+                                        tag::bcoutlet >,
```
jbakosi: This is what puts in the grammar/parsing block, specific to outlet + farfield pressure, _only_ for compflow and not for transport or multimat (at this point).

src/Control/Inciter/InputDeck/InputDeck.hpp
```diff
@@ -184,7 +183,7 @@ class InputDeck : public tk::TaggedTuple< InputDeckMembers > {
                                    kw::bc_inlet,
                                    kw::bc_outlet,
                                    kw::bc_extrapolate,
-                                   kw::p_farfield,
+                                   kw::farfield_pressure,
```
jbakosi: Renamed it.

src/Control/Inciter/Types.hpp
```diff
@@ -161,17 +156,19 @@ using TransportPDEParameters = tk::TaggedTuple< brigand::list<
...
+    tag::depvar,        std::vector< char >
+  , tag::physics,       std::vector< PhysicsType >
+  , tag::problem,       std::vector< ProblemType >
+  , tag::bcdir,         std::vector< std::vector<
+                          kw::sideset::info::expect::type > >
+  , tag::bcsym,         std::vector< std::vector<
+                          kw::sideset::info::expect::type > >
+  , tag::bcinlet,       std::vector< std::vector<
+                          kw::sideset::info::expect::type > >
+  , tag::bcoutlet,      std::vector< std::vector<
```
jbakosi: Also renamed this. @adityakpandare is that okay with you? I thought this was okay because there is not other "outlet" bc here.

src/Control/Inciter/Types.hpp
```diff
@@ -161,17 +156,19 @@ using TransportPDEParameters = tk::TaggedTuple< brigand::list<
...
+                          kw::sideset::info::expect::type > >
+  , tag::farfield_pressure,
+                        std::vector< kw::farfield_pressure::info::expect::type >
```
jbakosi: This is where the farfield pressure values (for potentially multiple compflow blocks) are now stored.

src/Control/Keywords.hpp
```diff
@@ -1768,23 +1768,23 @@ struct ncomp_info {
...
   struct expect {
     using type = tk::real;
     static constexpr type lower = 0.0;
     static std::string description() { return "real"; }
   };
 };
-using p_farfield =
-          keyword< p_farfield_info, TAOCPP_PEGTL_STRING("p_farfield") >;
+using farfield_pressure = keyword< farfield_pressure_info,
+                            TAOCPP_PEGTL_STRING("farfield_pressure") >;
```
jbakosi: Just a rename.

src/PDE/CompFlow/DGCompFlow.hpp
```diff
@@ -820,13 +820,14 @@ class CompFlow {
                     tk::real, tk::real, tk::real, tk::real,
                     const std::array< tk::real, 3 >& )
     {
-      const auto p_farfield = g_inputdeck.get< tag::bc, tag::p_farfield >();
+      auto fp =
+        g_inputdeck.get< tag::param, eq, tag::farfield_pressure >()[ system ];
```
jbakosi: And this how we access the value from the compflow parameter block. IT is now a vector so we also must index it by the `system` index, i.e., _"which compflow block's farfield pressure you want"?_

tests/regression/inciter/compflow/Euler/GaussHump/gauss_hump_pdg.q
```diff
@@ -26,7 +26,7 @@ inciter
       sideset 2 end
     end
     bc_outlet
-      p_farfield 1.0
+      farfield_pressure 1.0
```
jbakosi: This of course also changes, since I renamed the keyword. Also, this is now optional, because the default is 1.0 but I just kept it.

jbakosi:
@WeizhaoLi2018 Some of the pdg tests take a long time. Can you shorten some of them?
```
 1/11 Test #372: inciter:compflow_euler_sedovblastwave_pdg_pe4 .............   Passed    6.20 sec
 2/11 Test #366: inciter:compflow_euler_sodshocktube_pdg_pe4 ...............   Passed   10.10 sec
 3/11 Test #370: inciter:compflow_euler_sedovblastwave_pdg_pe1 .............   Passed   13.79 sec
 4/11 Test #367: inciter:compflow_euler_sodshocktube_pdg_u0.8_pe4 ..........   Passed   18.12 sec
 5/11 Test #373: inciter:compflow_euler_sedovblastwave_pdg_u0.9_migr_pe4 ...   Passed   21.88 sec
 6/11 Test #365: inciter:compflow_euler_sodshocktube_pdg_pe1 ...............   Passed   26.19 sec
 7/11 Test #328: inciter:compflow_euler_gauss_hump_pdg_u0.8_migr_pe4 .......   Passed   57.89 sec
 8/11 Test #368: inciter:compflow_euler_sodshocktube_pdg_u0.8_migr_pe4 .....   Passed   62.24 sec
 9/11 Test #327: inciter:compflow_euler_gauss_hump_pdg_u0.8_pe4 ............   Passed   79.98 sec
10/11 Test #326: inciter:compflow_euler_gauss_hump_pdg_pe4 .................   Passed  103.32 sec
11/11 Test #325: inciter:compflow_euler_gauss_hump_pdg_pe1 .................   Passed  152.91 sec
```

src/Control/Inciter/Types.hpp
adityakpandare: I think we should keep it as ```bcsubsonicoutlet```. We will soon have a characteristic outlet BC which detects if the flow at the boundary-face is sub/super sonic and behaves accordingly. Does that make sense?

jbakosi: Yeah okay. Or we can rename this when we implement that. I'm fine either way.

jbakosi:
Go for it @WeizhaoLi2018 ! Thanks, again, for the hard work on this. This is excellent!

PR #358, opened 2019-11-01
Adding communication before indicator calculation

WeizhaoLi2018:
This PR includes the commit to add solution communication before indicator calculation so that inter element indicator could be applied.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/358)

### DIFF
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -52,6 +52,7 @@ DG::DG( const CProxy_Discretization& disc,
   m_ncomfac( 0 ),
   m_nadj( 0 ),
   m_nsol( 0 ),
+  m_idof( 0 ),
   m_ninitsol( 0 ),
   m_nlim( 0 ),
   m_fd( Disc()->Inpoel(), bface, tk::remap(triinpoel,Disc()->Lid()) ),
@@ -1029,7 +1030,6 @@ DG::next()
 // Advance equations to next time step
 // *****************************************************************************
 {
-  std::cout << " start to comsol" << std::endl;
   // communicate solution ghost data (if any)
   if (m_ghostData.empty())
     comsol_complete();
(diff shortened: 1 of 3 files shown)


jbakosi:
Hey @WeizhaoLi2018 : which parallel pdg test case fails? Remember we cannot use `-q` with migration.

src/Inciter/DG.cpp
```diff
@@ -1115,6 +1091,97 @@ DG::comsol( int fromch,
...
+    Assert( j >= m_fd.Esuel().size()/4, "Receiving solution non-ghost data" );
+    auto b = tk::cref_find( m_bid, j );
+    Assert( b < m_ndofc[0].size(), "Indexing out of bounds" );
+    m_ndofc[0][b] = ndof[i];
+  }
+
+  // if we have received all ndof ghost contributions from those chares we
+  // communicate along chare-boundary faces with, solve the system
+  if (++m_nsol == m_ghostData.size()) {
+    m_nsol = 0;
```
jbakosi: You are reusing the counter `m_nsol` for detecting completeness of the communication for `comdof`, and the same is used for `comsol`. These will overwrite each other if/when `comsol` and `comdof` are overlapped -- unless you make sure they cannot overlap. I would use a new counter.

jbakosi:
> Thank you pointing out this! You are right! I gave it a new counter and now there is no issue for quiescence. But I still have question about this. In theory, for each chare, we will have started ndof communication after solution communication is completed. So I thought comsol will definitely happen before comdof. Then why are they having overlap?

`comdof()` on chare A can be still called from chare B while chare A is still running the solution communication if chare B went ahead and was ready with _its_ solution communication. This is why it is important not to write to the same places in `DG::sendInd()` and `DG::comdof()`. This is now fine, because the former writes to `m_u` and `m_p` while the latter only writes to `m_ndofc` and now `m_idof`. So there is no contention of writing to the same memory location.

PR #366, opened 2019-12-06
ALECG

jbakosi:
But way too much changes so it is high time to do a review and a merge to avoid major code divergence. No need to review here, instead I'll setup a time and let's get together in person next week, so I'll explain the major changes and reasons behind them.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/366)


jbakosi:
Alright everyone, this is now in good shape, I think, ready to merge in to `develop`. Please review and let's discuss.

I have tested convergence using vortical flow, checks out as second order for all fields in the diagnostics file. Also, the results are nicely reproducible in serial, parallel, parallel + virtualization, parallel + virtualization + migration with the limiter on. Also added regression tests exercising `Transport`: Gauss hump, and `CompFlow`: Sod and vortical flow. Will probably continue with other MMS tests, e.g. nonlinear energy growth, which will test time-dependent Dirichlet BCs and source terms.

PR #380, opened 2020-01-14
Improvements to DiagCG performance via more task parallelism

jbakosi:
This set of commits collects multiple steps of improvements that enable finer grains for task-parallelism in `DiagCG`. At a high level the steps are

* Break up the rhs calculation into gather and scatter steps in a single time step. Since the gather step does not depend on `dt`, it can be started before the new time step size is available, allowing more overlap of this computation step with communication (for computing `dt`, immediately following gather).

* Break up the element list of a mesh chunk held by a chare into two lists: those elements that contribute to chare-boundary nodes and those that only contribute to internal nodes. This allows performing the scatter step on the boundary element list first, followed by starting communication, followed by computing the internal elements. This allows better overlap of communication and computation because the chare-boundary element list is much small smaller compared to the internal element list.

The result of the above two improvements is reduced idle time between time steps and better CPU utilization and improved overall performance. The reduction in idle time at 14,400 CPUs running a 800M-cell mesh is about 50% -> 10% and the performance improvement is about 15% reduction of grind-time. Will still have to test if this also yields better strong scaling.

Note that the above does _not_ split PDE operators into tasks. Depending on the internals of the time step, specific to a scheme, here `DiagCG`, such a split may or may not yield better performance and/or more overlap of communication and computation. For `DiagCG`, as is written currently, both the gather and scatter steps would have to be called after each other for all operators, and this is required for correctness because gather must complete before scatter can start. Splitting tasks into operators with `DiagCG` would entail having to recompute some heavy stuff in elements, e.g., gradients of the shape functions, so splitting operators that need derivatives, would make it necessary to recompute these for each operator, so it is better to combine them. Splitting operators, however, may lead to better performance and/or more overlap of communication and computation if the operators can be scheduled at different times during the time step. This is _not_ the case for `DiagCG` at this point, so it yields no improvement. However, other schemes, e.g., `ALECG` and `DG` may have their own reasons as to why splitting operators on the rhs into their own tasks might make sense.

Overall this PR improves `DiagCG` performance by about 20%, measured on 14,400 CPUs using a 800M-cell mesh. It also improves CPU utilization by decreasing idle times from about 50% to 10%, using an eyeball metric at 14,400 CPUs using projections.

Grind-time is now measured _not_ between two one-liner status outputs but for a single time step (but still only output with the one-liner status as before).

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/380)


src/Inciter/Discretization.cpp
```diff
@@ -506,6 +521,13 @@ Discretization::status()
   // Query after how many time steps user wants TTY dump
   const auto tty = g_inputdeck.get< tag::interval, tag::tty >();
 
+  // estimate grind time (taken between this and the previous time step)
+  using std::chrono::duration_cast;
+  using ms = std::chrono::milliseconds;
+  using clock = std::chrono::high_resolution_clock;
+  auto grind_time = duration_cast< ms >(clock::now() - m_prevstatus).count();
+  m_prevstatus = clock::now();
```
jbakosi: Note to myself: This chunk of code really only need to be done for chareid 0, since `m_prevstatus` is only used by that, but I guess it does not really hurt much.

jbakosi:
https://github.com/quinoacomputing/quinoa/pull/380/commits/3819eb0a5308d9ee2aed15a9ccc513776e03ec1e implements a local (i.e., independently on/by each chare) mesh node reordering that generates local ids based on the PDE operator access pattern, specific to `DiagCG`.

For the cases tested so far, using the current algorithm, this does not yield a significant difference in performance, but we should keep this, in a user configurable-fashion (off by default), so it is handy in the future.

How this is done in `DiagCG` also serves as an example of how this can be done for `DG` and `ALECG`.

jbakosi:
Hey @adityakpandare ! I have reverted the splits in 54de70b, because of the reasons explained in its commit message. I think you have seen all of this except for that commit. I think I'll merge this.

jbakosi:
Note: Both splits, discussed in the first message of this PR have been reverted. See more explanation of the commit message of https://github.com/quinoacomputing/quinoa/commit/54de70b1ffc92a44d8dbbbc7594857fbb298f5b0. However, other changes in this PR are useful, so merging it in.

PR #400, opened 2020-04-03
History output for nodal schemes in inciter

jbakosi:
_Note: Let's merge #397 as well as #399 first!._

This adds history output in points. Example:
```
inciter
  ...
  history
    interval 2
    format   scientific
    precision 6
    point 0.1 0    0 end
    point 0.9 0.02 0 end
  end
  ...
end
```
This will save time history in spearate files as:
```
 * Output time history:
 ----------------------
     At point { 0.1 0 0 }                     : out.hist.{0.1_0_0}
     At point { 0.9 0.02 0 }                  : out.hist.{0.9_0.02_0}
```

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/400)


jbakosi:
Still have to add a couple of regression tests to exercise and test history output, will push it here.

PR #403, opened 2020-04-09
Fixed bug in conserved variable reconstruction, for boundary cells.

adityakpandare:
For reconstruction of conserved variables on boundary cells, primitive variables are needed so that BCs are used correctly. Previously, zeros were being sent for this, resulting in incorrect BCs for conserved variables.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/403)


jbakosi:
This PR has also fixed the multimat tests that have been failing on TC with Intel for a long time! Yohooo! ;-)

adityakpandare:
I had to widen some tolerances, and with that, could get away with narrowing some others. All in all, good. 😄

PR #408, opened 2020-04-13
fixed missing std namespace

charest:
Found when testing #404

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/408)


jbakosi:
Cool! Thanks. Which compiler (and its version) triggered this? Just curious.

charest:
I was using gcc 9.2.0

PR #410, opened 2020-04-17
Revamp source

jbakosi:
This does a partial revamp of how we use optional source terms. This is basically a WIP towards a vectorizable source integral in ALECG, but since the source terms are connected across the different discretization schemes and problem types, the changes here affect all source-related function interfaces. This is an attempt to make the source-term-related functions a little more specific to the type of the PDE, which makes it more amenable to vectorization. ALECG's source integral is not yet vectorized, because the intel compiler segfaults on calling a static member function from a vectorized loop. So that part is left for future work, nevertheless the changes here are a good idea, IMO.

I will disable testing on this because ALECG tests will fail at this point, due to reorganizing some of its loops. However, we will only update baselines once the gradient caluclation has also been revamped and made ready for vectorization. Other than the ALECG tests, the unit tests, as well as non-ALECG regression tests should pass.

_Note that this is to be merged to branch `vec_domain`, as this is a sub-task that belongs to that line of work, and when `vec_domain` is ready, reviewed, and passes all (updated) tests we will merge that into `develop`._

Thanks, in advance, for a partial code review on this. (`vec_domain`, when ready, will be close-to-impossible to read based on the diffs, and so this should help in catching problems better.)

_PS:_ Tested using aa2faf010, without vectorization, using a 48M-cell mesh on a single grizzly node, we are now only 2.7x slower than chicoma on the same problem on the same hardware. (o

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/410)


jbakosi:
Thanks guys, will consolidate this to another branch.

PR #412, opened 2020-04-17
Vectorize ALECG's nodal gradients using point loop

jbakosi:
Vectorize ALECG's CompFlow's nodal gradients calculation using a point-based algorithm

This gets a modest speedup of about 10% measured over the full time step.

Note the merge to branch `src` - this is a partial CR. (ALECG regression tests will fail, canceled CI.)

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/412)


jbakosi:
After 43df6d2, the speedup due to this change-set is about 20%.

jbakosi:
After 8ee2a06 and bcfb874 the speedup tue to this change-set is about 27%.

jbakosi:
Thanks guys, will consolidate this to another branch.

PR #421, opened 2020-05-15
Local time stepping for ALECG

jbakosi:
_Note_: merge to branch `vec`. Need to get that one in first. Can you guys take a look at `vec`? Can we get `vec` in?

This adds the capability to ALECG to march to steady state using local time stepping.

By default this is off, and can be enabled by, e.g.,
```
inciter
  ...
  steady_state true
  residual 1.0e-8
  ...
end
```
Note that if marching to steady state is on, `nstep` and `term` are still considered when deciding on the end of time stepping: either of `nstep`, `term`, or the L2-norm of the residual, configured by `residual` above, reaching their respective condition, time stepping will stop.

No regression tests yet, but this is so far tested with vortical flow and the Onera M6 wing, for both this makes a difference in convergence rate when a final steady state can be assumed. Note that we do not deactivate any nodes, but local time stepping allows signals to propagate faster based on local time step sizes.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/421)


jbakosi:
Here is a convergence plot, comparing local (red) with global (green) time stepping for the vortical flow problem. For this flow, it is not expected to make a large difference, but it still helps.
image

PR #423, opened 2020-05-21
Farfield

jbakosi:
_Note: merge to `local_timestepping`. Let's do that first!_

This puts in farfield BCs, and better boundary-point normal calculations into ALECG.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/423)


jbakosi:
Hey @adityakpandare and @charest. This has been kinda stalled, but I have two more PRs to build on it, so I'd like to merge this. I _think_ it is correct, but I'm not sure if it is bug-free. But if I start pushing more commits to this, it will grow out of control. Can you please take a look and let's merge this in? Thx.

PR #443, opened 2020-08-31
Outref P0

jbakosi:
This is a WIP PR, to avoid huge change sets. This is work towards saving field output on uniformly refined meshes for DG. For now, the logic is in place to call AMR doing a uniform mesh refinement followed by a derefinement for output field data. Only the DG(P0), cell-centered solution is saved for now. The next step after this is to save a higher-fidelity DG(P1) solution, which will be another PR.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443)

Oops. I think this branch was forked from the 'amr' branch. Perhaps it is better to do a PR on that one first.


jbakosi:
Yeah, we should merge #444 first.

adityakpandare (changes requested):
Thanks for putting this in @jbakosi. I have a few questions.

> Reviewed 17 of 22 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-:-MGQGwqRAY0XPJQvBQUB:bsk90dr)* status: 17 of 22 files reviewed, 5 unresolved discussions (waiting on @adityakpandare, @jbakosi, and @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 4090 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGOuo6W9bsq3XsdDjvM:-MGOuo6X-ZaG8Iez005J:b7hoysa) ([raw file](https://github.com/quinoacomputing/quinoa/blob/d8103cbb36c0800e5719c8c613040b1300798c5c/src/Control/Keywords.hpp#L4090)):*
> ```cpp
>   };
> };
> using refined =keyword< refined_info, TAOCPP_PEGTL_STRING("refined") >;
> ```

Would it be better to call this "outputref" or "refinedout" or something similar? I know that this keyword is in the `plotvar ... end` block so there's potentially no ambiguity about what "refined" means, but just curious about what you think.
___
*[src/Inciter/DG.cpp, line 1338 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQFQ896OWZYPwq6k5K:-MGQFQ896OWZYPwq6k5L:b-np8x4s) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1338)):*
> ```cpp
>     if (g_inputdeck.get< tag::cmd, tag::io, tag::refined >()) {
>       const auto& tr = tk::remap( m_fd.Triinpoel(), d->Gid() );
>       d->Ref()->outref( m_fd.Bface(), {}, tr, c );
> ```

How is the `RefMode` argument passed to `Refiner::outref()`? I'm curious because I don't see the call doing the output derefinement (using `OUTDEREF`).
___
*[src/Inciter/DG.cpp, line 1386 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQGJAX9BmTIF0zfdw0:-MGQGJAX9BmTIF0zfdw1:burta47) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1386)):*
> ```cpp
>     elemfieldnames.insert( end(elemfieldnames), begin(n), end(n) );
>     //auto nn = eq.nodalFieldNames();
>     //nodalfieldnames.insert( end(nodalfieldnames), begin(nn), end(nn) );
> ```

So currently there's no nodal-output in the exo file, correct? 
___
*[src/Inciter/DG.cpp, line 1424 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQGhi7B1kq-ltE_SEm:-MGQGhi7B1kq-ltE_SEn:b-adci47) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1424)):*
> ```cpp
>   //std::vector<tk::real> ndof( begin(m_ndof), end(m_ndof) );
>   //ndof.resize( nielem );  // cut off ghosts
>   //elemfields.push_back( ndof );
> ```

Why can't we leave the adaptive-indicator in the exodus output? I ask because that's also cell-based output.
___
*[src/Inciter/Refiner.cpp, line 983 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPesYY6wmyA3FgwZ0S:-MGPesYY6wmyA3FgwZ0T:bfq8xbs) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L983)):*
> ```cpp
> 
>   // // Free up memory if no dtref
>   // if (!g_inputdeck.get< tag::amr, tag::dtref >()) {
> ```

Shouldn't we free up memory even in case the output-refinement is not configured?
___
*[src/Inciter/Refiner.cpp, line 1126 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPfBrs3Qa1lTlJ-E79:-MGPfBrs3Qa1lTlJ-E7A:bx8ol7u) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L1126)):*
> ```cpp
> 
> ```

Could you explain why these if-tests are empty?
___
*[tests/regression/inciter/compflow/Euler/VorticalFlow/CMakeLists.txt, line 81 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPdrHn-vP3whtuey1G:-MGPdrHn-vP3whtuey1H:bro2g8g) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/tests/regression/inciter/compflow/Euler/VorticalFlow/CMakeLists.txt#L81)):*
> ```CMake
>                     BIN_BASELINE vortical_flow_outref_dg.std.exo
>                     BIN_RESULT out.e-s.0.1.0
>                     BIN_DIFF_PROG_CONF exodiff_dg.cfg
> ```

Cool! Using the same diagnostic file as the unrefined output test. That's a good sanity check.

jbakosi (commented):
Sure. I'm glad the existing design in `Refiner` could be reused for this purpose without much pain. Thanks for the extra set of eyes!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-:-MGjMqqY5J4ALRpFT_MP:b8xvh1s)* status: 17 of 22 files reviewed, 5 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 4090 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGOuo6W9bsq3XsdDjvM:-MGjHja_BBERwQ8qcYRi:b-d2vf6h) ([raw file](https://github.com/quinoacomputing/quinoa/blob/d8103cbb36c0800e5719c8c613040b1300798c5c/src/Control/Keywords.hpp#L4090)):*

Would it be better to call this "outputref" or "refinedout" or something similar? I know that this keyword is in the `plotvar ... end` block so there's potentially no ambiguity about what "refined" means, but just curious about what you think.
</blockquote></details>

Yeah, that is how I was thinking about this. Since in the input file it will be inside the `plotvar ... end` block, it will be clear. Example:
```
plotvar
    refined true
    interval 5
  end
```

If this becomes an ambiguity, we will fix it.
___
*[src/Inciter/DG.cpp, line 1338 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQFQ896OWZYPwq6k5K:-MGjILLDBMB-sv12fPwU:bm5mxaq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1338)):*

How is the `RefMode` argument passed to `Refiner::outref()`? I'm curious because I don't see the call doing the output derefinement (using `OUTDEREF`).
</blockquote></details>

Rather ambiguously: The default last function argument is `RefMode::OUTREF`:
```
    //! Start mesh refinement (for field output)
    void outref( const std::map< int, std::vector< std::size_t > >& bface,
                 const std::map< int, std::vector< std::size_t > >& bnode,
                 const std::vector< std::size_t >& triinpoel,
                 CkCallback c,
                 RefMode mode = RefMode::OUTREF );
```
Should we not use a default argument and require the RefMode argument instead? The idea behind the default argument is that this is called from `DG::writeFields()` and this way `DG` (client code) does not even need to know about the type `RefMode`which is defined inside `Refiner` and so is not visible to client code. The design is that if there is a call to `Refiner::outref()`, that must always be followed by a call to it again with `RefMode::OUTDEREF`, and that is _not_ done by client code (e.g., `DG`), but _automatic_ from inside `Refiner`. This happens after `Refiner::outref()` is finished, in `Refiner::next()`. (An outref must be followed by an outderef, because whatever the state of the mesh is at a given time (perhaps even after during-time-stepping refinement, this outref+outderef simply is superimposed on it, and so outderef must be followed by an outderef. If not, that is a problem, and I wanted to reduce the possibility of not calling outderef by the design.)
___
*[src/Inciter/DG.cpp, line 1386 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQGJAX9BmTIF0zfdw0:-MGjKA4R9n0x-wdyEBuJ:b-r2o4bi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1386)):*

So currently there's no nodal-output in the exo file, correct? 
</blockquote></details>

Correct. That's a next step - as we discussed.
___
*[src/Inciter/DG.cpp, line 1424 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQGhi7B1kq-ltE_SEm:-MGjKGBv4glrLAzqNHJG:b5j2hwf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1424)):*

Why can't we leave the adaptive-indicator in the exodus output? I ask because that's also cell-based output.
</blockquote></details>

We can, but I wanted keep it simple for now so I took it out. I am putting it back in right now, as I am working on making this refined field output work for p-adaptive DG.
___
*[src/Inciter/Refiner.cpp, line 983 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPesYY6wmyA3FgwZ0S:-MGjKUBJ0X2pwbw3bsbF:b-bjt0vz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L983)):*

Shouldn't we free up memory even in case the output-refinement is not configured?
</blockquote></details>

Yeah, that is an optimization I did not attempted yet. But good point. I will try that.
___
*[src/Inciter/Refiner.cpp, line 1126 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPfBrs3Qa1lTlJ-E79:-MGjKrum59BFfvC0TIgH:b-5s230x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L1126)):*

Could you explain why these if-tests are empty?
</blockquote></details>

This is inside `Refiner::solution()`. This function is called whenever `Refiner` needs to evaluate the solution on the current mesh. So far, this is done inside `Refiner::refinementFields()`, which collects output fields from Refiner (for debugging purposes), and in `Refiner::errorRefine()`, which is called when we do error-based refinement (in either before time stepping or during time stepping). So far this is only implemented (and somewhat tested) for node-centered schemes and code is not written yet for cell-centered cases (and for output-refined this will probably not even make sense, so we may just have to throw if that is called, I guess). I have not even thought much about those use cases / if branches so far; I only put in these empty branches when I converted the previous bool that became the current `RefMode`, so we don't forget to deal with those cases.
___
*[tests/regression/inciter/compflow/Euler/VorticalFlow/CMakeLists.txt, line 81 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPdrHn-vP3whtuey1G:-MGjMpG_3gxVUP-L6_lG:b-7xtei2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/tests/regression/inciter/compflow/Euler/VorticalFlow/CMakeLists.txt#L81)):*

Cool! Using the same diagnostic file as the unrefined output test. That's a good sanity check.
</blockquote></details>

Yeah.

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443)* status: 17 of 22 files reviewed, 1 unresolved discussion (waiting on @adityakpandare, @jbakosi, and @WeizhaoLi2018)

___
*[src/Inciter/DG.cpp, line 1338 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGQFQ896OWZYPwq6k5K:-MHCittR9qP7aI3RQVzv:bxpbzy9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/DG.cpp#L1338)):*

Rather ambiguously: The default last function argument is `RefMode::OUTREF`:
```
    //! Start mesh refinement (for field output)
    void outref( const std::map< int, std::vector< std::size_t > >& bface,
                 const std::map< int, std::vector< std::size_t > >& bnode,
                 const std::vector< std::size_t >& triinpoel,
                 CkCallback c,
                 RefMode mode = RefMode::OUTREF );
```
Should we not use a default argument and require the RefMode argument instead? The idea behind the default argument is that this is called from `DG::writeFields()` and this way `DG` (client code) does not even need to know about the type `RefMode`which is defined inside `Refiner` and so is not visible to client code. The design is that if there is a call to `Refiner::outref()`, that must always be followed by a call to it again with `RefMode::OUTDEREF`, and that is _not_ done by client code (e.g., `DG`), but _automatic_ from inside `Refiner`. This happens after `Refiner::outref()` is finished, in `Refiner::next()`. (An outref must be followed by an outderef, because whatever the state of the mesh is at a given time (perhaps even after during-time-stepping refinement, this outref+outderef simply is superimposed on it, and so outderef must be followed by an outderef. If not, that is a problem, and I wanted to reduce the possibility of not calling outderef by the design.)
</blockquote></details>

This design makes sense to me; it's less error prone. No need for any change. I had missed the call to `outref()` with argument `RefMode::OUTDEREF` in `Refiner::next()`. Thanks for the explanation!
___
*[src/Inciter/Refiner.cpp, line 983 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPesYY6wmyA3FgwZ0S:-MHBzET1AMR-t_JyPjf2:b70tozj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L983)):*

Yeah, that is an optimization I did not attempted yet. But good point. I will try that.
</blockquote></details>

The memory free-up was there before too, so I would think that should work even with the output-refinement. Should we try to put it back and see if it works?
___
*[src/Inciter/Refiner.cpp, line 1126 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPfBrs3Qa1lTlJ-E79:-MHBzj3P7dhoQd1w54Bu:b-wurqvk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L1126)):*

This is inside `Refiner::solution()`. This function is called whenever `Refiner` needs to evaluate the solution on the current mesh. So far, this is done inside `Refiner::refinementFields()`, which collects output fields from Refiner (for debugging purposes), and in `Refiner::errorRefine()`, which is called when we do error-based refinement (in either before time stepping or during time stepping). So far this is only implemented (and somewhat tested) for node-centered schemes and code is not written yet for cell-centered cases (and for output-refined this will probably not even make sense, so we may just have to throw if that is called, I guess). I have not even thought much about those use cases / if branches so far; I only put in these empty branches when I converted the previous bool that became the current `RefMode`, so we don't forget to deal with those cases.
</blockquote></details>

Ah okay, this function call is only used for node-centered schemes for now; makes sense.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443)* status: 17 of 22 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/Refiner.cpp, line 983 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPesYY6wmyA3FgwZ0S:-MHH7mBtDJBes2-kaUOd:bv52ae4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L983)):*

The memory free-up was there before too, so I would think that should work even with the output-refinement. Should we try to put it back and see if it works?
</blockquote></details>

The idea behind these frees was that when we are certain that the whole `Refiner` class is no longer needed, we free up the state. This should really have been done from Transporter, by calling a Charm++ function on the Refiner proxy to destroy the object. However, last time I checked, this just does not exist in Charm++. (Maybe it does now, I don't know.) So I put in these individual frees here to save memory because this allowed me to continue with some large meshes in the past. 

Now that we are using this for field output, `Refiner` has to stick around, at least for `DG`, at least when there is not `dtref`, and at least for now until we figure out what state is needed and what is not for outref.

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-:-MHHElLn8nzctRnuVr2o:bnfp4nl)* status: 17 of 22 files reviewed, all discussions resolved (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/Refiner.cpp, line 983 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/443#-MGPesYY6wmyA3FgwZ0S:-MHHEeQ-CWLwZ7nL3wKt:b-bzxlkw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e30befb7fb3b6a92cb221b92750ddbacd41cb664/src/Inciter/Refiner.cpp#L983)):*

The idea behind these frees was that when we are certain that the whole `Refiner` class is no longer needed, we free up the state. This should really have been done from Transporter, by calling a Charm++ function on the Refiner proxy to destroy the object. However, last time I checked, this just does not exist in Charm++. (Maybe it does now, I don't know.) So I put in these individual frees here to save memory because this allowed me to continue with some large meshes in the past. 

Now that we are using this for field output, `Refiner` has to stick around, at least for `DG`, at least when there is not `dtref`, and at least for now until we figure out what state is needed and what is not for outref.
</blockquote></details>

Ok, I understand now. Thanks for the explanation.

PR #455, opened 2020-11-03
Combine steps for DG node field output into single step

jbakosi:
This forks and builds on #451, so let's merge that one first.

Still no regression tests, but visual inspection of the element and node field output in paraview look okay for P0, P1, P2 coarse and refined mesh output.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/455)

PR #456, opened 2020-11-04
Time-dependent source for DG

adityakpandare:
This set of commits does a couple of things:
1. refactor code for user defined box-IC,
2. add vertex-based limiter for DG(P1),
3. added time-dependent sources for the box-IC for the DG discretization.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456)


jbakosi (changes requested):
Nice work! This nicely cleans up my previous butchery with the box ICs. Thanks! ;-) I only have a few comments/suggestions.

> Reviewed 64 of 64 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLPPZNQEzr0BhUkjGTG:b-mb9u2i)* status: all files reviewed, 4 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 86 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPJDmZ9Cu8L57Y5kxh:-MLPJDmZ9Cu8L57Y5kxi:br5t3yq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/CGCompFlow.hpp#L86)):*
> ```cpp
>     //! \param[in,out] inbox List of nodes at which box user ICs are set
>     void inIcBox( const tk::UnsMesh::Coords& coord,
>       std::vector< std::size_t >& inbox ) const
> ```

Should we _return_ `inbox` instead of passing it as a non-const reference? I think that would be more readable. Also if we rename `inIcBox` to `boxNodes`, calling code would look like:
```
auto boxnodes = boxNodes( coord );
```
or
```
auto boxnodes = ICBoxNodes( coord );
```
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 139 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPK8XUDAz55cfN31tg:-MLPK8XUDAz55cfN31th:b5rlkbl) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/CGCompFlow.hpp#L139)):*
> ```cpp
>             break;
>           }
>         }
> ```

The above is now a `O(n^2)` search if the box is large, i.e., comparable in size to that of the mesh. I would use `std::unordered_set< std::size_t >` for the type of `inbox`, to avoid scaling problems.
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 178 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOGtI-d52P19diEJu:-MLPOGtI-d52P19diEJv:b4ipm2c) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L178)):*
> ```cpp
>                         tk::real t,
>                         tk::Fields& unk ) const
>     {
> ```

Seems like repeating a bunch of code here from CG. Is there a way to reuse at least some? Not necessarily calling into CG from here, but to relocate common code to a free-function.
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 1075 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOpCID83oSf77E7xS:-MLPOpCID83oSf77E7xT:bs9s62i) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L1075)):*
> ```cpp
>       const std::vector< std::size_t >& ndofel,
>       tk::Fields& R ) const
>     {
> ```

This seems different enough to _not_ attempt code reuse between `cg::CompFlow::boxSrc()`.

adityakpandare (commented):
Thanks a lot; but you'd done all the hard work already :smiley:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLPdaf45e-idQ1HoE3U:b-d4vyq3)* status: 50 of 66 files reviewed, 4 unresolved discussions (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 86 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPJDmZ9Cu8L57Y5kxh:-MLPbBBk2I3ZIzq1W-84:b-ho5wpv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/CGCompFlow.hpp#L86)):*

Should we _return_ `inbox` instead of passing it as a non-const reference? I think that would be more readable. Also if we rename `inIcBox` to `boxNodes`, calling code would look like:
```
auto boxnodes = boxNodes( coord );
```
or
```
auto boxnodes = ICBoxNodes( coord );
```
</blockquote></details>

Agreed. Done.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 139 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPK8XUDAz55cfN31tg:-MLPbDuMEw7NP2rXcqgD:b-4rejzh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/CGCompFlow.hpp#L139)):*

The above is now a `O(n^2)` search if the box is large, i.e., comparable in size to that of the mesh. I would use `std::unordered_set< std::size_t >` for the type of `inbox`, to avoid scaling problems.
</blockquote></details>

Nice catch! Fixed it. Thanks!
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 178 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOGtI-d52P19diEJu:-MLPepmK30CBy8Z7NjYw:bajw2nf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L178)):*

Seems like repeating a bunch of code here from CG. Is there a way to reuse at least some? Not necessarily calling into CG from here, but to relocate common code to a free-function.
</blockquote></details>

I thought so too, but this function has an eos-call, which needs the equation tag (here it's `eq`), so I would have to template it on `eq` or something similar, like the functions in EoS.hpp. The other (better) option is to move this over to the Problem/UserDefined class; but then every compflow problem-class will have to have a member function `initializeBox()`, right?
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 1075 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOpCID83oSf77E7xS:-MLPdZMPBL3ey-U6YnJX:b-pbdqz3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L1075)):*

This seems different enough to _not_ attempt code reuse between `cg::CompFlow::boxSrc()`.
</blockquote></details>

Agreed.

jbakosi (changes requested):
Haha! Not as a source, which was unstable previously.. ;-)

> Reviewed 16 of 16 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLPjPUSAru2izQn-d4C:b-kluhzg)* status: all files reviewed, 4 unresolved discussions (waiting on @adityakpandare)

___
*[src/Inciter/ALECG.cpp, line 379 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPh6zj7YOwd1rMqr5C:-MLPh6zk8gDHKp807b7e:b-6w3sl5) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/Inciter/ALECG.cpp#L379)):*
> ```cpp
> 
>   // Determine nodes inside user-defined IC box
>   for (auto& eq : g_cgpde) m_boxnodes = eq.IcBoxNodes( d->Coord() );
> ```

Ugh! This okay, I guess, but we have to be careful because every equation system now _overwrites_ `m_boxnodes` instead of augmenting it. Is there a use-case in which `eq.IcBoxNodes()` would be _contributed_ to by more than a single eq system? (Sorry, I just realized this by looking at this piece of code.) If the answer to that question is yes, that we should revert this to passing a non-const ref, so `m_boxnodes` can be contributed to.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 138 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPi0G08MreO1EUyz7v:-MLPi0G08MreO1EUyz7w:b-8atzjp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/PDE/CompFlow/CGCompFlow.hpp#L138)):*
> ```cpp
>       for (ncomp_t i=0; i<x.size(); ++i) {
>         int boxed = 0;
>         if (inbox.find(i) != inbox.end()) boxed = 1;
> ```

```
bool boxed = inbox.find(i) != end(inbox);
```
or as `boxed` appears to be only used once, you could just move it down to
```
if (inbox.find(i) != end(inbox)) initializeBox(V, t, s);
```
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 178 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOGtI-d52P19diEJu:-MLPiwszArbC3cNKLYrQ:b-8vgtcj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L178)):*

I thought so too, but this function has an eos-call, which needs the equation tag (here it's `eq`), so I would have to template it on `eq` or something similar, like the functions in EoS.hpp. The other (better) option is to move this over to the Problem/UserDefined class; but then every compflow problem-class will have to have a member function `initializeBox()`, right?
</blockquote></details>

Or you can just move everything into a free function and parameterize it with the _result_ of the EOS call. That way you don't have to template it.
___
*[src/PDE/Transport/CGTransport.hpp, line 77 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPjDgK1DDKSaNl8S98:-MLPjDgK1DDKSaNl8S99:b-32sfa9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/PDE/Transport/CGTransport.hpp#L77)):*
> ```cpp
>     {
>       std::unordered_set< std::size_t > inbox;
>       return inbox;
> ```

```
return {};
```
will probably also work

adityakpandare (commented):
That's true :wink:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLPoCRhC-s13kaW2Nhv:b-wucw5)* status: all files reviewed, 4 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/ALECG.cpp, line 379 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPh6zj7YOwd1rMqr5C:-MLPmJyNE3-lmoTkpUOR:b-pgilf8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/Inciter/ALECG.cpp#L379)):*

Ugh! This okay, I guess, but we have to be careful because every equation system now _overwrites_ `m_boxnodes` instead of augmenting it. Is there a use-case in which `eq.IcBoxNodes()` would be _contributed_ to by more than a single eq system? (Sorry, I just realized this by looking at this piece of code.) If the answer to that question is yes, that we should revert this to passing a non-const ref, so `m_boxnodes` can be contributed to.
</blockquote></details>

I think different systems *can* contribute to it. Reverted this. Good catch! 
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 138 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPi0G08MreO1EUyz7v:-MLPo3eDDNZuTCMYI1ee:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/PDE/CompFlow/CGCompFlow.hpp#L138)):*

```
bool boxed = inbox.find(i) != end(inbox);
```
or as `boxed` appears to be only used once, you could just move it down to
```
if (inbox.find(i) != end(inbox)) initializeBox(V, t, s);
```
</blockquote></details>

Done.
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 178 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOGtI-d52P19diEJu:-MLTQ2to-qD_bJqELeFz:bvzkuy8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L178)):*

Or you can just move everything into a free function and parameterize it with the _result_ of the EOS call. That way you don't have to template it.
</blockquote></details>

Okay, I was able to do this. I've put the function into PDE/CompFlow/Problem/BoxInitialization.cpp; because it is equation dependent and we should be able to define similar functions for multimat and transport.
___
*[src/PDE/Transport/CGTransport.hpp, line 77 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPjDgK1DDKSaNl8S98:-MLPo6cj39b_iVFsyWEt:bma7zrb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/PDE/Transport/CGTransport.hpp#L77)):*

```
return {};
```
will probably also work
</blockquote></details>

This function doesn't return after this change.

jbakosi (approved):
> Reviewed 13 of 13 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/Inciter/ALECG.cpp, line 379 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPh6zj7YOwd1rMqr5C:-MLTYHwy0wczZuT4AYxI:b-tfd28j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/Inciter/ALECG.cpp#L379)):*

I think different systems *can* contribute to it. Reverted this. Good catch! 
</blockquote></details>

Okay, sorry about this.
___
*[src/PDE/CompFlow/DGCompFlow.hpp, line 178 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPOGtI-d52P19diEJu:-MLTYbS90qZkD-KXPOD5:bdovrez) ([raw file](https://github.com/quinoacomputing/quinoa/blob/01e06a6f2e8a376f850f242fbf5ad854118ccf8d/src/PDE/CompFlow/DGCompFlow.hpp#L178)):*

Okay, I was able to do this. I've put the function into PDE/CompFlow/Problem/BoxInitialization.cpp; because it is equation dependent and we should be able to define similar functions for multimat and transport.
</blockquote></details>

Sounds great!
___
*[src/PDE/CompFlow/Problem/BoxInitialization.cpp, line 160 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLTYkYmBi7LDXdG7xPm:-MLTYkYmBi7LDXdG7xPn:b-dv6coz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a434809fa48f5872dad5f29ee51451af6c0d7ef6/src/PDE/CompFlow/Problem/BoxInitialization.cpp#L160)):*
> ```cpp
> 
>   } else Throw( "IC box initiate type not implemented" );
> }
> ```

Awesome! So this is now reused between ALECG and DG.
___
*[src/PDE/Transport/CGTransport.hpp, line 77 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPjDgK1DDKSaNl8S98:-MLTYrqr2r9JEZLkbcpJ:b-voqgxl) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/PDE/Transport/CGTransport.hpp#L77)):*

This function doesn't return after this change.
</blockquote></details>

Even better! ;-)

adityakpandare (commented):
I'll review https://github.com/quinoacomputing/quinoa/pull/455 before, and once you merge that one, I'll merge this. Thanks for the quick turnaround!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLTZq06EPUh_PpVLlK3:bw2s4p7)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/Inciter/ALECG.cpp, line 379 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLPh6zj7YOwd1rMqr5C:-MLTZNReA8ykUr43GnZ0:bhzoc3h) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30d7b8d89cbf59f03faa5d7fc221a7a7b68a2465/src/Inciter/ALECG.cpp#L379)):*

Okay, sorry about this.
</blockquote></details>

Not at all. Iterations are the way to figure out what's the right thing to do. :smiley: 
___
*[src/PDE/CompFlow/Problem/BoxInitialization.cpp, line 160 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-MLTYkYmBi7LDXdG7xPm:-MLTZeIn5_OPoKCqOMyx:b-2logwn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a434809fa48f5872dad5f29ee51451af6c0d7ef6/src/PDE/CompFlow/Problem/BoxInitialization.cpp#L160)):*

Awesome! So this is now reused between ALECG and DG.
</blockquote></details>

Yes! Thanks for the help with the implementation!

jbakosi (approved):
Either way is fine.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456#-:-MLT_zbJD6nwacRtZL_q:bypykxx)* status: :shipit: complete! all files reviewed, all discussions resolved

jbakosi (approved):
> Reviewed 6 of 6 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456)* status: :shipit: complete! all files reviewed, all discussions resolved

jbakosi (approved):
> Reviewed 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/456)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #462, opened 2020-12-03
The multi-material THINC reconstruction for multimat and transport

adityakpandare:
This is a huge change-set including but most likely not limited to:
1. THINC reconstruction function for multi-material interfaces (see Pandare A. K., Waltz J., & Bakosi J. (2021) Multi-Material Hydrodynamics with Algebraic Sharp Interface Capturing. Accepted in Computers & Fluids), also called MM-THINC,
2. Interfaces to the MM-THINC function for multimat and transport, but hardcoded for multimat for now. Polymorphism like the one used for Riemann solvers will be needed so that the correct THINC-interface is called for multimat and transport.
3. Communication for min/max volume fractions in the neighborhood. This is currently unused, but very likely that it will be needed when we return to improving MM-THINC. This is also passed all they way down to the DG::rhs() so that it's ready to be used by THINC.
4. Added a problem for transport that tests MM-THINC reconstruction of a 4-species cylinder in a vortical flow (see S. Schofield, R. Garimella, M. Francois, R. Loubere, (2009) [A second-order accurate material-order-independent interface reconstruction technique for multi-material flow simulations, Journal of Computational Physics 228 (3) 731–745](https://doi.org/10.1016/j.jcp.2008.09.023)).

I might have missed out on other changes in this list. Apologies for the large changes; let's discuss more on a phone call if you think that will be simpler.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462)


jbakosi (changes requested):
Looks good, as far as I can understand. Few minor comments and pretty much ready to go.

> Reviewed 41 of 41 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-:-MOHhusV61K96nZzAtjd:b-nli7qf)* status: all files reviewed, 10 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 5538 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaP3d6RxhNAg7N_Eh:-MOHaP3d6RxhNAg7N_Ei:b-zf91hu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5538)):*
> ```cpp
>   static std::string name() { return "Interface sharpening"; }
>   static std::string shortDescription() { return
>     "Turn multi-material interface sharpening on/off"; }
> ```

Is there a reason a user might _not_ want interface sharpening? (I understand a developer's need for this, but for a user?)
___
*[src/Control/Keywords.hpp, line 5551 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOH_xxICtqyYS5yV6ew:-MOH_xxICtqyYS5yV6ex:b-okw2fp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5551)):*
> ```cpp
>     using type = int;
>     static std::string description() { return "string"; }
>     static std::string choices() { return "1 | 0"; }
> ```

Why not bool?
___
*[src/Control/Keywords.hpp, line 5562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaBPwEZzvHd17j_cH:-MOHaBPxExGVFED3nF-J:b-6jvk7u) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5562)):*
> ```cpp
>   static std::string longDescription() { return
>     R"(This keyword is used to specify the parameter for the interface
>        sharpening. The default value is 1.0. This parameter affects how many
> ```

I assume you set the default value somewhere else. If that is changed, it is likely that this comment will not be updated.
___
*[src/Inciter/DG.hpp, line 313 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHazl32k3jNYqZSOMn:-MOHazl32k3jNYqZSOMo:bp86ssp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Inciter/DG.hpp#L313)):*
> ```cpp
>     tk::Fields m_geoElem;
>     //! Vector of maximum volume fraction for each mesh element
>     tk::Fields m_vfm;
> ```

`m_maxvolfrac`?
___
*[src/PDE/DGPDE.hpp, line 130 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHc-PQ3RmCApELhRj7:-MOHc-PQ3RmCApELhRj8:bsbuk5n) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/DGPDE.hpp#L130)):*
> ```cpp
>     //! Public interface to find number of materials for the diff eq
>     std::size_t nmat() const
>     { return self->nmat(); }
> ```

What is nmat for transport? Number of scalar components?
___
*[src/PDE/Reconstruction.cpp, line 564 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHdAN17mGoHYRnHaOO:-MOHdAN17mGoHYRnHaOP:b-6f5b9o) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L564)):*
> ```cpp
>   [[maybe_unused]] const std::vector< std::size_t >& inpoel,
>   const Fields& U,
>   Fields& VolFracMax )
> ```

Why don't we `return` this instead?
___
*[src/PDE/Reconstruction.cpp, line 641 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHf2Po2JLpBZAMwedJ:-MOHf2Po2JLpBZAMwedK:b-23gqtg) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L641)):*
> ```cpp
>   const std::array< real, 3 >& ref_xp,
>   const Fields& U,
>   [[maybe_unused]] const Fields& P,
> ```

I see `P` being used below. Does the compiler warn on being unused?
___
*[src/PDE/Reconstruction.cpp, line 695 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHe1Eb5hyXdDD8dKeO:-MOHe1Eb5hyXdDD8dKeP:bubdqk9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L695)):*
> ```cpp
>   // Step-1: Perform THINC reconstruction
>   // create a vector of volume-fractions and pass it to the THINC function
>   Fields alSol(1, rdof*nmat);
> ```

I don't see a reason why this should be a `Fields` - it only has a single element. Why not just a `std::vector< tk::real >`?
___
*[src/PDE/Transport/DGTransport.hpp, line 95 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHgdcx2VySlDhTeJ7Y:-MOHgdcx2VySlDhTeJ7Z:b-df5xx8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/DGTransport.hpp#L95)):*
> ```cpp
>     std::size_t nmat() const
>     {
>       return m_ncomp;
> ```

I guess this isn't too bad after all.
___
*[src/PDE/Transport/DGTransport.hpp, line 389 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhJSg8Wzx0WEE1kU2:-MOHhJSg8Wzx0WEE1kU3:babssno) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/DGTransport.hpp#L389)):*
> ```cpp
>                  const std::array< std::vector< tk::real >, 3 >& coord,
>                  const tk::Fields& U,
>                  [[maybe_unused]] const tk::Fields& = tk::Fields() ) const
> ```

This should no longer be necessary.
___
*[src/PDE/Transport/Problem/CylVortex.hpp, line 13 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhYgk-3Xo5a21mOLS:-MOHhYgk-3Xo5a21mOLT:b-9g1oce) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/Problem/CylVortex.hpp#L13)):*
> ```cpp
>     node-centered continuous Galerkin (CG) and PDE/Transport/DGTransport.h
>     implementing cell-centered discontinuous Galerkin (DG) discretizations.
>     See PDE/Transport/Problem.h for general requirements on Problem policy
> ```

`Problem.hpp`

adityakpandare (commented):
Thanks for the quick review. I know this was a relatively large change.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-:-MOHziB-9MNvrWUk0USR:b-e9w4an)* status: all files reviewed, 10 unresolved discussions (waiting on @jbakosi and @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 5538 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaP3d6RxhNAg7N_Eh:-MOHsJIO6cbh_PyVxWGv:b-m4slp1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5538)):*

Is there a reason a user might _not_ want interface sharpening? (I understand a developer's need for this, but for a user?)
</blockquote></details>

One instance that comes to mind is when comparing the effects of interface sharpening/reconstruction. This might be for the purposes of reproducing results from our paper, or while trying out a new test problem.
___
*[src/Control/Keywords.hpp, line 5551 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOH_xxICtqyYS5yV6ew:-MOHsfkC5tma4KaM3lMw:b-8rkx93) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5551)):*

Why not bool?
</blockquote></details>

We would need integers to refer to other methods of interface reconstruction (in the near future we might want to add more sophisticated treatment for more than 2 materials), hence not a `bool`.
___
*[src/Control/Keywords.hpp, line 5562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaBPwEZzvHd17j_cH:-MOHuovk1VxmIoclR4IW:bg87oiq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5562)):*

I assume you set the default value somewhere else. If that is changed, it is likely that this comment will not be updated.
</blockquote></details>

Yeah... Should I just remove that part of the comment?
___
*[src/Inciter/DG.hpp, line 313 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHazl32k3jNYqZSOMn:-MOHutki-wlnp9kGC32P:b6f6zju) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Inciter/DG.hpp#L313)):*

`m_maxvolfrac`?
</blockquote></details>

Good idea. But it doesn't only store the max, but also the min... Any ideas based on that?
___
*[src/PDE/DGPDE.hpp, line 130 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHc-PQ3RmCApELhRj7:-MOHv1NE4J0LJ-cjRfUh:bk4r8ot) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/DGPDE.hpp#L130)):*

What is nmat for transport? Number of scalar components?
</blockquote></details>

Yes. It's defined in DGTransport.hpp.
___
*[src/PDE/Reconstruction.cpp, line 564 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHdAN17mGoHYRnHaOO:-MOHvyix0ssOjU-Po7G1:br4yc3j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L564)):*

Why don't we `return` this instead?
</blockquote></details>

I did not return it because DG allocates memory for `m_vfm`. It's just like `m_rhs` in DG.cpp, which is not returned by `eq.rhs()`.
___
*[src/PDE/Reconstruction.cpp, line 641 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHf2Po2JLpBZAMwedJ:-MOHx5JF1AYygpgMK9lH:b-4mvfix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L641)):*

I see `P` being used below. Does the compiler warn on being unused?
</blockquote></details>

Good catch! Removed it.
___
*[src/PDE/Reconstruction.cpp, line 695 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHe1Eb5hyXdDD8dKeO:-MOHzKKq-X-hc0kDhvIh:b-trgz64) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L695)):*

I don't see a reason why this should be a `Fields` - it only has a single element. Why not just a `std::vector< tk::real >`?
</blockquote></details>

Nice catch! Changed it to a vector.
___
*[src/PDE/Transport/DGTransport.hpp, line 389 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhJSg8Wzx0WEE1kU2:-MOHvM5UEzGOzUfCR79Z:bhpay5n) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/DGTransport.hpp#L389)):*

This should no longer be necessary.
</blockquote></details>

Yes, I guess I missed this when I was updating this branch based on changes in the mainline. I'll remove it. I have a question though: does the field output include a material-indicator field for MultiMat as well as Transport?
___
*[src/PDE/Transport/Problem/CylVortex.hpp, line 13 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhYgk-3Xo5a21mOLS:-MOHxmxX4KynMyKEBt1w:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/Problem/CylVortex.hpp#L13)):*

`Problem.hpp`
</blockquote></details>

Done.

jbakosi (approved):
Wasn't _that_ quick. ;-)

> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-:-MOI2eP273azGiNGt1JE:booaogm)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 5538 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaP3d6RxhNAg7N_Eh:-MOI2qJWEMM4XFu5_q6W:b-1fqx6k) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5538)):*

One instance that comes to mind is when comparing the effects of interface sharpening/reconstruction. This might be for the purposes of reproducing results from our paper, or while trying out a new test problem.
</blockquote></details>

Yeah, trying out a new test problem makes sense. Now that you added user-configurability, I guess it does not hurt much to leave it in there.
___
*[src/Control/Keywords.hpp, line 5551 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOH_xxICtqyYS5yV6ew:-MOI3-4f0ZoELYVo7I_h:b-8fx2u7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5551)):*

We would need integers to refer to other methods of interface reconstruction (in the near future we might want to add more sophisticated treatment for more than 2 materials), hence not a `bool`.
</blockquote></details>

:+1: 
___
*[src/Control/Keywords.hpp, line 5562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaBPwEZzvHd17j_cH:-MOI33252khtjGXkxMEv:bqi94dc) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5562)):*

Yeah... Should I just remove that part of the comment?
</blockquote></details>

Yeah, I think a misleading comment is worse than a nonexistent one. One can always grep for where that default is set (and to what value).
___
*[src/Inciter/DG.hpp, line 313 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHazl32k3jNYqZSOMn:-MOI3Fla71ExnStvHhG7:b4ab4s3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Inciter/DG.hpp#L313)):*

Good idea. But it doesn't only store the max, but also the min... Any ideas based on that?
</blockquote></details>

`m_volfracext` (for extrema)
___
*[src/PDE/Reconstruction.cpp, line 564 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHdAN17mGoHYRnHaOO:-MOI3VGc6cmkCtNU1Osh:b2xjkwe) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L564)):*

I did not return it because DG allocates memory for `m_vfm`. It's just like `m_rhs` in DG.cpp, which is not returned by `eq.rhs()`.
</blockquote></details>

Oh, okay. I did not connect it with the caller. Thx.
___
*[src/PDE/Transport/DGTransport.hpp, line 389 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhJSg8Wzx0WEE1kU2:-MOI3hQD8gV3ejGpz3T6:bt36a3d) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/DGTransport.hpp#L389)):*

Yes, I guess I missed this when I was updating this branch based on changes in the mainline. I'll remove it. I have a question though: does the field output include a material-indicator field for MultiMat as well as Transport?
</blockquote></details>

No. But it is good opportunity to add it (using the new output machinery). Give me call and we can add it together if that's faster.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462)* status: 34 of 44 files reviewed, all discussions resolved (waiting on @jbakosi and @WeizhaoLi2018)

___
*[src/Control/Keywords.hpp, line 5562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHaBPwEZzvHd17j_cH:-MOI5R2b14MpfLUqk-kA:b-b5cz09) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Control/Keywords.hpp#L5562)):*

Yeah, I think a misleading comment is worse than a nonexistent one. One can always grep for where that default is set (and to what value).
</blockquote></details>

True. Done.
___
*[src/Inciter/DG.hpp, line 313 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHazl32k3jNYqZSOMn:-MOI5VZc80EOatTcOD-C:bi0okjk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/Inciter/DG.hpp#L313)):*

`m_volfracext` (for extrema)
</blockquote></details>

Okay. I'll use `m_volfracExtr` if that's okay.
___
*[src/PDE/Transport/DGTransport.hpp, line 389 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOHhJSg8Wzx0WEE1kU2:-MOIBnps305PcaI8Hu2E:bzidavr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Transport/DGTransport.hpp#L389)):*

No. But it is good opportunity to add it (using the new output machinery). Give me call and we can add it together if that's faster.
</blockquote></details>

Okay, I added it. Thanks!

jbakosi (approved):
> Reviewed 10 of 10 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

WeizhaoLi2018 (changes requested):
> Reviewed 38 of 41 files at r1, 2 of 4 files at r2, 7 of 10 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462)* status: all files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/Limiter.cpp, line 1049 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIISD_79mimUJxdRey:-MOIISD_79mimUJxdRez:bm2uw1y) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Limiter.cpp#L1049)):*
> ```cpp
>   // limits under which compression is to be performed
>   auto al_eps = 1e-08;
>   auto loLim = 2.0 * al_eps;
> ```

Why do we need to have 2 times `al_eps` here?
___
*[src/PDE/Reconstruction.cpp, line 725 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIJkOc7xXlFskiMepu:-MOIJkOdCybB2JhrlGUu:bhlt2ly) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L725)):*
> ```cpp
>         state[energyIdx(nmat,k)] = alReco[k]
>           * U(e, energyDofIdx(nmat,k,rdof,0), offset)/alCC;
>         state[ncomp+pressureIdx(nmat,k)] = alReco[k]
> ```

What is this? Does it apply consistent limiting to pressure terms? I thought the consistent limiting is only for conserved variables.
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 210 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIVJ_72ARCXOyG0CZz:-MOIVJ_72ARCXOyG0C_-:b-76q3lt) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L210)):*
> ```cpp
>     //!   become unphysical and cause solution to diverge; thus requiring such
>     //!   a ``reset''.
>     void cleanTraceMaterial( const tk::Fields& geoElem,
> ```

Is the changes in the `cleanTraceMaterial` the reason why we need to update the regresion test for water-air shocktube problem?
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 280 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIWDwR-7pVscFx1urq:-MOIWDwR-7pVscFx1urr:b-t3fq1q) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L280)):*
> ```cpp
>         p_target /= ratio;
>         p_target = std::max(p_target, 1e-14);
>         p_target = std::max(pmax, 1e-14);
> ```

If we already calculate the `p_target` as equilibrium pressure. Why overwrite it equal to `pmax`. In this way, it seems that the calculation of equilibrium pressure is not necessary.
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 296 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIVvU8EkCLfe2xDuZK:-MOIVvU8EkCLfe2xDuZL:b2x21u1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L296)):*
> ```cpp
>             // of material present in majority. If both these conditions are
>             // true, perform pressure relaxation.
>             if ((alk < al_eps) || (pk+Pck < 0.0)/*&& (std::fabs((pk-pmax)/pmax) > 1e-08)*/)
> ```

Should we also added comments for why `pk+Pck < 0.0`?

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/462)* status: 43 of 44 files reviewed, 5 unresolved discussions (waiting on @jbakosi and @WeizhaoLi2018)

___
*[src/PDE/Limiter.cpp, line 1049 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIISD_79mimUJxdRey:-MOIcKGHDunnU3QfYgLi:b-o1bi0i) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Limiter.cpp#L1049)):*

Why do we need to have 2 times `al_eps` here?
</blockquote></details>

Good question. It's a convention. Take a look at some of the papers for two-phase flows with THINC by Xiao, or Kitamura for some more explanation.
___
*[src/PDE/Reconstruction.cpp, line 725 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIJkOc7xXlFskiMepu:-MOIc_sn1vFgeEGH-fNW:b-ylyjhy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/078de8022db2d436fc400751a6a42ae6430a0122/src/PDE/Reconstruction.cpp#L725)):*

What is this? Does it apply consistent limiting to pressure terms? I thought the consistent limiting is only for conserved variables.
</blockquote></details>

For interface reconstruction, be it geometric or algebraic, interface cells usually need to be first order in primitive variables (such as pressure) for stability reasons. I suggest referring to some of the papers where THINC is used in context of diffuse interface methods.
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 210 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIVJ_72ARCXOyG0CZz:-MOId0HXDVY8Vdh9aC5Q:b8pei5l) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L210)):*

Is the changes in the `cleanTraceMaterial` the reason why we need to update the regresion test for water-air shocktube problem?
</blockquote></details>

Yes, that's right.
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 280 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIWDwR-7pVscFx1urq:-MOId4GhDYwFLaCkSk7w:b-q1fwnn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L280)):*

If we already calculate the `p_target` as equilibrium pressure. Why overwrite it equal to `pmax`. In this way, it seems that the calculation of equilibrium pressure is not necessary.
</blockquote></details>

Good catch. It's not necessary. I have now commented out the p_equil calculations. But the commented part is kept in there for future use, when we incorporate other pressure-relaxation mechanisms and the target-pressure might need to be the equilibrium pressure.
___
*[src/PDE/MultiMat/DGMultiMat.hpp, line 296 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/462#-MOIVvU8EkCLfe2xDuZK:-MOIdQVZ5ELpzlFYNfEr:b8zrxig) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1816949be7c3548fb6180a2976ef56e7a6483614/src/PDE/MultiMat/DGMultiMat.hpp#L296)):*

Should we also added comments for why `pk+Pck < 0.0`?
</blockquote></details>

Good point. I'll update this comment.

PR #463, opened 2020-12-10
Add mesh transfer for ALECG

jbakosi:
This adds solution transfer between two meshes to the ALECG hydro scheme. Initial conditions and 2 meshes only for now. Appears to be correct, running the vortical flow test case, transferring the ICs for all scalars solved.

Also added hooks for solution transfer for every time step stage, see `ALECG::transfer()`, but when that's enabled, the cb is never called by `exam2m::setDestPoints()`. Next is to investigate why that is.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/463)


src/Inciter/Discretization.cpp
```diff
@@ -144,12 +133,40 @@ Discretization::Discretization(
...
+// *****************************************************************************
+// Our mesh has been registered with the mesh-to-mesh transfer library (if
+// coupled to other solver)
+// *****************************************************************************
+{
+  // Lambda to decide if a node is not counted by this chare. If a node is
+  // found in the node communication map and is associated to a lower chare id
+  // than thisIndex, it is counted by another chare (and not thisIndex), hence
+  // a "slave" (for the purpose of this count).
+  auto slave = [ this ]( std::size_t p ) {
```
epmikida, line 159: Interesting. Does this mean we won't be testing for duplicate collisions in ExaM2M? It's still probably a feature we should add to ExaM2M itself as well maybe? But if it works here that's good.

epmikida (approved):
This looks good to me. As for why iteration isn't working yet, I hadn't kept the ability to invoke the library multiple times in mind during development so there are probably a few small tweaks that need to be made. I'll look it over. It may also be a similar issue with the collision detection library used internally. Probably just an issue of resetting some state after the transfer is complete.

src/Inciter/Discretization.cpp
jbakosi: No, this is only used to sum the total number of mesh points without multi-counting those on the chare-boundary. We still pass the mesh connectivity and the node coordinates to ExaM2M in which the same physical points on chare-boundaries will show up multiple times bot on different chares.

adityakpandare (approved):
> Reviewed 45 of 65 files at r1, 2 of 19 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/463)* status: 47 of 77 files reviewed, 1 unresolved discussion (waiting on @epmikida)

PR #490, opened 2021-05-04
Fix segfault after merging in multi-box ICs, triggered by new shocktube run

jbakosi:
Fixes indexing out background pressure ICs are unspecified (and unused anyway).

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490)

### DIFF
--- a/src/Inciter/ALECG.cpp
+++ b/src/Inciter/ALECG.cpp
@@ -458,7 +458,7 @@ ALECG::box( tk::real v )
 
   // Set initial conditions for all PDEs
   for (auto& eq : g_cgpde)
-    eq.initialize( d->Coord(), m_u, d->T(), d->Boxvol(), m_boxnodes );
+    eq.initialize( d->Coord(), m_u, d->T(), v, m_boxnodes );
 
   // Multiply conserved variables with mesh volume
   volumetric( m_u );
--- a/src/PDE/CompFlow/CGCompFlow.hpp
+++ b/src/PDE/CompFlow/CGCompFlow.hpp
@@ -150,9 +150,7 @@ class CompFlow {
         (bgpreic.size() > m_system && !bgpreic[m_system].empty()) ?
         bgpreic[m_system][0] : 0.0;
 
-      const auto& cvol = g_inputdeck.get< tag::param, eq, tag::cv >();
-      tk::real cv = (cvol.size() > m_system && !cvol[m_system].empty()) ?
-                    cvol[m_system][0] : 0.0;
+      auto cv = g_inputdeck.get< tag::param, eq, tag::cv >()[ m_system ][ 0 ];
 
       // Set initial and boundary conditions using problem policy
       for (ncomp_t i=0; i<x.size(); ++i) {
--- a/src/PDE/CompFlow/CGCompFlow.hpp
+++ b/src/PDE/CompFlow/CGCompFlow.hpp
@@ -145,9 +145,14 @@ class CompFlow {
 
       const auto eps = 1000.0 * std::numeric_limits< tk::real >::epsilon();
 
-      // if an ic box was not specified, avoid division by zero by setting V
       const auto& bgpreic = ic.get< tag::pressure >();
-      const auto& cv = g_inputdeck.get< tag::param, eq, tag::cv >();
+      tk::real bgpre =
+        (bgpreic.size() > m_system && !bgpreic[m_system].empty()) ?
+        bgpreic[m_system][0] : 0.0;
+
+      const auto& cvol = g_inputdeck.get< tag::param, eq, tag::cv >();
+      tk::real cv = (cvol.size() > m_system && !cvol[m_system].empty()) ?
+                    cvol[m_system][0] : 0.0;
 
       // Set initial and boundary conditions using problem policy
       for (ncomp_t i=0; i<x.size(); ++i) {
@@ -165,8 +170,7 @@ class CompFlow {
                 b.template get< tag::zmin >(), b.template get< tag::zmax >() };
               auto V_ex = (box[1]-box[0]) * (box[3]-box[2]) * (box[5]-box[4]);
               if (V_ex < eps) V = 1.0;
-              initializeBox( m_system, V_ex/V, t, b, bgpreic[m_system][0],
-                             cv[m_system][0], s );
+              initializeBox( m_system, V_ex/V, t, b, bgpre, cv, s );
             }
             ++bcnt;
           }


adityakpandare (changes requested):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 155 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsHU5n57TrGIXnxV3I:-MZsHU5n57TrGIXnxV3J:b7w4f9w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L155)):*
> ```cpp
>       const auto& cvol = g_inputdeck.get< tag::param, eq, tag::cv >();
>       tk::real cv = (cvol.size() > m_system && !cvol[m_system].empty()) ?
>                     cvol[m_system][0] : 0.0;
> ```

I don't understand this particular change. `cv` is the specific heat at constant volume, the defaults for which are set in Grammar, `check_compflow()`. We shouldn't have to set this to zero here. Also, I suggest keeping the variable name `cv`, instead of `cvol`: it's a pretty standard thermodynamic symbol for the specific heat at const volume.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 173 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsIBxOE4kJ0A5HPKXH:-MZsIBxOE4kJ0A5HPKXI:b-2job8r) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L173)):*
> ```cpp
>               auto V_ex = (box[1]-box[0]) * (box[3]-box[2]) * (box[5]-box[4]);
>               if (V_ex < eps) V = 1.0;
>               initializeBox( m_system, V_ex/V, t, b, bgpre, cv, s );
> ```

Is there a reason that we need to pass vectors rather than scalar values of pressure and specific heat to `initializeBox()`? I think we should pass scalars; that way the `initializeBox` function for this system of equations (compflow) does not have to know about it being a vector, which it shouldn't.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 155 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsHU5n57TrGIXnxV3I:-MZsmGPq4nXNA5UWQY_E:bi6deey) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L155)):*

I don't understand this particular change. `cv` is the specific heat at constant volume, the defaults for which are set in Grammar, `check_compflow()`. We shouldn't have to set this to zero here. Also, I suggest keeping the variable name `cv`, instead of `cvol`: it's a pretty standard thermodynamic symbol for the specific heat at const volume.
</blockquote></details>

I just used `cvol` to make the vector of vectors in the input deck have a different name compared to the value of `cv` that is actually used.

The reason I have changed `cv` as well as `bpgressure` is because they both come from a vector of vectors. I think in the grammar we have done a different kind of initialization and/or error checking for these, but they are a both read out of a vector of vectors and if these vectors are not initialized (by the user or by assigning some defaults), they will fail for the same reason with an ugly segfault, which I wanted to avoid. I think we could probably do a little better error checking and/or initialization for both, but I wanted a quick fix so this does not segfault at least.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 173 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsIBxOE4kJ0A5HPKXH:-MZsovL42WbWlCG2X3JR:b-d21c2x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L173)):*

Is there a reason that we need to pass vectors rather than scalar values of pressure and specific heat to `initializeBox()`? I think we should pass scalars; that way the `initializeBox` function for this system of equations (compflow) does not have to know about it being a vector, which it shouldn't.
</blockquote></details>

We are passing scalars, both previous to this change and now.

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 155 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsHU5n57TrGIXnxV3I:-MZsqEck5r1rjDGjZfoT:b-1pja5d) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L155)):*

I just used `cvol` to make the vector of vectors in the input deck have a different name compared to the value of `cv` that is actually used.

The reason I have changed `cv` as well as `bpgressure` is because they both come from a vector of vectors. I think in the grammar we have done a different kind of initialization and/or error checking for these, but they are a both read out of a vector of vectors and if these vectors are not initialized (by the user or by assigning some defaults), they will fail for the same reason with an ugly segfault, which I wanted to avoid. I think we could probably do a little better error checking and/or initialization for both, but I wanted a quick fix so this does not segfault at least.
</blockquote></details>

I understand the `bgpressure` change. But I don't understand it with `cv`. We have set the cv value to the cv of air if left unspecified by the user, in `check_compflow()`. I don't get why it would be necessary to have this 0.0 default here.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 173 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsIBxOE4kJ0A5HPKXH:-MZspUzOBnaRTXDHJZv1:b-emg1y4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L173)):*

We are passing scalars, both previous to this change and now.
</blockquote></details>

Oh, yeah, I missed that. Thanks for the explanation.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 155 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsHU5n57TrGIXnxV3I:-MZsvaG41K7udowWoa9l:b-5eyt1v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L155)):*

I understand the `bgpressure` change. But I don't understand it with `cv`. We have set the cv value to the cv of air if left unspecified by the user, in `check_compflow()`. I don't get why it would be necessary to have this 0.0 default here.
</blockquote></details>

You are right, I will revert that. I was just trying to use my hammer on everything I found in the neighborhood that look like a nail. ;-)

adityakpandare (approved):
Thanks for fixing this! :lgtm:

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-:-MZszuo79vSQet731YbK:b-8jr5f1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 155 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/490#-MZsHU5n57TrGIXnxV3I:-MZszcjhEMKJRmJJZ8kC:b-9y6e2j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ecb9548e18f58db3314aa827c7e744a3ca220421/src/PDE/CompFlow/CGCompFlow.hpp#L155)):*

You are right, I will revert that. I was just trying to use my hammer on everything I found in the neighborhood that look like a nail. ;-)
</blockquote></details>

:-D Looks good now.

PR #491, opened 2021-05-07
Dg nodal comm

WeizhaoLi2018:
This PR removes the extension of esup vector and add the new communication for nodal extremes used for vertex-based limiter

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)


WeizhaoLi2018 (commented):
Hi guys, this is the PR that will remove the extended part of `esup` vector and make sure the extreme values for conservative/primitive variables are communicated with chare boundary nodes. This PR includes the communication of nodal max/min cell average value used for P1 limiter. After we merge this, I will rebase the branch for P2 limiter and we will start to merge that one next.

The regression test for multi-material sod shocktube problem do pass which means that when adding this communication, we can still maintain our previous results.

However, even though the regression test pass on my local machine, there is still one issue when I tried to test it with more cores and more iterations. The problem is about the initialization of `m_nodalmaxc` and `m_nodalminc`. Since there will be max/min function operator, so I initialize both unordered maps in `DG::reco()` to make sure when we access its data on `DG::comnodal()`, it's already allocated. But I found that sometimes, when `DG::comnodal()` calls, the size of `m_nodalmaxc` is still zero and return as segmentation fault. So I was wondering for `nodalmaxc` and `nodalminc`, how can I initialize properly (this initialization should happen at each stage of our simulation). 

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_6nGpE4_twExSyefBL:bqt21of)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @adityakpandare and @jbakosi)

adityakpandare (changes requested):
@weizhaoli2018, the nodal-neighbor least squares reconstruction `recoLeastSqExtStencil()` uses the extended `esup`. I do not think you have updated that function to work with the new way to communicate nodal neighborhood. `recoLeastSqExtStencil()` is used routinely to get good material interface gradient estimates, which will not work in parallel because of your change. You should not remove the extensions to `esup` yet. Please reinstate the `esup` extensions so that `recoLeastSqExtStencil()` does not break in parallel. I will take a look at your changes after that. Thanks!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_Leyqz7dcglSJctkac:b-xs8gzn)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @jbakosi)

WeizhaoLi2018 (commented):
@adityakpandare I will bring back the extension of `esup`. But if we have remained the extension of `esup`, it seems that we don't need this vertex-based communication. Because we do use the loop as `for (auto er : pesup)` in Limiter.cpp to find the extreme values for nodes.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_M28Z97uhdp9sjdp38:bamrfom)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @jbakosi)

WeizhaoLi2018 (commented):
Sorry for my confusion of last reply. I should add `if` statement in the loop `for (auto er : pesup)` with the extension of `esup` to make sure limiter function only access data for internal elements and this is important for DGP2 limiter. While other parts like reconstruction could still use the extension of `esup`. Thanks!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_MD3-H-GvGX9SbJdq2:b-apsyvc)* status: 0 of 17 files reviewed, all discussions resolved (waiting on @jbakosi)

jbakosi (commented):
Hey @weizhaoli2018! On the initialization of `m_nodalmaxc` and `m_nodalminc`: There is no global sync during time stepping except calling `solve()` at the bottom of `dt()`:

```c++
  // Contribute to minimum dt across all chares then advance to next step
  contribute( sizeof(tk::real), &mindt, CkReduction::min_double,
              CkCallback(CkReductionTarget(DG,solve), thisProxy) );
```

Because of this, we preactically allow the runtime system to schedule any function call arbitrarily that adheres to the DAGs in `dg.ci` between two calls to `solve()` (in two consecutive time steps). As a consequence, initializing `m_nodalmaxc` and `m_nodalminc` in `reco()` in some chare may happen as late as during (or after) `comnodal()`, which already wants to write to these, even though `comnodal()` is executed only when both `ownreco` and `comreco` are complete on a chare, but which may nevertheless be called earlier than `reco()`. Don't be misled by the _location of the code_ of the the initialization of `m_nodalmaxc` and `m_nodalminc` being _after_ calling `comreco()`. Since `comreco()` is an asynchronous function call, its execution may be delayed by the runtime system, as far as the same chare already executing `comnodal()`, which can be called by some other chare (sending some data), even before `m_nodalmaxc` and `m_nodalminc` were even initialized in `reco()`. (Writing the code to call `comreco()` does not at all mean that the function will be executed where the call is written.)

To ensure that`m_nodalmaxc` and `m_nodalminc` get properly initialized, you could initialize them in the DG constructor, since `Disc()->NodeCommMap()` is already available for their initialization. Running

```c++
  // Initialization for the buffer vector of nodal extremes
  for (const auto& [c,n] : Disc()->NodeCommMap())
  {
    auto gid = std::vector<std::size_t>(std::begin(n),std::end(n));
    for (std::size_t i=0; i<gid.size(); ++i)
    {
      m_nodalmaxc[gid[i]].resize(ncomp+nprim,-std::numeric_limits< tk::real >::max());
      m_nodalminc[gid[i]].resize(ncomp+nprim, std::numeric_limits< tk::real >::max());
    }
  }
```
in the DG constructor should be sufficient for using `m_nodalmaxc` and `m_nodalminc` the first time, but not in subsequent time steps. To make sure that they are always properly initialized for the next time step, you could initialize them sometime _before_ issuing the reduction to `solve()` in `dt()`:

```c++
  // initialize m_nodalmaxc and m_nodalminc somewhere here

  // Contribute to minimum dt across all chares then advance to next step
  contribute( sizeof(tk::real), &mindt, CkReduction::min_double,
              CkCallback(CkReductionTarget(DG,solve), thisProxy) );
```

Since this solution calls the initialization from two places, I would put it into a function and just call it from two places, that way we don't repeat code. Since `dt()` and `solve()` happen towards the very end of a time step, this should ensure that `m_nodalmaxc` and `m_nodalminc` are initialized _before_ the next time step, since it is behind the "synchronization wall of" `solve()`.

@adityakpandare , do you agree?

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_WZyuP-O5_lLwwaNZ2:bromvi6)* status: 0 of 17 files reviewed, all discussions resolved

adityakpandare (commented):
@weizhaoli2018, yes, that if-test inside the p2-limiter sounds like what you should be doing, so that those `esup`-extensions are not used in the limiter, but still used elsewhere.

@jbakosi, I agree with you. `m_nodalmaxc` (and min) could be uninitialized when `comnodal()` is already trying to write to them. I also agree with the solution you suggest.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-M_WkRaRC0YdPkXSeT3-:bxqhsmn)* status: 0 of 17 files reviewed, all discussions resolved

adityakpandare (changes requested):
Thanks @weizhaoli2018. I have a couple of comments below. Could you also take a look at why the tests on Azure are failing for this branch?

> Reviewed 5 of 17 files at r1, 12 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma-xybR7Cp5Et7_lHmv:b-p24214)* status: all files reviewed, 2 unresolved discussions (waiting on @WeizhaoLi2018)

___
*[src/Inciter/dg.ci, line 41 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-M_unU7CBI2Pw9w9kfdK:-M_unU7CBI2Pw9w9kfdL:b38m6rw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e1e4376d7070cedb7269a682f53df78d2050f524/src/Inciter/dg.ci#L41)):*
> ```
>         const std::unordered_map< std::size_t, std::vector< tk::real > >&
>           nodeBoundaryCells );
>       //entry void adj();
> ```

Why do we have this commented code?
___
*[src/PDE/Limiter.cpp, line 878 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma-rhTi0I6YK0YwglHN:-Ma-rhTi0I6YK0YwglHO:b-hmkgrc) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e1e4376d7070cedb7269a682f53df78d2050f524/src/PDE/Limiter.cpp#L878)):*
> ```cpp
> //! \param[in] indicator Indicator marking for which variables the limiting
> //!   function is applied. If 0, apply the limiter to conservative variables,
> //!   otherwise, apply the limiter to primitive variables
> ```

I do not like the use of this `indicator` parameter, which is required to index into the `nodalmax` and `nodalmin` vectors. I prefer the fact that the limiter function has no knowledge about whether it is limiting primitive or conserved quantities, especially because the function does not need to do anything different based on this knowledge. I can think of two options to deal with this:
1. `nodalmax` can be split into `uNodalMax` and `pNodalMax` (and the same for `nodalmin`), so that the indicator argument will not be needed here. This is also consistent with the solution and primitive variable vectors `U` and `P`, and as such, is a cleaner solution. In fact, you can merge `uNodalMax` and `uNodalMin` into a single vector `uNodalExtrm`, with two values per scalar component: min and max, and access them as `uNodalExtrm(e,c,0)` and `uNodalExtrm(e,ncomp+c,0)` respectively. The same for `pNodalMax` and `pNodalMin`.
2. Another option is keeping the code as it is, but specifying a default for `indicator` (can be done in function declaration as `const std::size_t indicator=0`), so that client code for conserved variable limiting will not change, and can leave this argument unspecified. This is a quick fix, but in my opinion not an elegant one, basically because using this option, the limiter function still has knowledge about whether it is operating on primitive or conserved quantities.

I suggest the first option.

WeizhaoLi2018 (commented):
Hi guys, I found out why the regression tests fail. 
It turns out that during the test of `amr_dtref_trans_gauss_hump_dg` the code in my sending function:

```Javascript
for (const auto& [c,n] : d->NodeCommMap()) 
```

the elements in the shared node vector `n` here becomes some invalid huge numbers like `12128069337997769099`. This happens at 5th time step which is exactly where amr is applied.
@jbakosi do you have any idea about this? Is this related to amr procedure?

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4EL4c5zlKGDGqAlKY:b-213gir)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/dg.ci, line 41 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-M_unU7CBI2Pw9w9kfdK:-Ma40CEv0hW75nyS8c4f:b-e3v0zi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e1e4376d7070cedb7269a682f53df78d2050f524/src/Inciter/dg.ci#L41)):*

Why do we have this commented code?
</blockquote></details>

Oh, I introduce this declaration for the version without the extension of esup. Once we bring back that, this declaration does not need to be here. I should delete. Will do that.

jbakosi (commented):
The huge numbers are very likely valid and correct node ids. They are introduced by AMR when a new node is added to the mid-point of an edge due to refinement. New node id are generated by a hashing procedure, which yields the same new (large) integer value based on the old (edge-end-point) node ids, guaranteed to match across chares potentially residing on different CPUs. This eliminates data that need to be communicated, since the hashes (the new node ids)  are cryptographically, i.e., very highly likely, the same on different chares.

Note that, AFAIK, DG is not yet tested with AMR, so you might be bumping into a use case here that is not even supposed to work yet anyway.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4I-_9CGemVNBBsKXl:bbllifu)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

adityakpandare (commented):
I'm also taking a look at this now. I'll let you know if I find anything.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4MB5_DHiL3MX4rHhc:blye0rs)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

WeizhaoLi2018 (commented):
If this was caused by the new node ids. I could try to add `if(rdof > 1)` for my receiving part to see whether this could avoid this segmentation fault in `amr_dg` regression test.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4ObvkDVEkqs0VPnsF:bm78b72)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

adityakpandare (commented):
The `m_nodalmax` and `m_nodalmin` vector were not extended in `DG::resizePostAMR()`. I just pushed a fix for that. There is something else that's amiss. I'm looking at it.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4QTM_5JzD7cjYeK6X:b-9troh5)* status: 16 of 17 files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

adityakpandare (commented):
The other problem was that `m_nodalminc` and `m_nodalmaxc` have not been extended in `resizePostAMR()`. Just fixed that. That should fix the regression tests.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4aSRX1jVOGXZnSJBe:b-4q8gml)* status: 16 of 17 files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: 16 of 17 files reviewed, 3 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/DG.cpp, line 1805 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4akFt7iLcgdw6usqw:-Ma4akFt7iLcgdw6usqx:b-fi3po) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L1805)):*
> ```cpp
> 
> void
> DG::nodal()
> ```

There's many functions that do communication in DG now, making the class pretty complicated. Because of this, let's have more meaningful function names. I suggest `nodalExtrema()`. Similarly, `comnodal()` can be `comNodalExtrema()`

WeizhaoLi2018 (commented):
@adityakpandare Thank you! I just tested on my machine. All the about `amr_dtref_gauss_hump_dg` pass.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Ma4dHBe9zLEcvICDpkU:b50tone)* status: 16 of 17 files reviewed, 3 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

jbakosi (changes requested):
> Reviewed 5 of 17 files at r1, 5 of 12 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: all files reviewed, 7 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/DG.cpp, line 93 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4zne79UbY3dr7gmAe:-Ma4zne79UbY3dr7gmAf:b-2l1uk1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L93)):*
> ```cpp
>   m_nodalmin( Disc()->Bid().size(),
>               g_inputdeck.get< tag::component >().nprop() +
>               m_p.nprop() / g_inputdeck.get< tag::discr, tag::rdof >() ),
> ```

Shorter versions:
```
m_nodalmin( m_nodalmax.nunk(), m_nodalmax.nprop() ),
```
or
```
m_nodalmin( m_nodalmax ),
```
___
*[src/Inciter/DG.cpp, line 151 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma51cfS1wtiH1Qe6gIa:-Ma51cfS1wtiH1Qe6gIb:bpaadrk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L151)):*
> ```cpp
>   {
>     auto gid = std::vector<std::size_t>(std::begin(n),std::end(n));
>     for (std::size_t i=0; i<gid.size(); ++i)
> ```

This could also be
```c++
for (auto i : n) {
  m_nodalmaxc[i].resize(ncomp+nprim,-std::numeric_limits< tk::real >::max());
  m_nodalminc[i].resize(ncomp+nprim, std::numeric_limits< tk::real >::max());
}
```
or, I would write
```c++
auto large = std::numeric_limits< tk::real >::max();
for (auto i : n) {
  m_nodalmaxc[i].resize( ncomp+nprim, -large );
  m_nodalminc[i].resize( ncomp+nprim,  large );
}
```
which is simpler code and also eliminates constructing vector, `gid`, from set, `n`, and I don't think it will be any slower.
___
*[src/Inciter/DG.cpp, line 154 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma516VO2oToLeWo3owJ:-Ma516VPAW9D6m-lvdmh:b-snwjpi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L154)):*
> ```cpp
>     {
>       m_nodalmaxc[gid[i]].resize(ncomp+nprim,-std::numeric_limits< tk::real >::max());
>       m_nodalminc[gid[i]].resize(ncomp+nprim, std::numeric_limits< tk::real >::max());
> ```

Let's keep the max column width 80.
___
*[src/Inciter/DG.cpp, line 2137 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma52dVM5VSAxm-lSaWW:-Ma52dVM5VSAxm-lSaWX:b-q9xumu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L2137)):*
> ```cpp
>       m_nodalminc[gid[i]].resize(ncomp+nprim, std::numeric_limits< tk::real >::max());
>     }
>   }
> ```

This code is now repeated from the constructor and also appears in `resizePostAMR()`. Can you pull this out to a function and just call it whenever it is needed?
___
*[src/PDE/Limiter.cpp, line 878 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma-rhTi0I6YK0YwglHN:-Ma57dbSCzgs8LibmM3n:b1q6px9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e1e4376d7070cedb7269a682f53df78d2050f524/src/PDE/Limiter.cpp#L878)):*

I do not like the use of this `indicator` parameter, which is required to index into the `nodalmax` and `nodalmin` vectors. I prefer the fact that the limiter function has no knowledge about whether it is limiting primitive or conserved quantities, especially because the function does not need to do anything different based on this knowledge. I can think of two options to deal with this:
1. `nodalmax` can be split into `uNodalMax` and `pNodalMax` (and the same for `nodalmin`), so that the indicator argument will not be needed here. This is also consistent with the solution and primitive variable vectors `U` and `P`, and as such, is a cleaner solution. In fact, you can merge `uNodalMax` and `uNodalMin` into a single vector `uNodalExtrm`, with two values per scalar component: min and max, and access them as `uNodalExtrm(e,c,0)` and `uNodalExtrm(e,ncomp+c,0)` respectively. The same for `pNodalMax` and `pNodalMin`.
2. Another option is keeping the code as it is, but specifying a default for `indicator` (can be done in function declaration as `const std::size_t indicator=0`), so that client code for conserved variable limiting will not change, and can leave this argument unspecified. This is a quick fix, but in my opinion not an elegant one, basically because using this option, the limiter function still has knowledge about whether it is operating on primitive or conserved quantities.

I suggest the first option.
</blockquote></details>

I agree and vote for Option 1 using a single data structure for both max and min.

WeizhaoLi2018 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: 8 of 17 files reviewed, 7 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/DG.cpp, line 93 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4zne79UbY3dr7gmAe:-MaAREyK3JM404nwUsw1:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L93)):*

Shorter versions:
```
m_nodalmin( m_nodalmax.nunk(), m_nodalmax.nprop() ),
```
or
```
m_nodalmin( m_nodalmax ),
```
</blockquote></details>

Done. Thanks!
___
*[src/Inciter/DG.cpp, line 151 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma51cfS1wtiH1Qe6gIa:-MaARUi603p2v9ASVD7d:b-7pdx6d) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L151)):*

This could also be
```c++
for (auto i : n) {
  m_nodalmaxc[i].resize(ncomp+nprim,-std::numeric_limits< tk::real >::max());
  m_nodalminc[i].resize(ncomp+nprim, std::numeric_limits< tk::real >::max());
}
```
or, I would write
```c++
auto large = std::numeric_limits< tk::real >::max();
for (auto i : n) {
  m_nodalmaxc[i].resize( ncomp+nprim, -large );
  m_nodalminc[i].resize( ncomp+nprim,  large );
}
```
which is simpler code and also eliminates constructing vector, `gid`, from set, `n`, and I don't think it will be any slower.
</blockquote></details>

Done. Thanks!
___
*[src/Inciter/DG.cpp, line 154 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma516VO2oToLeWo3owJ:-MaARcBE67dRnzgUfMC7:bt2vr0y) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L154)):*

Let's keep the max column width 80.
</blockquote></details>

Fixed!
___
*[src/Inciter/DG.cpp, line 1805 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4akFt7iLcgdw6usqw:-MaAX8qk6rnRdWrEPM5i:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L1805)):*

There's many functions that do communication in DG now, making the class pretty complicated. Because of this, let's have more meaningful function names. I suggest `nodalExtrema()`. Similarly, `comnodal()` can be `comNodalExtrema()`
</blockquote></details>

Done. Thanks!
___
*[src/Inciter/DG.cpp, line 2137 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma52dVM5VSAxm-lSaWW:-MaAXC728Kb63chEkvaf:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L2137)):*

This code is now repeated from the constructor and also appears in `resizePostAMR()`. Can you pull this out to a function and just call it whenever it is needed?
</blockquote></details>

Done. Thanks!
___
*[src/PDE/Limiter.cpp, line 878 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma-rhTi0I6YK0YwglHN:-MaAfpfI8E5ab3HSkuAE:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e1e4376d7070cedb7269a682f53df78d2050f524/src/PDE/Limiter.cpp#L878)):*

I agree and vote for Option 1 using a single data structure for both max and min.
</blockquote></details>

Done. Thanks!

adityakpandare (changes requested):
Thanks for the hard work with this @weizhaoli2018. I have a few more comments. Should be good to go from my side after these are addressed.

> Reviewed 11 of 11 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-MaAvEtdCW0lctWnhFHk:baceet7)* status: all files reviewed, 10 unresolved discussions (waiting on @jbakosi and @WeizhaoLi2018)

___
*[src/Inciter/DG.hpp, line 313 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAtWyBC2LfQhFyv-Ut:-MaAtWyBC2LfQhFyv-Uu:brnk7vs) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/Inciter/DG.hpp#L313)):*
> ```cpp
>     //! Counter signaling that we have received all our reconstructed ghost data
>     std::size_t m_nreco;
>     //! Counter signaling that we have received all our nodal extremes from
> ```

"nodal extremas"
___
*[src/Inciter/DG.cpp, line 1805 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4akFt7iLcgdw6usqw:-MaAt-zaELACQN3X2iXh:bvdznbx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L1805)):*

Done. Thanks!
</blockquote></details>

"Extrema" has a specific mathematical meaning (see https://en.wikipedia.org/wiki/Maxima_and_minima, it is a plural of "Extremum"). I recommend you make this and other relevant function names (and in comments) "Extrema" instead of "Extreme" or "Extremes".
___
*[src/Inciter/dg.ci, line 99 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAteyU-P-JEpS7GESj:-MaAteyU-P-JEpS7GESk:b3x5n7o) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/Inciter/dg.ci#L99)):*
> ```
> 
>       entry void wait4nodalExtreme() {
>         when ownnodalExtreme_complete(), comnodalExtreme_complete()
> ```

make all relevant function names "extrema", not "extreme"
___
*[src/PDE/Limiter.cpp, line 312 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAqxnq1ljx_aZ37Wc2:-MaAqxnq1ljx_aZ37Wc3:b-8219j2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L312)):*
> ```cpp
>   const std::unordered_map< std::size_t, std::size_t >& bid,
>   tk::Fields& uNodalExtrm,
>   tk::Fields&,
> ```

If we are not using this argument for `pNodalExtrm`, which we do not for CompFlow and Transport, we should remove it from the function altogether.
___
*[src/PDE/Limiter.cpp, line 401 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAr62m6CBg4cDQBMsQ:-MaAr62m6CBg4cDQBMsR:bovq06y) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L401)):*
> ```cpp
>   const std::unordered_map< std::size_t, std::size_t >& bid,
>   tk::Fields& uNodalExtrm,
>   tk::Fields&,
> ```

Same here, remove this argument if it's not used.
___
*[src/PDE/Limiter.cpp, line 917 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaArpk36JwPSjTU-RTM:-MaArpk36JwPSjTU-RTN:brz1uri) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L917)):*
> ```cpp
>     auto p = inpoel[4*e+lp];
>     const auto& pesup = tk::cref_find(esup, p);
> ```

The comment for the chunk of code below this "Step-1: find min/max in the neighborhood" should be moved here, since this code also does the same, but at chare boundaries. Further, let's add a one-two line comment here explaining what this new chunk of code does, i.e. min/max at partition boundaries. As I have mentioned previously, please remember to add comments to code that you add, it makes understanding easier when we return to the code in the future.
___
*[src/PDE/Limiter.cpp, line 924 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaArLxM3GfTtlyZ93gI:-MaArLxM3GfTtlyZ93gJ:bi3xsy0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L924)):*
> ```cpp
>       {
>         uMax[c] = std::max(NodalExtrm(gip->second,c,0),       uMax[c]);
>         uMin[c] = std::min(NodalExtrm(gip->second,c+ncomp,0), uMin[c]);
> ```

Excellent! Thanks!

WeizhaoLi2018 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: 9 of 17 files reviewed, 10 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/DG.hpp, line 313 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAtWyBC2LfQhFyv-Ut:-MaAziCV-JyzIh9AxukS:b-1jl013) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/Inciter/DG.hpp#L313)):*

"nodal extremas"
</blockquote></details>

Fixed. Thank you!
___
*[src/Inciter/DG.cpp, line 1805 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4akFt7iLcgdw6usqw:-MaAznE1ALBO0Fs2mMrs:bz1ydms) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L1805)):*

"Extrema" has a specific mathematical meaning (see https://en.wikipedia.org/wiki/Maxima_and_minima, it is a plural of "Extremum"). I recommend you make this and other relevant function names (and in comments) "Extrema" instead of "Extreme" or "Extremes".
</blockquote></details>

Got it. I have replaced all the "Extremes" into "Extrema".
Thank you!
___
*[src/Inciter/dg.ci, line 99 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAteyU-P-JEpS7GESj:-MaAyo5ICjHihipQhiK_:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/Inciter/dg.ci#L99)):*

make all relevant function names "extrema", not "extreme"
</blockquote></details>

Done. Thanks!
___
*[src/PDE/Limiter.cpp, line 312 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAqxnq1ljx_aZ37Wc2:-MaAzaC2EG7QOvMJaXa1:boyexb3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L312)):*

If we are not using this argument for `pNodalExtrm`, which we do not for CompFlow and Transport, we should remove it from the function altogether.
</blockquote></details>

Done. Thanks!
___
*[src/PDE/Limiter.cpp, line 401 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAr62m6CBg4cDQBMsQ:-MaAzdO26R4R4cof_9oB:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L401)):*

Same here, remove this argument if it's not used.
</blockquote></details>

Done.
___
*[src/PDE/Limiter.cpp, line 917 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaArpk36JwPSjTU-RTM:-MaB1S4pDLM-9nNEQsr_:b-wj9w88) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L917)):*

The comment for the chunk of code below this "Step-1: find min/max in the neighborhood" should be moved here, since this code also does the same, but at chare boundaries. Further, let's add a one-two line comment here explaining what this new chunk of code does, i.e. min/max at partition boundaries. As I have mentioned previously, please remember to add comments to code that you add, it makes understanding easier when we return to the code in the future.
</blockquote></details>

I just move this chunk of code next to the loop of internal elements and add a comment.

adityakpandare (approved):
:lgtm:

> Reviewed 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-MaBBNwj-ici_ozCa0Jp:bnfp4nl)* status: all files reviewed, 4 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/DG.hpp, line 313 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaAtWyBC2LfQhFyv-Ut:-MaBASWr16Z9N0bJ3w4U:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/Inciter/DG.hpp#L313)):*

Fixed. Thank you!
</blockquote></details>

Thanks!
___
*[src/Inciter/DG.cpp, line 1805 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma4akFt7iLcgdw6usqw:-MaBAkWo36sbs0NBjIhF:b-ec358g) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L1805)):*

Got it. I have replaced all the "Extremes" into "Extrema".
Thank you!
</blockquote></details>

Great! Thanks for doing this.
___
*[src/PDE/Limiter.cpp, line 917 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaArpk36JwPSjTU-RTM:-MaBBDgG0F2L0paM4OO7:b-jhyyta) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b488782eae8841e6e01b0a5acb50a591fc2ca0d1/src/PDE/Limiter.cpp#L917)):*

I just move this chunk of code next to the loop of internal elements and add a comment.
</blockquote></details>

Got it. Thanks!

jbakosi (changes requested):
Looks pretty good. I still have some minor comments, that we should address before merging. Thanks.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-MaiYCxZ9cNqS_4ifwSQ:b-lgkt5v)* status: all files reviewed, 10 unresolved discussions (waiting on @WeizhaoLi2018)

___
*[src/Inciter/DG.hpp, line 314 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiNWoz9Yzwf7PWNAjY:-MaiNWoz9Yzwf7PWNAjZ:b-td7i3x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.hpp#L314)):*
> ```cpp
>     std::size_t m_nreco;
>     //! Counter signaling that we have received all our nodal extrema from
>     //! ghost chare partitions
> ```

The doxygen comment in two line will only show up correctly on the web/doc if it is prefixed by `\brief`. (If it is one-line, it is okay without it.)
```c++
//! \brief Counter signaling that we have received all our nodal extrema from
//!       ghost chare partitions
```
___
*[src/Inciter/DG.hpp, line 340 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiR20PC3uvVAIVLd0n:-MaiR20PC3uvVAIVLd0o:bote894) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.hpp#L340)):*
> ```cpp
>     //! Vector of nodal extrema
>     tk::Fields m_uNodalExtrm;
>     tk::Fields m_pNodalExtrm;
> ```

The second one will not have a comment/explanation in the doxygen doc. Let's add something!
___
*[src/Inciter/DG.cpp, line 154 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-Ma516VO2oToLeWo3owJ:-MaiP1rX0mL0P2H73U0t:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4982f9ce4550e581cd88b0fc740927b5ea26f2ce/src/Inciter/DG.cpp#L154)):*

Fixed!
</blockquote></details>

Thanks!
___
*[src/Inciter/DG.cpp, line 142 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiP8AV1RrMGUqGam18:-MaiP8AW6OD0VWUl6K9n:b8dc4fj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L142)):*
> ```cpp
> 
>   // Initialization for the buffer vector of nodal extrema
>   ResizeNodalExtremac(Disc()->NodeCommMap(), m_uNodalExtrmc, m_pNodalExtrmc);
> ```

There is no need to pass any of the arguments, since they are all accessible from the function body. And when this is called from `DG::resizePostAMR()`, the node communication map in `Discretization` will have already been updated.
___
*[src/Inciter/DG.cpp, line 142 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiPMs44Ivx4rfh59ln:-MaiPMs44Ivx4rfh59lo:btqhvn0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L142)):*
> ```cpp
> 
>   // Initialization for the buffer vector of nodal extrema
>   ResizeNodalExtremac(Disc()->NodeCommMap(), m_uNodalExtrmc, m_pNodalExtrmc);
> ```

Let's keep the convention of camel-case for function names: `resizeNodalExtrema`. (Types start with uppercase.)
___
*[src/Inciter/DG.cpp, line 1877 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQeWh567SIvGumdRP:-MaiQeWh567SIvGumdRQ:bjkh3zr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1877)):*
> ```cpp
>     for (const auto& [c,n] : d->NodeCommMap()) {
>       std::vector< std::vector< tk::real > > g1( n.size() );
>       std::vector< std::vector< tk::real > > g2( n.size() );
> ```

Let's write it shorter:
```c++
std::vector< std::vector< tk::real > > g1( n.size() ), g2( n.size() );
```
___
*[src/Inciter/DG.cpp, line 1882 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQXvH0OgeP_jPBopo:-MaiQXvH0OgeP_jPBopp:bs7s8l0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1882)):*
> ```cpp
>       {
>         g1[ j   ] = m_uNodalExtrm[ tk::cref_find(d->Bid(),i) ];
>         g2[ j++ ] = m_pNodalExtrm[ tk::cref_find(d->Bid(),i) ];
> ```

Let's not perform the find twice.
___
*[src/Inciter/DG.cpp, line 1884 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQQTJ7GSn1hyqo-yL:-MaiQQTJ7GSn1hyqo-yM:b5gcpd) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1884)):*
> ```cpp
>         g2[ j++ ] = m_pNodalExtrm[ tk::cref_find(d->Bid(),i) ];
>       }
>       thisProxy[c].comnodalExtrema( std::vector<std::size_t>(begin(n),end(n)), g1, g2 );
> ```

Let's keep to 80 columns.
___
*[src/Inciter/DG.cpp, line 1923 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiRKty8Js2x0w96NsO:-MaiRKty8Js2x0w96NsP:b-zq4s3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1923)):*
> ```cpp
>       m_uNodalExtrmc[gid[i]][c+ncomp] =
>         std::min( G1[i][c+ncomp], m_uNodalExtrmc[gid[i]][c+ncomp] );
>     }
> ```

The above performs a _search_ behind `m_uNodalExtrmc[gid[i]]` `4*c` times(!) yielding the same result. Let's do that only once:
```c++
  for (std::size_t i=0; i<gid.size(); ++i) {
    auto& n = m_uNodalExtrmc[gid[i]];
    for (std::size_t c=0; c<ncomp; ++c) {
      n[c] = std::max( G1[i][c], n[c] );
      n[c+ncomp] = std::min( G1[i][c+ncomp], n[c+ncomp] );
    }
  }
```
Same below for the primitive variables.

(Sorry, I think I missed this earlier.)
___
*[src/Inciter/DG.cpp, line 1967 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiWLcK8Iqr9mASnj4x:-MaiWLcK8Iqr9mASnj4y:bnv68gr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1967)):*
> ```cpp
>         uNodalExtrmc[i][k+ncomp] = large;
>       for(std::size_t k = 0; k < nprim; k++)
>         pNodalExtrmc[i][k+nprim] = large;
> ```

Will the second block (loops) overwrite the second blocks of what has already been set to `-large` in the first block?

Let's not do the searches of `uNodalExtrmc[i]` `ncomp` times!

Let's not do the searches of `pNodalExtrmc[i]` `nprim` times!
___
*[src/PDE/DGPDE.hpp, line 445 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiXKYsDYFshKWFUYKh:-MaiXKYtE6aWB814q-P_:b-wyf94r) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/PDE/DGPDE.hpp#L445)):*
> ```cpp
>                   const std::unordered_map< std::size_t, std::size_t >& bid,
>                   tk::Fields& uNodalExtrm,
>                   tk::Fields& pNodalExtrm,
> ```

`uNodalExtrm` and `pNodalExtrm` are only read from. Let's make them `const` so the compiler knows how we use these so it can perform additional optimizations. This constness trickles down all the way to `VertexBasedFunction()`.

jbakosi (commented):
> Reviewed 1 of 11 files at r4, 8 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: all files reviewed, 10 unresolved discussions (waiting on @WeizhaoLi2018)

WeizhaoLi2018 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: all files reviewed, 10 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/DG.hpp, line 314 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiNWoz9Yzwf7PWNAjY:-Mb7yazS5xODeOc1u4OR:b-cmt4a6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.hpp#L314)):*

The doxygen comment in two line will only show up correctly on the web/doc if it is prefixed by `\brief`. (If it is one-line, it is okay without it.)
```c++
//! \brief Counter signaling that we have received all our nodal extrema from
//!       ghost chare partitions
```
</blockquote></details>

Done! Thanks!
___
*[src/Inciter/DG.hpp, line 340 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiR20PC3uvVAIVLd0n:-Mb7yvZr1eQxrlqZ9r2d:brr53h1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.hpp#L340)):*

The second one will not have a comment/explanation in the doxygen doc. Let's add something!
</blockquote></details>

Just added. Thanks!
___
*[src/Inciter/DG.cpp, line 142 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiP8AV1RrMGUqGam18:-Mb7zdtT4BIXpS5SaTF5:b-vj6v3m) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L142)):*

There is no need to pass any of the arguments, since they are all accessible from the function body. And when this is called from `DG::resizePostAMR()`, the node communication map in `Discretization` will have already been updated.
</blockquote></details>

Just removed the arguments. Thanks!
___
*[src/Inciter/DG.cpp, line 142 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiPMs44Ivx4rfh59ln:-Mb7zh2H1RbGgNY-pgGS:b-tanz96) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L142)):*

Let's keep the convention of camel-case for function names: `resizeNodalExtrema`. (Types start with uppercase.)
</blockquote></details>

Fixed! Thank you!
___
*[src/Inciter/DG.cpp, line 1877 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQeWh567SIvGumdRP:-Mb81_rz4osCQ3B5sLz5:bol3kou) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1877)):*

Let's write it shorter:
```c++
std::vector< std::vector< tk::real > > g1( n.size() ), g2( n.size() );
```
</blockquote></details>

Done! Thank you!
___
*[src/Inciter/DG.cpp, line 1882 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQXvH0OgeP_jPBopo:-Mb81qVtDbb2pHaqXnSQ:b-cmt4a6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1882)):*

Let's not perform the find twice.
</blockquote></details>

Done! Thanks!
___
*[src/Inciter/DG.cpp, line 1884 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiQQTJ7GSn1hyqo-yL:-Mb81yIj2Ipdirgs9J3T:bka5xvu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1884)):*

Let's keep to 80 columns.
</blockquote></details>

Fixed! Thanks!
___
*[src/Inciter/DG.cpp, line 1923 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiRKty8Js2x0w96NsO:-Mb82q_53ecEkgTY3Fa5:b-b09y3v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1923)):*

The above performs a _search_ behind `m_uNodalExtrmc[gid[i]]` `4*c` times(!) yielding the same result. Let's do that only once:
```c++
  for (std::size_t i=0; i<gid.size(); ++i) {
    auto& n = m_uNodalExtrmc[gid[i]];
    for (std::size_t c=0; c<ncomp; ++c) {
      n[c] = std::max( G1[i][c], n[c] );
      n[c+ncomp] = std::min( G1[i][c+ncomp], n[c+ncomp] );
    }
  }
```
Same below for the primitive variables.

(Sorry, I think I missed this earlier.)
</blockquote></details>

Just fixed it! Thank you for letting me know this. I didn't realize this when I first wrote this code.
___
*[src/Inciter/DG.cpp, line 1967 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiWLcK8Iqr9mASnj4x:-Mas-Ffg-er3yZmB_xTr:b-qqnkbw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/Inciter/DG.cpp#L1967)):*

Will the second block (loops) overwrite the second blocks of what has already been set to `-large` in the first block?

Let's not do the searches of `uNodalExtrmc[i]` `ncomp` times!

Let's not do the searches of `pNodalExtrmc[i]` `nprim` times!
</blockquote></details>

Yes, the second overwrites half of what has been set in the first block to `-large`. For `NodalExtrmc`, the first half of elements in  each vector should be `-large`(used for maximum initialization) and the rest of the elements should be `large`(used for minimum initialization).
Just fix this part. Let me know whether this makes sense.
___
*[src/PDE/DGPDE.hpp, line 445 at r5](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-MaiXKYsDYFshKWFUYKh:-Mb80adu30eX0uH57DFO:b-cmt4a6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/1c2cf1f7d1b88abac301f3d3aff48d46d7d2257c/src/PDE/DGPDE.hpp#L445)):*

`uNodalExtrm` and `pNodalExtrm` are only read from. Let's make them `const` so the compiler knows how we use these so it can perform additional optimizations. This constness trickles down all the way to `VertexBasedFunction()`.
</blockquote></details>

Done! Thanks!

adityakpandare (approved):
> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: all files reviewed, 10 unresolved discussions (waiting on @jbakosi)

jbakosi (approved):
Looks good! Thanks for the fixes.

> Reviewed 8 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Mb88wDZ1q5-bFVlPMxd:b-eorl45)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

WeizhaoLi2018 (commented):
Thank you guys! I will merge this branch once it passes the online test.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491#-:-Mb89_lrDjfEYR_bLns7:bxdgyq9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

jbakosi (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/491)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

PR #493, opened 2021-05-28
Sponge pressure

jbakosi:
This adds a capability to reduce the normal pressure gradient at selected side sets where symmetry boundary conditions are applied. This is similar to removing a percentage of the kinetic energy, but instead of reducing the tangential velocity at the wall, this scales back the normal pressure derivative at the wall. Both intended to model the effect of a solid wall on the fluid via fluid-structure interaction. Sponge-velocity will create an effect similar to viscosity, while spong-pressure decreases the normal pressure gradient as if a solid wall has moved outward.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)


adityakpandare (changes requested):
> Reviewed 13 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbC3Rgb60mu-pQ685AC:b-fjil3o) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*
> ```cpp
>                        rR, ruR, rvR, rwR, reR,
>                        w1L, w2L, w3L, w1R, w2R, w3R,
>                        spmult.data()+p*nset, spmult.data()+q*nset, nset,
> ```

Should the domain integral be affected by the sponge-pressure BC? I believe only `bndint` should be affected by it.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbC6JXTAHpraVoCHXLT:bqtc1ed) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*
> ```cpp
>           g_inputdeck.get< tag::param, eq, tag::sponge, tag::pressure >();
>         nset = sbc[m_system].size();  // number of symbc side sets configured
>         spmult.resize( x.size() * nset, 0.0 );
> ```

Why does `spmult` have to be of size `x.size()*nset`, and not `symbcnodes.size()*nset`?

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCAuyN-552dnSndmgE:b-8eg9v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Should the domain integral be affected by the sponge-pressure BC? I believe only `bndint` should be affected by it.
</blockquote></details>

No, the domain integral applies to all nodes and the two boundary integrals to boundary nodes only.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCC0cp0nGT3m8ejfLz:b-9bowrx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Why does `spmult` have to be of size `x.size()*nset`, and not `symbcnodes.size()*nset`?
</blockquote></details>

It does not have to be this way, but I thought it was simpler this way, because this way we do not need a map between local node ids and a "symbcnode" id. If we make it `symbcnodes.size()*nset`, we would need to access the pressure multipliers via
```c++
 spmult[ map_from_local_id_to_symbcnode_id[p] * nset + s ] = ...
```

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCJrrO9aOKCRm6cnsC:b-rbk7o4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

No, the domain integral applies to all nodes and the two boundary integrals to boundary nodes only.
</blockquote></details>

Then why is `spmult` used here? If the `domainint()` is unaffected by BC's, `Rusanov::flux()` should be unaffected by the sponge-pressure BC.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCKJRt2s4xZRMNmBaN:b-4rjz8w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

It does not have to be this way, but I thought it was simpler this way, because this way we do not need a map between local node ids and a "symbcnode" id. If we make it `symbcnodes.size()*nset`, we would need to access the pressure multipliers via
```c++
 spmult[ map_from_local_id_to_symbcnode_id[p] * nset + s ] = ...
```
</blockquote></details>

A map is required between `symbcnodes` and local-node-ids? But in the lines below, `p` (from `symbcnodes`) is used directly to access `x` (from `coord`), so shouldn't the ids be the same?

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCLKbB3T-jJclENhYt:b-v37qza) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Then why is `spmult` used here? If the `domainint()` is unaffected by BC's, `Rusanov::flux()` should be unaffected by the sponge-pressure BC.
</blockquote></details>

Why? If the left or right state is on a symbc boundary where sponge pressure is prescribed, shouldn't that state be different and so affect the flux on that edge?
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCLf4R-I66A9frgkcc:bdw9ndi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

A map is required between `symbcnodes` and local-node-ids? But in the lines below, `p` (from `symbcnodes`) is used directly to access `x` (from `coord`), so shouldn't the ids be the same?
</blockquote></details>

Yes, `p` is the local node id (not a running index). You can use that to index anything with the size that equals the number of nodes on that chare, but not something with a different size, e.g., `symbcnodes.size()`.

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCMY8IDC32mWa5Xthy:b-1kdsrv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Why? If the left or right state is on a symbc boundary where sponge pressure is prescribed, shouldn't that state be different and so affect the flux on that edge?
</blockquote></details>

So in this implementation, the sponge-pressure BC *does* affect the `domainint()`. My understanding is that the `domainint` should not be affected by BCs; just like we apply other types of BCs only in the `bndint()`. So basically, `domainint()` should be ignorant of the sponge-BC, just like it is ignorant of other BCs.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCNHZZ0cT-LwheMOpf:b-6jf24q) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Yes, `p` is the local node id (not a running index). You can use that to index anything with the size that equals the number of nodes on that chare, but not something with a different size, e.g., `symbcnodes.size()`.
</blockquote></details>

Ah, I get it. Then how about making `spmult` into a map? That would allow accessing `p` without a mapping between node-ids, while keeping the memory low (size will be only equal to number of nodes with symbc, as opposed to total number of nodes)

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCOm_e7PCzSRxuhkIc:bb7njb0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

So in this implementation, the sponge-pressure BC *does* affect the `domainint()`. My understanding is that the `domainint` should not be affected by BCs; just like we apply other types of BCs only in the `bndint()`. So basically, `domainint()` should be ignorant of the sponge-BC, just like it is ignorant of other BCs.
</blockquote></details>

Yeah, you are right, I guess what you are saying has been our BC "policy" for ALECG. I can remove it from the domain integral to make it consistent with the other BCs, but somehow I felt like this was the right thing to do. Not sure how much of a difference it will make for the pressure sponge BCs. Should I remove it from the domain integral?
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCNtZl8YjYZuVpKW8D:bhmtm5m) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Ah, I get it. Then how about making `spmult` into a map? That would allow accessing `p` without a mapping between node-ids, while keeping the memory low (size will be only equal to number of nodes with symbc, as opposed to total number of nodes)
</blockquote></details>

Yeah, that was also my first thought too, but then neither the boundary nor the domain integrals will vectorize. (Though we do not gain much performance from either being vectorized at this point, I'd like to keep the door open for it by keeping the data access pattern simple, which requires no search in maps, sets, or anything fancy.)

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 1 unresolved discussion (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCT80uEO1vhgmb-5Lw:b-4u9fjf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Yeah, you are right, I guess what you are saying has been our BC "policy" for ALECG. I can remove it from the domain integral to make it consistent with the other BCs, but somehow I felt like this was the right thing to do. Not sure how much of a difference it will make for the pressure sponge BCs. Should I remove it from the domain integral?
</blockquote></details>

Yes, the BC "policy" for ALECG is what I meant. We could remove this from the domain int and see how it affects the results. This way, the `Rusanov::flux()` function is also ignorant of the BC's, which is how numerical flux functions should be.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCSzdEAXBVnMmLBK9O:b-cnw9d3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Yeah, that was also my first thought too, but then neither the boundary nor the domain integrals will vectorize. (Though we do not gain much performance from either being vectorized at this point, I'd like to keep the door open for it by keeping the data access pattern simple, which requires no search in maps, sets, or anything fancy.)
</blockquote></details>

Makes sense. Let's leave this be then. Thanks for the explanation!

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbCTxLY6ZXNXVyXkMDG:b5d0hyb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Yes, the BC "policy" for ALECG is what I meant. We could remove this from the domain int and see how it affects the results. This way, the `Rusanov::flux()` function is also ignorant of the BC's, which is how numerical flux functions should be.
</blockquote></details>

Okay, sounds good. I have done a couple of runs with this already and the third one is running right now. I used 0.001 and 0.01 and they both had the intuitively (qualitatively) correct effect but extremely little. So now I'm running the same with 0.1, i.e., reducing the `dp/dn` by 10%. Next then I will remove it from the domain integral and rerun the 0.1 case and compare. I will let you know how it does, it will be a couple of days. I will not attempt to merge this until we see the effect of this.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbCU_QV1vXiayhmDz76:b-69yh2w) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Makes sense. Let's leave this be then. Thanks for the explanation!
</blockquote></details>

Okay, as always, I highly appreciate the extra set of scrutinizing eyes. ;-)

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 13 of 22 files reviewed, 1 unresolved discussion (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MbC_WmRD3DEOeaYdWGX:brxp9sq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Okay, sounds good. I have done a couple of runs with this already and the third one is running right now. I used 0.001 and 0.01 and they both had the intuitively (qualitatively) correct effect but extremely little. So now I'm running the same with 0.1, i.e., reducing the `dp/dn` by 10%. Next then I will remove it from the domain integral and rerun the 0.1 case and compare. I will let you know how it does, it will be a couple of days. I will not attempt to merge this until we see the effect of this.
</blockquote></details>

Sounds good.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1591 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC6JXS--EKlHEJDwkt:-MbC_R9R6H9EqfH5m25C:b3h9v4i) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1591)):*

Okay, as always, I highly appreciate the extra set of scrutinizing eyes. ;-)
</blockquote></details>

No problem. I learn from the best :wink:

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 3 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-McLewMl1WKBiwXr8mUy:b-d29spk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Sounds good.
</blockquote></details>

Alright, should we merge this? As far as I remember, the sponge being in both domain and boundary integrals was more stable. I still have to ensure we do not modify the flux within but from the outside. Alright, stay tuned.

adityakpandare (changes requested):
I'll wait for you to modify the Rusanov function (change it back to what it was) by modifying the pressures outside Rusanov.

> Reviewed 12 of 18 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-:-McLpIOr9PYTRHTR4NPU:bs2cnr7)* status: 15 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-McLpG5y6pXYzeEFBZf-:bbhsstc) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Alright, should we merge this? As far as I remember, the sponge being in both domain and boundary integrals was more stable. I still have to ensure we do not modify the flux within but from the outside. Alright, stay tuned.
</blockquote></details>

Yes, that's right, once Rusanov is reverted to its previous state, this is good to go.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 518 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-McLpp6OEYaBwgVBpOYu:-McLpp6OEYaBwgVBpOYv:b-20vg73) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4389d1f1ff57d0c1b47f7dd2fd4fbaae64feff5f/src/PDE/CompFlow/CGCompFlow.hpp#L518)):*
> ```cpp
>       for (ncomp_t c=0; c<m_ncomp; ++c) R.fill( c, m_offset, 0.0 );
> 
>       // compute sponge pressure multiplers at symmetry BCs
> ```

This comment is no more entirely correct, since sponge-BCs are new separate from symmetry-BCs. Let's just remove the "at symmetry BCs" part.

adityakpandare (changes requested):
> Reviewed 1 of 18 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 16 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 16 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-McjedIO6rpagXeK8Ugn:bbhgu5q) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Yes, that's right, once Rusanov is reverted to its previous state, this is good to go.
</blockquote></details>

I just looked at this closer. So sponge pressure requires modifying the pressure but pressure is evaluated inside the flux function. If we want to apply the sponge parameter to the pressure we either: (1) evaluate the pressure inside and apply sponge, (as is currently in this PR), or (2)  have to evaluate the pressure outside of the flux-function so we can apply the sponge on the pressure and then pass this modified pressure into the flux function. Either case will require modifying the old version of the flux function: (1) adds the pressure multipliers, and (2) adds the pressure. Which one should we do?

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 16 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 518 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-McLpp6OEYaBwgVBpOYu:-McjgE022ArWH6b5Ju9l:b-sv1i77) ([raw file](https://github.com/quinoacomputing/quinoa/blob/4389d1f1ff57d0c1b47f7dd2fd4fbaae64feff5f/src/PDE/CompFlow/CGCompFlow.hpp#L518)):*

This comment is no more entirely correct, since sponge-BCs are new separate from symmetry-BCs. Let's just remove the "at symmetry BCs" part.
</blockquote></details>

Good catch, will push a fix for this shortly.

jbakosi (commented):
Done, see ae591b7.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-:-MdJMnXz6jjdS5y3uysq:b-6yydhk)* status: 8 of 26 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

adityakpandare (approved):
Just one quick comment about the `sponge()` function, otherwise looks good.

> Reviewed 2 of 18 files at r2, 9 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-:-MdJPQhN7s5S0BkWb7Lq:b78hwz5)* status: 19 of 26 files reviewed, 1 unresolved discussion (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MdJO5yT7Z1C5YNqEl2M:bnfmrpw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

I just looked at this closer. So sponge pressure requires modifying the pressure but pressure is evaluated inside the flux function. If we want to apply the sponge parameter to the pressure we either: (1) evaluate the pressure inside and apply sponge, (as is currently in this PR), or (2)  have to evaluate the pressure outside of the flux-function so we can apply the sponge on the pressure and then pass this modified pressure into the flux function. Either case will require modifying the old version of the flux function: (1) adds the pressure multipliers, and (2) adds the pressure. Which one should we do?
</blockquote></details>

Looks good! Now this is in line with the multimat Riemann solver, which also separately provides L/R pressures.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 848 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MdJO_7L-WrHOydPTwur:-MdJO_7M-efLrC4NH6HU:b7fh387) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ae591b7ae9c8ed50a44251aa02b0acdb8fa0d35e/src/PDE/CompFlow/CGCompFlow.hpp#L848)):*
> ```cpp
>     }
> 
>     //! Apply sponge conditions at sponge nodes
> ```

Do you think we should provide some explanation here as to what "sponge" BC really does (removes user-specified portion of KE), since it's not a conventional BC?

adityakpandare (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 19 of 26 files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 848 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MdJO_7L-WrHOydPTwur:-MdJRunmAI8r7Nd10gYZ:b-wh1ozc) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ae591b7ae9c8ed50a44251aa02b0acdb8fa0d35e/src/PDE/CompFlow/CGCompFlow.hpp#L848)):*

Do you think we should provide some explanation here as to what "sponge" BC really does (removes user-specified portion of KE), since it's not a conventional BC?
</blockquote></details>

I thought about this more, and I think this is good as it is. We have enough explanations in the "help" section for the sponge keyword.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 18 of 26 files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 1124 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MbC3Rgb60mu-pQ685AB:-MdJQApc5qNOqkaUSpJG:b-837i3v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/f676db8303ea37d8532037eab152b9d9df8f0d0d/src/PDE/CompFlow/CGCompFlow.hpp#L1124)):*

Looks good! Now this is in line with the multimat Riemann solver, which also separately provides L/R pressures.
</blockquote></details>

Okay.
___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 848 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MdJO_7L-WrHOydPTwur:-MdJTdlG1X0fcC8BChd7:b-vawlja) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ae591b7ae9c8ed50a44251aa02b0acdb8fa0d35e/src/PDE/CompFlow/CGCompFlow.hpp#L848)):*

I thought about this more, and I think this is good as it is. We have enough explanations in the "help" section for the sponge keyword.
</blockquote></details>

Well, I agree, and I have added some more details, see a26235caf. ;-)

adityakpandare (approved):
:lgtm:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-:-MdJTwpa5ig0StsIWOU-:bnfp4nl)* status: 19 of 26 files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[src/PDE/CompFlow/CGCompFlow.hpp, line 848 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-MdJO_7L-WrHOydPTwur:-MdJTrEI3QePXWwVL3XP:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ae591b7ae9c8ed50a44251aa02b0acdb8fa0d35e/src/PDE/CompFlow/CGCompFlow.hpp#L848)):*

Well, I agree, and I have added some more details, see a26235caf. ;-)
</blockquote></details>

Thanks!

jbakosi (commented):
Thanks.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493#-:-MdJUL4F6Lc4Uxesb-QH:b-ucfkn4)* status: 19 of 26 files reviewed, all discussions resolved (waiting on @adityakpandare)

adityakpandare (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/493)* status: 19 of 26 files reviewed, all discussions resolved (waiting on @jbakosi)

PR #497, opened 2021-06-23
ALECG ALE mode now updates mesh velocity during time stepping

jbakosi:
This hooks up updating the mesh velocity during ALECG's time step if ALE mesh motion is enabled.

As an example, the mesh velocity may be assigned to follow the fluid velocity, which is then optionally smoothed to mitigate mesh tangling. Both unit-, and regression tests are added that exercise this and other details that will be required later, such as computing the vorticity from the fluid velocity. There are also improvements and bugfixes to the conjugate gradients linear solver, used to solve a Laplacian for smoothing the mesh velocity.

There is a lot more detail in the commit messages.

Next steps:
- Inherit mesh velocity smoother boundary conditions from symmetry conditions generalized to arbitrarily oriented side sets
- Enable _dvCFL_ (CFL coefficient for ALE volume change)
- Test current mesh velocity smoother with a case that has non-zero vorticity, e.g., triple-point
- Implement Helmholtz decomposition to locally remove part of the vorticity (as in https://onlinelibrary.wiley.com/doi/abs/10.1002/fld.4403)
- Add input options to exercise ALE mesh motion by moving a boundary with a prescribed motion

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497)


adityakpandare (changes requested):
I've made notes of what we spoke about offline. Thanks for helping me go through the code @jbakosi! 

> Reviewed 37 of 63 files at r1, 1 of 1 files at r2, 5 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-:-MdYpwhXCOGNLlXI8hoN:b-yw6nqf)* status: 34 of 63 files reviewed, 5 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Control/Inciter/InputDeck/InputDeck.hpp, line 333 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdONCea9yw1__U9cdJ2:-MdONCea9yw1__U9cdJ3:bm3sm07) ([raw file](https://github.com/quinoacomputing/quinoa/blob/511aab01428f0514d459d612ca213766d3fd3c05/src/Control/Inciter/InputDeck/InputDeck.hpp#L333)):*
> ```cpp
>       get< tag::ale, tag::ale >() = false;
>       get< tag::ale, tag::dvcfl >() = 0.0;
>       get< tag::ale, tag::maxit >() = 50;
> ```

Can we make the default max-iteration number to 5?
___
*[src/Inciter/ALECG.cpp, line 1359 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdObn3J1i3TiITMhG1Q:-MdObn3J1i3TiITMhG1R:b-tta9xu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/511aab01428f0514d459d612ca213766d3fd3c05/src/Inciter/ALECG.cpp#L1359)):*
> ```cpp
>   // if not all Runge-Kutta stages complete, continue to next time stage,
>   // otherwise output field data to file(s)
>   if (m_stage < 3) advance( Disc()->Dt() ); else out();
> ```

Is this needed if `advance()` is called in `dt()`?
___
*[src/LinearSolver/ConjugateGradients.cpp, line 46 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY2pKp2BPemvU5Z5B5:-MdY2pKp2BPemvU5Z5B6:bnoem0k) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L46)):*
> ```cpp
>   m_gid( gid ),
>   m_lid( lid ),
>   m_nodeCommMap( nodecommmap ),
> ```

Can we not store `nodeCommMap`, `gid` and `lid` and access these through the `Discretization` class?
___
*[src/LinearSolver/ConjugateGradients.cpp, line 463 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY3TfoD6B-cFUIZOZM:-MdY3TfpAfIhIJKAK2fg:b-zepsu3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L463)):*
> ```cpp
>   // solution, x=x0, or the BCs are incomplete or wrong, in either case the
>   // solve cannot continue.
>   const auto eps = std::numeric_limits< tk::real >::epsilon();  
> ```

Can we remove the white-space at the end of the above line?
___
*[tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q, line 21 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdYn6XwDUt1tEAJwWLr:-MdYn6Xy1wuJcW35ffiB:b-epzx2m) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q#L21)):*
> ```q
>   end
> 
>   ale
> ```

Do you think it's a good idea to move this inside the `compflow...end` block?

jbakosi (commented):
Thanks to you for bearing with this large change and your suggestions. I have fixed a couple of things and more details, as I remember more of why I did things, below.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-:-Mdd2xIcDNtVgrkxj6PJ:b-lh65sh)* status: 34 of 63 files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[src/Control/Inciter/InputDeck/InputDeck.hpp, line 333 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdONCea9yw1__U9cdJ2:-Mdcdp7O-IzUsati4Qmj:b-9f087b) ([raw file](https://github.com/quinoacomputing/quinoa/blob/511aab01428f0514d459d612ca213766d3fd3c05/src/Control/Inciter/InputDeck/InputDeck.hpp#L333)):*

Can we make the default max-iteration number to 5?
</blockquote></details>

Yep, done.
___
*[src/Inciter/ALECG.cpp, line 1359 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdObn3J1i3TiITMhG1Q:-MdcgnVX9M1SDAlVNWgt:b-f3us9z) ([raw file](https://github.com/quinoacomputing/quinoa/blob/511aab01428f0514d459d612ca213766d3fd3c05/src/Inciter/ALECG.cpp#L1359)):*

Is this needed if `advance()` is called in `dt()`?
</blockquote></details>

It does not. As I suspected, this was a leftover from debugging the logic changes. I reverted it.
___
*[src/LinearSolver/ConjugateGradients.cpp, line 46 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY2pKp2BPemvU5Z5B5:-MdckszaAk2ykPWl3SGP:bszyu0t) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L46)):*

Can we not store `nodeCommMap`, `gid` and `lid` and access these through the `Discretization` class?
</blockquote></details>

The naive way of doing this, of course, causes a circular dependency between `Discretization` and `ConjugateGradients`.

One way to circumvent the problem is to template one class on the other, but that complicates code significantly because both are Charm++ chare arrays. Besides complication, this can also lead to very weird build-dependency problems. I have ended up using such things in Walker and they work there, but I think that could have been done better and I'd rather not complicate inciter with this.

Another approach I tried is to create a function pointer as a`std::function` that points to, e.g., the accessor `Discretization::Gid()` and pass that to `ConjugateGradient`'s constructor, store it (the function object) and then call it later when access is needed to access to `Discretization::m_gid`. The problem with this is that anything that is passed to the constructor of a Charm++ chare array must be PUPable and a` std::function` is not and I have no idea how to make it so. The underlying problem of PUPing a `std::function` of course is that internally that stores a function pointer which of course makes no sense to PUP because that is certainly different across migrating machines. (`Scheme` solves a similar problem, but only similar because that really solves the problem of virtual functions and inheritance in a way that is PUPable and does not lead to too complex code. It is similar because the virtual functions internally work with compiler-generated pointers to child-class functions. Charm++ has support for this, but it is ugly to use.)

Yet another approach might be to pass the accessor as a Charm++ `ckCallback` but then I think that will be called asynchronously, and that's not what we want here.

I don't have a good idea how to do this. Luckily, `gid`,` lid`, and `nodecommap` are not very large.
___
*[src/LinearSolver/ConjugateGradients.cpp, line 463 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY3TfoD6B-cFUIZOZM:-Mdcq2939Baj165q6Bx2:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L463)):*

Can we remove the white-space at the end of the above line?
</blockquote></details>

Done.
___
*[tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q, line 21 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdYn6XwDUt1tEAJwWLr:-MdcvnP77MJQchxuaM42:bpk9o20) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q#L21)):*

Do you think it's a good idea to move this inside the `compflow...end` block?
</blockquote></details>

Alright, I remember now this better and why I decided to move the `ale ... end` block _outside_ the equation blocks.

The input, parser, input deck hierarchy, as well as the solvers (`DiagCG,`, `ALECG`, `DG`) are designed to be able to handle multiple systems of equations and strong coupling among them, *all solved on the same mesh*. Since ALE moves the mesh, there should be a single equation system that is responsible for moving the mesh (likely a flow solver but from the software engineering viewpoint this is not required). Therefore, ALE is prescribed (and configured) at at level that is above every single equation system (e.g., flow solver or transport). If each equation system were allowed to define ALE mesh motion but there is only a single mesh, then which one should be the one that gets to move the mesh? So this is the rationale.

Of course, the "who is the boss" problem still shows up, because if the mesh is moved based on some "velocity", then who defines this velocity? But with the current design this question only shows up when we need to call `CGPDE::velocity()`, which then allows `CompFlow` to define this, and `Transport` is a no-op, which makes sense from the physics viewpoint.

Conversely, there is a bigger problem if we move the `ale ... end` block _inside_ equation blocks, e.g., compflow, a logical and also more of a software engineering one. The solvers (`DiagCG,`, `ALECG`, `DG`) do not know anything about what systems they solve and how many (and how many scalar components they solve and how they are coupled). Such detail is the job of the PDE classes. Since ALE, as is currently implemented, is done in `ALECG`, it would be awkward to access anything that is single-equation-system-specific, but easy if it is system-agnostic. The awkwardness in code would manifest in having to do loops over types (at compile-time) across all equation types any time we want to query an ALE parameter, e.g., the mesh velocity (movement) strategy (that is currently behind `g_inputdeck.get< tag::ale, tag::meshvelocity >()`), and every time we want to access an ALE parameter, we would have to run some code that allows deciding which system's ALE settings should win and that code would probably have to be at compile-time since the loop is over types.. This is more complex than the current design, which only needs this decision where the velocity is decided on.

I also may be unable to come up with a better idea for a design for this at this time, let me know if you can think if something better. My feeling is that even if compile-time loops would not be a complication, I would still prefer the design that configures ALE that pertains to a (single) mesh on which multiple (potentially) coupled systems are solved and which is then configured at a higher (outer) level compared to the equations and this is reflected in the input file having the `ale ... end` block outside the equation config blocks.

adityakpandare (changes requested):
Sure thing! I have one last question; then it's good to go.

> Reviewed 1 of 63 files at r1, 2 of 12 files at r3, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-:-Me0GHap3E39igSBwLZj:b-kr6vcz)* status: 39 of 63 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/LinearSolver/ConjugateGradients.cpp, line 46 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY2pKp2BPemvU5Z5B5:-Me0ASq-7Wl70VFyUgtn:b-m5waw6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L46)):*

The naive way of doing this, of course, causes a circular dependency between `Discretization` and `ConjugateGradients`.

One way to circumvent the problem is to template one class on the other, but that complicates code significantly because both are Charm++ chare arrays. Besides complication, this can also lead to very weird build-dependency problems. I have ended up using such things in Walker and they work there, but I think that could have been done better and I'd rather not complicate inciter with this.

Another approach I tried is to create a function pointer as a`std::function` that points to, e.g., the accessor `Discretization::Gid()` and pass that to `ConjugateGradient`'s constructor, store it (the function object) and then call it later when access is needed to access to `Discretization::m_gid`. The problem with this is that anything that is passed to the constructor of a Charm++ chare array must be PUPable and a` std::function` is not and I have no idea how to make it so. The underlying problem of PUPing a `std::function` of course is that internally that stores a function pointer which of course makes no sense to PUP because that is certainly different across migrating machines. (`Scheme` solves a similar problem, but only similar because that really solves the problem of virtual functions and inheritance in a way that is PUPable and does not lead to too complex code. It is similar because the virtual functions internally work with compiler-generated pointers to child-class functions. Charm++ has support for this, but it is ugly to use.)

Yet another approach might be to pass the accessor as a Charm++ `ckCallback` but then I think that will be called asynchronously, and that's not what we want here.

I don't have a good idea how to do this. Luckily, `gid`,` lid`, and `nodecommap` are not very large.
</blockquote></details>

Yeah, this is a problem. Let's leave it be then. Thanks for the explanation!
___
*[tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q, line 21 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdYn6XwDUt1tEAJwWLr:-Me0FwYl8gZ-dbrDP8kZ:b-1jne5x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q#L21)):*

Alright, I remember now this better and why I decided to move the `ale ... end` block _outside_ the equation blocks.

The input, parser, input deck hierarchy, as well as the solvers (`DiagCG,`, `ALECG`, `DG`) are designed to be able to handle multiple systems of equations and strong coupling among them, *all solved on the same mesh*. Since ALE moves the mesh, there should be a single equation system that is responsible for moving the mesh (likely a flow solver but from the software engineering viewpoint this is not required). Therefore, ALE is prescribed (and configured) at at level that is above every single equation system (e.g., flow solver or transport). If each equation system were allowed to define ALE mesh motion but there is only a single mesh, then which one should be the one that gets to move the mesh? So this is the rationale.

Of course, the "who is the boss" problem still shows up, because if the mesh is moved based on some "velocity", then who defines this velocity? But with the current design this question only shows up when we need to call `CGPDE::velocity()`, which then allows `CompFlow` to define this, and `Transport` is a no-op, which makes sense from the physics viewpoint.

Conversely, there is a bigger problem if we move the `ale ... end` block _inside_ equation blocks, e.g., compflow, a logical and also more of a software engineering one. The solvers (`DiagCG,`, `ALECG`, `DG`) do not know anything about what systems they solve and how many (and how many scalar components they solve and how they are coupled). Such detail is the job of the PDE classes. Since ALE, as is currently implemented, is done in `ALECG`, it would be awkward to access anything that is single-equation-system-specific, but easy if it is system-agnostic. The awkwardness in code would manifest in having to do loops over types (at compile-time) across all equation types any time we want to query an ALE parameter, e.g., the mesh velocity (movement) strategy (that is currently behind `g_inputdeck.get< tag::ale, tag::meshvelocity >()`), and every time we want to access an ALE parameter, we would have to run some code that allows deciding which system's ALE settings should win and that code would probably have to be at compile-time since the loop is over types.. This is more complex than the current design, which only needs this decision where the velocity is decided on.

I also may be unable to come up with a better idea for a design for this at this time, let me know if you can think if something better. My feeling is that even if compile-time loops would not be a complication, I would still prefer the design that configures ALE that pertains to a (single) mesh on which multiple (potentially) coupled systems are solved and which is then configured at a higher (outer) level compared to the equations and this is reflected in the input file having the `ale ... end` block outside the equation config blocks.
</blockquote></details>

I mostly agree with you. I have only one question regarding your last paragraph: how would ALE know which mesh to move, if there are multiple meshes set up for multiple systems?

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497)* status: 39 of 63 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/LinearSolver/ConjugateGradients.cpp, line 46 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdY2pKp2BPemvU5Z5B5:-Me0dkAM8Voi2yxfsnAZ:b6fon95) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/src/LinearSolver/ConjugateGradients.cpp#L46)):*

Yeah, this is a problem. Let's leave it be then. Thanks for the explanation!
</blockquote></details>

No problem. I think C++ modules will help a lot with resolving such circular dependency problems in the future, but this is not production-ready yet in any compiler that I know of.
___
*[tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q, line 21 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdYn6XwDUt1tEAJwWLr:-Me0e2Rk6ZMYVVV8KkUK:b1qlejn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q#L21)):*

I mostly agree with you. I have only one question regarding your last paragraph: how would ALE know which mesh to move, if there are multiple meshes set up for multiple systems?
</blockquote></details>

For when we have multiple meshes, each can have its own `ale ... end` block configuring its own mesh movement relative to itself (its own initial position). I think that will be relative easy to add: within the `ale ... end` block we could just specify which mesh we refer to and there can be multiple `ale ... end` blocks referring to different solvers (operating on different meshes). Then for each single mesh (with its own ALE config) there can still be multiple equation systems (solved on the same mesh) as it is currently.

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-:-Me0iz079DP2bnohT2nn:bnfp4nl)* status: 39 of 63 files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q, line 21 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/497#-MdYn6XwDUt1tEAJwWLr:-Me0iuNM4pGcQ0o6iioo:b979vx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/133205af2a416e3e738f7627c45c3f901881effd/tests/regression/inciter/ale/sod_shocktube_alecg_ale_fluid.q#L21)):*

For when we have multiple meshes, each can have its own `ale ... end` block configuring its own mesh movement relative to itself (its own initial position). I think that will be relative easy to add: within the `ale ... end` block we could just specify which mesh we refer to and there can be multiple `ale ... end` blocks referring to different solvers (operating on different meshes). Then for each single mesh (with its own ALE config) there can still be multiple equation systems (solved on the same mesh) as it is currently.
</blockquote></details>

Okay. Thanks!

PR #498, opened 2021-06-23
Docker

jbakosi:
Maintenance commits attempting to fix the debian and alpine docker builds due to updates to the docker daemon on the test machine and to the alpine linux distribution.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/498)


jbakosi:
No need to review, it tests pass, will merge.

PR #507, opened 2021-07-27
Add field-, surface-, and history-output in units of physics time

jbakosi:
Field-, surface-, and history-output can now be requested at physics time intervals. The keyword is `time_interval`, followed by a real number, specifying the increment in physics time. This works in the `history_output` and `field_output` blocks.

History output example
-------
```
  history_output
    interval 10
    time_interval 0.01
    point p1 0.1 0.05 0.025 end
    point p2 0.9 0.05 0.025 end
  end
```
The above will do history output in two points at every 10th time step _and_ approximately at every `dt=0.01`. Screen output:
```
             flg - status flags, legend:
                   f - field (volume and surface)
                   d - diagnostics
                   t - physics time history
                   h - h-refinement
                   l - load balancing
                   r - checkpoint
                   a - ALE mesh velocity linear solver did not converge
      
       it             t            dt        ETE        ETA        EGT  flg
  -------------------------------------------------------------------------
        1  5.888815e-03  5.888815e-03  000:00:00  000:00:00         66  dlr
        2  1.067889e-02  4.790079e-03  000:00:00  000:00:00         65  dtlr
        3  1.509437e-02  4.415478e-03  000:00:00  000:00:00         62  dlr
        4  1.930815e-02  4.213775e-03  000:00:00  000:00:00         63  dlr
        5  2.338267e-02  4.074524e-03  000:00:00  000:00:00         64  dtlr
        6  2.728611e-02  3.903436e-03  000:00:00  000:00:00         63  dlr
        7  3.106963e-02  3.783524e-03  000:00:00  000:00:00         65  dtlr
        8  3.477897e-02  3.709340e-03  000:00:00  000:00:00         62  dlr
        9  3.844561e-02  3.666637e-03  000:00:00  000:00:00         63  dlr
       10  4.206636e-02  3.620752e-03  000:00:00  000:00:00         70  dt
```

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507)


adityakpandare (changes requested):
Thanks for adding this! I have a few questions below. One general question: in the example above, both time *and* iteration based history outputs are set up. Can history (or field) outputs be configured to be written for *either* time *or* iteration frequencies?

> Reviewed 17 of 30 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-:-MfrlOHI6HK6OOZJOtQL:bpn6d5g)* status: 17 of 30 files reviewed, 4 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/ALECG.cpp, line 1472 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrtqS2492NcgC-TT1Q:-MfrtqS2492NcgC-TT1R:b-u8wna8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/ALECG.cpp#L1472)):*
> ```cpp
>   // physics time output frequency is hit or in the last time step
>   if (d->fielditer() or d->fieldtime() or m_finished)
>     writeFields( CkCallback(CkIndex_ALECG::step(), thisProxy[thisIndex]) );
> ```

How are the field output files based on time-freq and iteration-freq distinguished?
___
*[src/Inciter/Discretization.cpp, line 974 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrvxaBBkfdSUcvUb4n:-MfrvxaBBkfdSUcvUb4o:b-qqjebb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L974)):*
> ```cpp
>   const auto eps = std::numeric_limits< tk::real >::epsilon();
>   const auto ft = g_inputdeck.get< tag::interval_time, tag::field >();
>   if (std::abs(ft) > eps) m_physFieldFloor = std::floor( m_t / ft );
> ```

Why is an `abs` required here? Isn't `ft` supposed to be positive? If user specifies a negative value for `ft`, shouldn't we error out at the input-file parsing stage?
___
*[src/Inciter/Discretization.cpp, line 1107 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrwjLb-Xl9kQ_rXgXo:-MfrwjLb-Xl9kQ_rXgXp:b-35gckm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1107)):*
> ```cpp
>   const auto ft = g_inputdeck.get< tag::interval_time, tag::field >();
> 
>   if (std::abs(ft) < eps) return false;
> ```

Same here, is `abs` really required?
___
*[src/Inciter/Discretization.cpp, line 1109 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-Mfrx8zJ873YPomAYTU0:-Mfrx8zJ873YPomAYTU1:b-t8zvq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1109)):*
> ```cpp
>   if (std::abs(ft) < eps) return false;
> 
>   return std::abs(std::floor(m_t/ft) - m_physFieldFloor) > eps;
> ```

Just to make sure I understood this: `m_physfieldfloor = abs(m_t/ft)` has been set in `next()`, but then `m_t` has been incremented by `dt` since then. So if the `ft` time interval has been crossed after this increment, the above statement will return true, triggering time-freqency based field-output. Is that correct?

jbakosi (commented):
> Can history (or field) outputs be configured to be written for either time or iteration frequencies?

Yes, my intention was that if you only specify `interval`, you only get output based on iteration count, and if you only specify `time_interval`, you only get physics time output. If you specify both, you get the union of both, i.e., output at all time steps whose condition evaluate to true for the output (independent of how the logic comes together from pieces, combining them with logical _or_). But as I see, the defaults for the former are 1 so I think we should change those to a large number and that would achieve this and we are leaving the default physics time intervals uninitialized which evaluate to zero and testing for them on positivity later so those are fine. What do you think?

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-:-Mfs2jFl5Z_wGEgPXm5d:bvuiebn)* status: 17 of 30 files reviewed, 4 unresolved discussions (waiting on @adityakpandare)

___
*[src/Inciter/ALECG.cpp, line 1472 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrtqS2492NcgC-TT1Q:-Mfs41DE7b6aWQXg_WGn:b-67es6n) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/ALECG.cpp#L1472)):*

How are the field output files based on time-freq and iteration-freq distinguished?
</blockquote></details>

They are not. They are going to the same file (as before). The output (i.e., what happens) is the same and you get the union, so there will be output based on both, we don't save them to different files. I see no good reason to save different files from the same simulation just because the output is saved at different output intervals - if that's what you meant (but maybe I'm misunderstanding you here). Here the idea is that the output should happen in a given time step if _either_ of the conditions is true: iteration count interval is hit _or_ physics time interval is hit _or_ in the very last time step.
___
*[src/Inciter/Discretization.cpp, line 974 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrvxaBBkfdSUcvUb4n:-Mfs4hWa-1VSbxsHxxnV:b-qkcwz3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L974)):*

Why is an `abs` required here? Isn't `ft` supposed to be positive? If user specifies a negative value for `ft`, shouldn't we error out at the input-file parsing stage?
</blockquote></details>

Yeah. This is half-baked. Nice catch! The only reason I put in abs is because a real number can be negative. There is a lower bound given for the keyword and there is a test in the parser for the lower and upper bounds _but_ as I have just checked this, it only issues a warning if the value is out of bounds. I am going to fix this by turning those warnings to be errors. Also, I think the bounds for these should be positive, i.e., 0 is not acceptable if specified (will keep zero as the default which can mean 'unspecified'). After those fixes, the abs can and should go away.
___
*[src/Inciter/Discretization.cpp, line 1107 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrwjLb-Xl9kQ_rXgXo:-Mfs77bdCazlU0hIApsu:b-69s4y9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1107)):*

Same here, is `abs` really required?
</blockquote></details>

No. Same as above, will fix it.
___
*[src/Inciter/Discretization.cpp, line 1109 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-Mfrx8zJ873YPomAYTU0:-Mfs7nYj8n9RAPBc9KM5:bi8bt0l) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1109)):*

Just to make sure I understood this: `m_physfieldfloor = abs(m_t/ft)` has been set in `next()`, but then `m_t` has been incremented by `dt` since then. So if the `ft` time interval has been crossed after this increment, the above statement will return true, triggering time-freqency based field-output. Is that correct?
</blockquote></details>

My thinking here is that we want an output based on physics time in a given time step if the (new/current) physics time  (in`m_t` after it has been incremented for the current time step) has _just been_ crossed the discrete increment of `m_t/ft`, i.e., if `m_t/ft` is "newer" (larger) than the previous `m_t/ft`. This is implemented by comparing the current floor to the previous floor of `m_t/ft`.

(I was not entirely happy with the fact that a logic test, such as in `Discretization::fieldtime()` cannot be made `const`, because it changes the class state, because at first I was updating `m_physPhysFloor` there and only if there was output. Then I realized that the correct place to update `m_physPhysFloor` is (can also be) in `next()`, independent of whether there was output in a given time step. I think this logic is correct and appears to produce the desired behavior. So now I made both `histtime()` and `fieldtime()` `const`, to which your question reminded me of.)

adityakpandare (commented):
Yes, I agree for the most part. What I am now confused about is this: if a user configures `field_output` based on `interval` only (meaning the `time_interval` is left uninitialized, i.e. 0.0?), then by the above *or* logic, the `time_interval` condition will always be met, and `field_outputs` would happen every time step. Am I missing something?

> Reviewed 2 of 30 files at r1, 9 of 9 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-:-Mfsa0E4BLdAhQctG1Jl:b-wpvgtx)* status: 23 of 34 files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[src/Inciter/ALECG.cpp, line 1472 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrtqS2492NcgC-TT1Q:-MfsZk7c1a68yKIZ3hbI:b-xh3p6f) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/ALECG.cpp#L1472)):*

They are not. They are going to the same file (as before). The output (i.e., what happens) is the same and you get the union, so there will be output based on both, we don't save them to different files. I see no good reason to save different files from the same simulation just because the output is saved at different output intervals - if that's what you meant (but maybe I'm misunderstanding you here). Here the idea is that the output should happen in a given time step if _either_ of the conditions is true: iteration count interval is hit _or_ physics time interval is hit _or_ in the very last time step.
</blockquote></details>

I understand. And that was what I meant. Thanks for the explanation.
___
*[src/Inciter/Discretization.cpp, line 974 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-MfrvxaBBkfdSUcvUb4n:-Mfs_3SkD4ERAS38UDMH:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L974)):*

Yeah. This is half-baked. Nice catch! The only reason I put in abs is because a real number can be negative. There is a lower bound given for the keyword and there is a test in the parser for the lower and upper bounds _but_ as I have just checked this, it only issues a warning if the value is out of bounds. I am going to fix this by turning those warnings to be errors. Also, I think the bounds for these should be positive, i.e., 0 is not acceptable if specified (will keep zero as the default which can mean 'unspecified'). After those fixes, the abs can and should go away.
</blockquote></details>

Thanks!
___
*[src/Inciter/Discretization.cpp, line 1109 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-Mfrx8zJ873YPomAYTU0:-Mfs_k4k24Bi83DE8w0d:b-1fvo6l) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1109)):*

My thinking here is that we want an output based on physics time in a given time step if the (new/current) physics time  (in`m_t` after it has been incremented for the current time step) has _just been_ crossed the discrete increment of `m_t/ft`, i.e., if `m_t/ft` is "newer" (larger) than the previous `m_t/ft`. This is implemented by comparing the current floor to the previous floor of `m_t/ft`.

(I was not entirely happy with the fact that a logic test, such as in `Discretization::fieldtime()` cannot be made `const`, because it changes the class state, because at first I was updating `m_physPhysFloor` there and only if there was output. Then I realized that the correct place to update `m_physPhysFloor` is (can also be) in `next()`, independent of whether there was output in a given time step. I think this logic is correct and appears to produce the desired behavior. So now I made both `histtime()` and `fieldtime()` `const`, to which your question reminded me of.)
</blockquote></details>

Okay. Thanks for explaining!

jbakosi (commented):
No. If `time_interval` is left uninitialized, it is 0.0, and will be ignored. See `if (ft < eps) return false;` in `Discretization::fieldtime()`.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-:-MfsbtjFEG4vDgDeMx5x:b-8ucswv)* status: 23 of 34 files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[src/Inciter/Discretization.cpp, line 1109 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-Mfrx8zJ873YPomAYTU0:-MfscX0tB7U_hvnPdzTW:b-tv49k5) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6ece44c6e2c0679f909f7db0bce072d2149e722e/src/Inciter/Discretization.cpp#L1109)):*

Okay. Thanks for explaining!
</blockquote></details>

Not a problem! Thanks for asking. ;-)

adityakpandare (approved):
Ah okay. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/507#-:-MfsgZgt2okPyTFg9OjC:brn9doa)* status: 23 of 34 files reviewed, all discussions resolved (waiting on @jbakosi)

PR #511, opened 2021-08-23
ALE moving sidesets

jbakosi:
This puts in moving side sets to ALE. Now the user can prescribe a discrete function that is used to move a list of side sets. See `cyl_move.q` for an example. Forked from #510.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/511)

PR #515, opened 2021-09-17
Generic table

jbakosi:
This makes `tk::Table` and sampling from it generic. Forked from #511.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)


adityakpandare (changes requested):
I have a question about the `sample()` function.

> Reviewed 10 of 82 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-:-Mk7tTu_405MW6lubjbA:b-2f6178)* status: 10 of 82 files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Base/Table.hpp, line 53 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2RLmS38US4sW0TRSx:-Mk2RB9N-K2TCs_BkGYG:bjw4kss) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L53)):*
> ```cpp
> 
>   auto eps = std::numeric_limits< real >::epsilon();
>   if (x < abs(table.front())+eps) return ord( table.front() );
> ```

This line of code returns the first set of ordinates if the value of x is less than the first abscissa in the table, right? In that case, I'm not sure I understand this.
1. Wouldn't `table.front()` return the first element of the vector, which is an array (because table is a vector of arrays), whereas `x` is a scalar `real`. So the comparison should be with `table.front()[0]`.
2. The abscissa could be negative, in which case, if `x` is positive, the code does the wrong thing, by returning the first set of ordinates for an `x > table.front()[0]` because of the `abs()`. Due to this, the `abs()` should not be used, and the comparison should be `x < table.front()[0]+eps`.
___
*[src/Base/Table.hpp, line 59 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2TlnZ2NhHQ5gTKQuW:-Mk2TlnZ2NhHQ5gTKQuX:b-pccgjd) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L59)):*
> ```cpp
>     auto t2 = abs( table[i+1] );
>     if (t1 < x and x < t2) {
>       auto d = (t2-t1)/(x-t1);
> ```

Here, are `t1` and `t2` the i-th and i+1th arrays in the vector? If so, shouldn't this be `d=(t2[0]-t1[0])/(x-t1[0])`?

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)* status: 10 of 82 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/Base/Table.hpp, line 53 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2RLmS38US4sW0TRSx:-Mk7ux9V7BVIRT2b-RaX:b-osv25c) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L53)):*

This line of code returns the first set of ordinates if the value of x is less than the first abscissa in the table, right? In that case, I'm not sure I understand this.
1. Wouldn't `table.front()` return the first element of the vector, which is an array (because table is a vector of arrays), whereas `x` is a scalar `real`. So the comparison should be with `table.front()[0]`.
2. The abscissa could be negative, in which case, if `x` is positive, the code does the wrong thing, by returning the first set of ordinates for an `x > table.front()[0]` because of the `abs()`. Due to this, the `abs()` should not be used, and the comparison should be `x < table.front()[0]+eps`.
</blockquote></details>

`abs()` is not the absolute value, but the a lambda that returns the abscissa. We can rename to `abscissa` if that helps.
___
*[src/Base/Table.hpp, line 59 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2TlnZ2NhHQ5gTKQuW:-Mk7vHd17Y1MGHgOtXX5:b4ucdlb) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L59)):*

Here, are `t1` and `t2` the i-th and i+1th arrays in the vector? If so, shouldn't this be `d=(t2[0]-t1[0])/(x-t1[0])`?
</blockquote></details>

Same here. If it wasn't correct, it would not compile.

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)* status: 10 of 82 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/Base/Table.hpp, line 53 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2RLmS38US4sW0TRSx:-Mk7vvjR1kUQ-vp9eyAm:bvxn0z1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L53)):*

`abs()` is not the absolute value, but the a lambda that returns the abscissa. We can rename to `abscissa` if that helps.
</blockquote></details>

Ah, I missed that. We should rename it to `abscissa()` to avoid the confusion.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)* status: 10 of 82 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/Base/Table.hpp, line 53 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2RLmS38US4sW0TRSx:-Mk7wNLQ-ga9wwdH0jiW:be58rmz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L53)):*

Ah, I missed that. We should rename it to `abscissa()` to avoid the confusion.
</blockquote></details>

Sure, no problem. I will also rename `ord` to `ordinate`, to make it more explicit.

adityakpandare (approved):
:lgtm:

> Reviewed 14 of 82 files at r1, 24 of 64 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-:-Mk88vFbDEmvTF7N2GKZ:bnfp4nl)* status: 43 of 83 files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[doc/pages/build.dox, line 119 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk88fTN2MGvgJ_tNm_s:-Mk88fTPBYD0mWMzbVFm:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/bd62c1e0630e2658758afc532ed5707d372136d5/doc/pages/build.dox#L119)):*
> ```
> To make this work, we need to clone the TPLs separately:
> 
>        git clone --recurse-submodules https://github.com/quinoacomputing/quinoa-tpl.git
> ```

Thanks!
___
*[src/Base/Table.hpp, line 53 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/515#-Mk2RLmS38US4sW0TRSx:-Mk88pVRADreOLDu7Y8h:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/364638054fd5c304d05c155afbdd6f7fd78661af/src/Base/Table.hpp#L53)):*

Sure, no problem. I will also rename `ord` to `ordinate`, to make it more explicit.
</blockquote></details>

Thanks!

adityakpandare (approved):
> Reviewed 40 of 64 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbakosi)

adityakpandare (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/515)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbakosi)

PR #528, opened 2021-11-03
static cast for shockmarker assignment

adityakpandare:
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)

### DIFF
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -1493,6 +1493,13 @@ DG::extractFieldOutput(
 
   // Add shock detection marker array to element-centered field output
   std::vector< tk::real > shockmarker( begin(m_shockmarker), end(m_shockmarker) );
+  // Here m_shockmarker has a size of m_u.nunk() which is the number of the
+  // elements within this partition (nelem) plus the ghost partition cells. In
+  // terms of output purpose, we only need the solution data within this
+  // partition. Therefore, resizing it to nelem removes the extra partition
+  // boundary allocations in the shockmarker vector. Since the code assumes that
+  // the boundary elements are on the top, the resize operation keeps the lower
+  // portion.
   shockmarker.resize( nelem );
   for (const auto& [child,parent] : addedTets)
     shockmarker[child] = static_cast< tk::real >(m_shockmarker[parent]);
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -1495,7 +1495,7 @@ DG::extractFieldOutput(
   std::vector< tk::real > shockmarker( begin(m_shockmarker), end(m_shockmarker) );
   shockmarker.resize( nelem );
   for (const auto& [child,parent] : addedTets)
-    shockmarker[child] = m_shockmarker[parent];
+    shockmarker[child] = static_cast< tk::real >(m_shockmarker[parent]);
   m_elemfields.push_back( shockmarker );
 
   // Send node fields contributions to neighbor chares


jbakosi (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[src/Inciter/DG.cpp, line 1498 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaU2415piAzpst2Tdt:-MnaU2423DRgRKsdxBIX:bokmpba) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1498)):*
> ```cpp
>   shockmarker.resize( nelem );
>   for (const auto& [child,parent] : addedTets)
>     shockmarker[child] = static_cast< tk::real >(m_shockmarker[parent]);
> ```

Is there a disadvantage of storing `DG::m_shockmarker` as a `vector< tk:real >`? (That would eliminate the cast, and the storage is the same: 64 bit in both cases.)

jbakosi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/Inciter/DG.cpp, line 1496 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaUlPk-4K2fDslyfcT:-MnaUlPk-4K2fDslyfcU:b-xbrxwu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1496)):*
> ```cpp
>   // Add shock detection marker array to element-centered field output
>   std::vector< tk::real > shockmarker( begin(m_shockmarker), end(m_shockmarker) );
>   shockmarker.resize( nelem );
> ```

Also, this resize appears to be superfluous, since it follows a constructor that already sized the container to `m_shockmarker.size()`.

WeizhaoLi2018 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/DG.cpp, line 1496 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaUlPk-4K2fDslyfcT:-MnaYztL6FgM3GxRIYL0:bdm0iwa) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1496)):*

@weizhaoli2018 knows this code better. @weizhaoli2018, could you comment?
</blockquote></details>

`m_shockmarker` has a size of `m_u.nunk()` which is the number of elements within this partition (`nelem`) plus the ghost partition cells. In terms of output purpose, we only need the information within this partition. Therefore, here we resize it to `nelem` to remove the extra partition boundary allocations in the `shockmarker` vector.
___
*[src/Inciter/DG.cpp, line 1498 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaU2415piAzpst2Tdt:-Mna_7lZ26p6Iip1r6Bi:bz2ei89) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1498)):*

@weizhaoli2018 knows this code better. @weizhaoli2018, could you comment?
</blockquote></details>

I think it is okay to change it to `tk::real`.
Previously I use `std::size_t` because I tried to mark the cells applied limiter as 1 and the cells without limiter as 0 in `PDE/Limiter.cpp` so that we can just use `shockmarker == 1` to check whether we need to apply limiter or not. When it changes to `tk::real` I think we could change the condition in `PDE/Limiter.cpp` as `shockmarker > 0.5`. If this sounds good to you guys, I could push a commit to fix this.

jbakosi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @WeizhaoLi2018)

___
*[src/Inciter/DG.cpp, line 1496 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaUlPk-4K2fDslyfcT:-MnaatA5CJOtrolePZ_O:b-4gx5d3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1496)):*

`m_shockmarker` has a size of `m_u.nunk()` which is the number of elements within this partition (`nelem`) plus the ghost partition cells. In terms of output purpose, we only need the information within this partition. Therefore, here we resize it to `nelem` to remove the extra partition boundary allocations in the `shockmarker` vector.
</blockquote></details>

Uh! Nice! Then can you please put this in comments here? This is very-very useful for anyone who attempts to understand this code. (Also mention that this code assumes that the boundary elements are assumed to be on the top, so you are keeping the lower portion. I know this might seem obvious to you guys, but as this codes gets more and more complex, more and more comments will be very helpful.)
___
*[src/Inciter/DG.cpp, line 1498 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaU2415piAzpst2Tdt:-MnabbEbB9yaehVFixJX:b-d4p9ef) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1498)):*

I think it is okay to change it to `tk::real`.
Previously I use `std::size_t` because I tried to mark the cells applied limiter as 1 and the cells without limiter as 0 in `PDE/Limiter.cpp` so that we can just use `shockmarker == 1` to check whether we need to apply limiter or not. When it changes to `tk::real` I think we could change the condition in `PDE/Limiter.cpp` as `shockmarker > 0.5`. If this sounds good to you guys, I could push a commit to fix this.
</blockquote></details>

Hmm. That makes sense. What makes me uncomfortable with this is the that now the test involves a somewhat nontrivial expression on a real number (instead of a simple bool true/false test): `shockmarker > 0.5`, in which the 0.5 seems like a magic number (e.g., someone might want to play with in the future). I would just keep the `std::size_t` and the cast, because that is at least standard (= everyone should immediately understand it), and if we convert the whole container to `tk::real`, then in theory now the container could hold arbitrary real numbers for no good reason (e.g., due to some programmer error in the future), for which now we should write some asserts that ensure that the container should only hold 0.0s and 1.0s, which makes this whole thing unnecessarily complicated. In summary, I would keep the integer and the cast.

WeizhaoLi2018 (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/DG.cpp, line 1496 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaUlPk-4K2fDslyfcT:-Mnal6XK4ZtIkzPWmnCw:bw2t611) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1496)):*

Uh! Nice! Then can you please put this in comments here? This is very-very useful for anyone who attempts to understand this code. (Also mention that this code assumes that the boundary elements are assumed to be on the top, so you are keeping the lower portion. I know this might seem obvious to you guys, but as this codes gets more and more complex, more and more comments will be very helpful.)
</blockquote></details>

Just added the comment. Thanks!

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528)* status: all files reviewed, 2 unresolved discussions (waiting on @jbakosi)

___
*[src/Inciter/DG.cpp, line 1496 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaUlPk-4K2fDslyfcT:-MnamypVEbfKXjs5mohP:b9sqou5) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1496)):*

Just added the comment. Thanks!
</blockquote></details>

Thanks @weizhaoli2018 and @jbakosi for pointing this out!
___
*[src/Inciter/DG.cpp, line 1498 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-MnaU2415piAzpst2Tdt:-MnanCz09sTgSo7w1KOL:bg0pqwu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/0c16c9dfc2dbb62d796692490111b3af5eab9b8f/src/Inciter/DG.cpp#L1498)):*

Hmm. That makes sense. What makes me uncomfortable with this is the that now the test involves a somewhat nontrivial expression on a real number (instead of a simple bool true/false test): `shockmarker > 0.5`, in which the 0.5 seems like a magic number (e.g., someone might want to play with in the future). I would just keep the `std::size_t` and the cast, because that is at least standard (= everyone should immediately understand it), and if we convert the whole container to `tk::real`, then in theory now the container could hold arbitrary real numbers for no good reason (e.g., due to some programmer error in the future), for which now we should write some asserts that ensure that the container should only hold 0.0s and 1.0s, which makes this whole thing unnecessarily complicated. In summary, I would keep the integer and the cast.
</blockquote></details>

Agreed, let's keep the `std::size_t` and the cast.

jbakosi (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/528#-:-MnasF-VB3Pn-q_dqTUL:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #529, opened 2021-11-08
remove factory policy output and unique code

adityakpandare:
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/529)

### DIFF
--- a/src/Base/Has.hpp
+++ b/src/Base/Has.hpp
@@ -28,18 +28,6 @@ template < typename T >
 inline constexpr bool HasTypedef_alias_v = HasTypedef_alias< T >::value;
 
 
-//! Detect if a type defines type 'code'
-template< typename, typename = std::void_t<> >
-struct HasTypedef_code : std::false_type {};
-
-template< typename T >
-struct HasTypedef_code< T, std::void_t< typename T::code > >
-  : std::true_type {};
-
-template < typename T >
-inline constexpr bool HasTypedef_code_v = HasTypedef_code< T >::value;
-
-
 //! Detect if a type defines type 'i_am_tagged_tuple'
 template< typename, typename = std::void_t<> >
 struct HasTypedef_i_am_tagged_tuple : std::false_type {};
--- a/src/Base/Print.hpp
+++ b/src/Base/Print.hpp
@@ -579,23 +579,6 @@ class Print {
     template< Style s = VERBOSE >
     std::ostream& stream() const noexcept { return s ? m_stream : m_qstream; }
 
-    //! Function object for echoing policies to screen
-    struct echoPolicies {
-      //! Need to store reference to host class whose data we operate on
-      const Print* const m_host;
-      //! Constructor: store host object pointer
-      explicit echoPolicies( const Print* const host ) : m_host( host ) {}
-      //! Function call operator templated on the type that echos a policy
-      template< typename U > void operator()( brigand::type_<U> ) {
-        static_assert( tk::HasTypedef_code_v< typename U::info >,
-                       "Policy code undefined for keyword" );
-        // Print policy code - policy name
-        m_host->raw( m_host->m_item_indent + "   " +
-                     *U::code() + " - " + U::info::name() + '\n' );
-
-      }
-    };
-
     //! Print Inciter header. Text ASCII Art Generator used for executable
     //! names: http://patorjk.com/software/taag, Picture ASCII Art Generator
     //! used for converting the logo text "Quinoa": http://picascii.com.
--- a/src/Control/Inciter/Options/AMRInitial.hpp
+++ b/src/Control/Inciter/Options/AMRInitial.hpp
@@ -67,42 +67,7 @@ class AMRInitial : public tk::Toggle< AMRInitialType > {
             AMRInitialType::INITIAL_CONDITIONS },
           { kw::amr_edgelist::string(), AMRInitialType::EDGELIST },
           { kw::amr_coords::string(), AMRInitialType::COORDINATES } } )
-    {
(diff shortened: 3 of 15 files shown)


jbakosi (changes requested):
Looks good. I have one question/comment at which I just pause this review, because the answer we settle on will determine the further course of action.

> Reviewed 4 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-:-Mo4Z-yN9Du-LgRRmD_9:b6gqj7f)* status: 4 of 15 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/Transporter.cpp, line 292 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-Mo4XlMRELIsAj1fB4ug:-Mo4XlMS7boG5KpfU3Au:b-701xvj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b369837f672d6d9b6af05e3a7249abea1b23966/src/Inciter/Transporter.cpp#L292)):*
> ```cpp
>       print.item( "Initial refinement steps", initref.size() );
>       print.ItemVec< ctr::AMRInitial >( initref );
>       print.ItemVecLegend< ctr::AMRInitial >();
> ```

I am not entirely happy with removing this, because I think the following screen output is pretty useful:
```
  * Mesh refinement (h-ref):
  --------------------------
      ...
      Initial refinement steps                 : 4
      Initial refinement typelist              : idiu
         u - uniform refine
         d - uniform derefine
         i - initial conditions
         e - edge list
         c - coordinates
```
Though these various refinement types are mostly only useful for debugging AMR. I guess we can argue that the test-name also carries this information, but that does not mean the test name (for an old or newly added test) will remain consistent with what the input file is requesting and what the code is attempting to do. The above screen output is at least a feedback on the latter.

As a solution, we _could_ keep keyword::code but remove asserting requiring one for `Problem` and `Physics` classes and remove the uniqueness tests.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/529)* status: 4 of 15 files reviewed, 1 unresolved discussion (waiting on @jbakosi)

___
*[src/Inciter/Transporter.cpp, line 292 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-Mo4XlMRELIsAj1fB4ug:-Mo4_mEQ-cH0yR1MR88F:b-ghvdhi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b369837f672d6d9b6af05e3a7249abea1b23966/src/Inciter/Transporter.cpp#L292)):*

I am not entirely happy with removing this, because I think the following screen output is pretty useful:
```
  * Mesh refinement (h-ref):
  --------------------------
      ...
      Initial refinement steps                 : 4
      Initial refinement typelist              : idiu
         u - uniform refine
         d - uniform derefine
         i - initial conditions
         e - edge list
         c - coordinates
```
Though these various refinement types are mostly only useful for debugging AMR. I guess we can argue that the test-name also carries this information, but that does not mean the test name (for an old or newly added test) will remain consistent with what the input file is requesting and what the code is attempting to do. The above screen output is at least a feedback on the latter.

As a solution, we _could_ keep keyword::code but remove asserting requiring one for `Problem` and `Physics` classes and remove the uniqueness tests.
</blockquote></details>

My reasoning as to why we should remove this:
1. From a discussion we had with Jacob, initial refinement is not supposed to be used with arbitrary combinations of u, d, i, ... That was when we stopped debugging `t0ref`, at a point we had reached something like iiiddd, which passed and iiidddd failed with hanging nodes. Given that such arbitrary combinations are not expected in `t0ref`, we should put in a test that does not allow these; which I will put in soon.
2. Having so much code for only output purposes, that too for initial-refinement *only* is not justified in my opinion; especially now that such arbitrary combinations will not be allowed.

jbakosi (approved):
Looks good, good to merge. Thanks!

> Reviewed 11 of 15 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-:-Mo4dULQ5FEAHLL0C2wP:b-jkezam)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[src/Inciter/Transporter.cpp, line 292 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-Mo4XlMRELIsAj1fB4ug:-Mo4d6zDCXxiDR8a4Zx1:b-is4hma) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b369837f672d6d9b6af05e3a7249abea1b23966/src/Inciter/Transporter.cpp#L292)):*

My reasoning as to why we should remove this:
1. From a discussion we had with Jacob, initial refinement is not supposed to be used with arbitrary combinations of u, d, i, ... That was when we stopped debugging `t0ref`, at a point we had reached something like iiiddd, which passed and iiidddd failed with hanging nodes. Given that such arbitrary combinations are not expected in `t0ref`, we should put in a test that does not allow these; which I will put in soon.
2. Having so much code for only output purposes, that too for initial-refinement *only* is not justified in my opinion; especially now that such arbitrary combinations will not be allowed.
</blockquote></details>

Sounds good. I'm convinced. Let's remove this then.

adityakpandare (commented):
Thanks for the quick review!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/529#-:-Mo4ex6C-sUFF6NONfDI:b-ncgzc2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #531, opened 2021-11-17
Ensure boundary faces are pruned to match user requests

jbakosi:
This fixes a bug, affecting only the node-centered schemes. Had to update some base lines. More info in 7d39c14ae9e1a6d36946beb462fb797e2dacdb9d.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)


adityakpandare (changes requested):
> Reviewed 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 4 of 116 files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpHhTYiEMnrO9WFLFhM:b-mgyy79) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*
> ```cpp
>       // Verify boundarty condition (BC) side sets used exist in mesh file
>       bcs_set = matchBCs( bnode );
>       bcs_set = matchBCs( bface ) or bcs_set;
> ```

This problem was a result of confusing code, and would not have been caused if `matchBCs(bface)` was executed separately and assigned to a separate `bool`. I suggest we do the following:
```
bool bcnode_set = matchBCs(bnode);
bool bcface_set = matchBCs(bface);
bcs_set = bcface_set or bcnode_set;
```
so that it is *explicit* that `matchBCs(bface)` has to run *independent* of whether `bcnode_set` is true or false. The `or` operator is usually commutative, but here, the way the code is written forces it to be non-commutative. I strongly suggest we make this change.
___
*[tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt, line 588 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHh3sT-yGQ1ABFdoG_:-MpHh3sT-yGQ1ABFdoGa:b94tl3g) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b4f0efae0ac9775c99b3085a24884997233b367f/tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt#L588)):*
> ```CMake
>                     TEXT_DIFF_PROG_CONF sod_shocktube_diag.ndiff.cfg
>                                         sod_shocktube_hist.ndiff.cfg
>                                         sod_shocktube_hist.ndiff.cfg
> ```

So we are not testing these hist files anymore?

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 4 of 116 files reviewed, 2 unresolved discussions (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpHm1Vw0ZMOp1R-hGLB:b-gbvwq0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*

This problem was a result of confusing code, and would not have been caused if `matchBCs(bface)` was executed separately and assigned to a separate `bool`. I suggest we do the following:
```
bool bcnode_set = matchBCs(bnode);
bool bcface_set = matchBCs(bface);
bcs_set = bcface_set or bcnode_set;
```
so that it is *explicit* that `matchBCs(bface)` has to run *independent* of whether `bcnode_set` is true or false. The `or` operator is usually commutative, but here, the way the code is written forces it to be non-commutative. I strongly suggest we make this change.
</blockquote></details>

Also, according to Clean code by R. Martin: "Functions should either do something or answer something, but not both. Either it should change the state of an object, or it should return some information about the object" (see http://nicolecarpenter.github.io/2016/03/07/clean-code-chapter-3-functions.html or Bullet 7 in https://www.linkedin.com/pulse/clean-code-summary-chapter3-functions-mahmoud-ibrahim/?articleId=6676386112732176385). The problem that you're trying to fix here is a result of violating this guideline. This function can be rewritten to have an extra output argument `bool`, rather than returning a `bool`.

Regardless, the new code can also have the same problem: if `matchBCs(bface)` is true, `matchBCs(bnode)` will never run.

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 4 of 116 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpI-DtJC0zp4cLWbNdA:b9zh0z9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*

Also, according to Clean code by R. Martin: "Functions should either do something or answer something, but not both. Either it should change the state of an object, or it should return some information about the object" (see http://nicolecarpenter.github.io/2016/03/07/clean-code-chapter-3-functions.html or Bullet 7 in https://www.linkedin.com/pulse/clean-code-summary-chapter3-functions-mahmoud-ibrahim/?articleId=6676386112732176385). The problem that you're trying to fix here is a result of violating this guideline. This function can be rewritten to have an extra output argument `bool`, rather than returning a `bool`.

Regardless, the new code can also have the same problem: if `matchBCs(bface)` is true, `matchBCs(bnode)` will never run.
</blockquote></details>

The `or` operator is always commutative. The problem here was _not_ with commutativity. It was with C/C++ _not_ evaluating the second argument of `or` if the first one was already true, because there is no need to since the result is always true in that case. This is in the C++ as well in the C standard. But I agree that we should make this more explicit: I will update it as you write it.

I like Martin's guideline, but returning the bool as non-const-ref or a return value makes no difference from that viewpoint since the function will still do two things either way. A way to avoid the bool altogether is to evaluate whether the `bface` or `bnode` argument was modified, which, in my opinion, is unnecessarily cumbersome and less performant because I would have to make a copy in the calling code. I don't see another way of adhering to Martin.
___
*[tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt, line 588 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHh3sT-yGQ1ABFdoG_:-MpHy72v96Orm6yNAiTa:bjce3lp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b4f0efae0ac9775c99b3085a24884997233b367f/tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt#L588)):*

So we are not testing these hist files anymore?
</blockquote></details>

Not here. Because after this update I could not get this and the other two tests (compflow_euler_sodshocktube_reord_alecg and compflow_euler_sodshocktube_alecg) that test this same history output already, pass with multiple compilers, I decided to remove them. The output is already tested by those other two tests.

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 3 of 116 files reviewed, 1 unresolved discussion (waiting on @adityakpandare and @jbakosi)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpI4fq02HIL6H4pG2IX:b-p2kuyt) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*

The `or` operator is always commutative. The problem here was _not_ with commutativity. It was with C/C++ _not_ evaluating the second argument of `or` if the first one was already true, because there is no need to since the result is always true in that case. This is in the C++ as well in the C standard. But I agree that we should make this more explicit: I will update it as you write it.

I like Martin's guideline, but returning the bool as non-const-ref or a return value makes no difference from that viewpoint since the function will still do two things either way. A way to avoid the bool altogether is to evaluate whether the `bface` or `bnode` argument was modified, which, in my opinion, is unnecessarily cumbersome and less performant because I would have to make a copy in the calling code. I don't see another way of adhering to Martin.
</blockquote></details>

Yes, I understand that there might not be a better way. And I'm okay with the "explicit" way of calling the two functions separately and then "or"ing them, as in my first comment.
___
*[tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt, line 588 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHh3sT-yGQ1ABFdoG_:-MpI4WlR6U2k1pzCb0LM:bvsti56) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b4f0efae0ac9775c99b3085a24884997233b367f/tests/regression/inciter/compflow/Euler/SodShocktube/CMakeLists.txt#L588)):*

Not here. Because after this update I could not get this and the other two tests (compflow_euler_sodshocktube_reord_alecg and compflow_euler_sodshocktube_alecg) that test this same history output already, pass with multiple compilers, I decided to remove them. The output is already tested by those other two tests.
</blockquote></details>

Okay, thanks for the clarification.

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-:-MpI5JAY-XqdR9mVU-pw:bnfp4nl)* status: 3 of 116 files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpI5D388jNtTnbN59Q4:b-ucfknh) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*

Yes, I understand that there might not be a better way. And I'm okay with the "explicit" way of calling the two functions separately and then "or"ing them, as in my first comment.
</blockquote></details>

Thanks!

jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 3 of 116 files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[src/Inciter/Transporter.cpp, line 592 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/531#-MpHhTYiEMnrO9WFLFhL:-MpI5O0IBo1dpR3Uy4hM:bfmw0qe) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5101a96e860866924ec32331f80a4c6369af2579/src/Inciter/Transporter.cpp#L592)):*

Thanks!
</blockquote></details>

Of course. This was obscure code as it was written and it is much better now.

adityakpandare (approved):
> Reviewed 1 of 113 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/531)* status: 5 of 116 files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #533, opened 2021-12-13
Removing riemann factory

adityakpandare:
This commit removes the Riemann factory, and replaces it with simpler code. This was done because the Riemann factory was not necessary, since separate implementations are anyway required across discretizations and PDEs being solved. The proposed implementation uses a simple if-test *only executed at the constructor* of the DG-PDE, i.e. not at every time-step but only at the initialization. Then, a `std::function` pointer to the relevant function (Riemann solver) is stored in the DG-PDE class, and passed to required functions that compute the surface integral.

---
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/533)


WeizhaoLi2018 (approved):
Do we need to specify the type of the riemann solver in the control file?
Also why don't we have same setup for Transport class? Is it because transport only use Upwind and it's set default? I didn't see this kind of default setting.

> Reviewed 16 of 16 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/533#-:-Mqun3fs4ZvWZwDCEvlP:b3n9ptl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

adityakpandare (commented):
Yes, we still need to specify the riemann solver in the control file; nothing has changed with that.
For transport (i.e. advection) we can only do upwinding at this point; we do not have any other options to choose from. Hence we do not have this setup for transport. If we want to experiment with other advection schemes, we will have to add a similar setup for transport too.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/533#-:-MquoVqK59jArcqjR_Jx:b-6ouz0t)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

WeizhaoLi2018 (approved):
Got it! Thanks!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/533#-:-MquvrimEmN-qL489CbK:bdxx1kl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #564, opened 2022-10-12
bugfix: support integration with ExaM2M transfer every iteration

ericjbohm:
This revised the callback scheme to be compatible with recent changes to ExaM2M.  ExaM2M will trigger a single callback to the destination mesh when the solution transfer is complete. Discretization will set that callback to be transferCompleteCBinDest. That will call the callback passed in by the destination mesh solver in the transfer call, and call transferCompleteNotifyFromDest in the Discretization source mesh, which will call the callback passed in by the source mesh solver.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/564)

### DIFF
--- a/src/Inciter/ALECG.cpp
+++ b/src/Inciter/ALECG.cpp
@@ -148,21 +148,8 @@ ALECG::ALECG( const CProxy_Discretization& disc,
   thisProxy[ thisIndex ].wait4meshblk();
 
   // Generate callbacks for solution transfers we are involved in
+  d->comfinal();
 
-  // Always add a callback to be used when we are not involved in any transfers
-  std::vector< CkCallback > cb;
-  auto c = CkCallback(CkIndex_ALECG::transfer_complete(), thisProxy[thisIndex]);
-  cb.push_back( c );
-
-  // Generate a callback for each transfer we are involved in (either as a
-  // source or a destination)
-  auto meshid = d->MeshId();
-  for (const auto& t : d->Transfers())
-    if (meshid == t.src || meshid == t.dst)
-      cb.push_back( c );
-
-  // Send callbacks to base
-  d->transferCallback( cb );
 }
 //! [Constructor]
 
@@ -643,7 +630,8 @@ ALECG::box( tk::real v, const std::vector< tk::real >& blkvols )
   volumetric( m_u, Disc()->Vol() );
 
   // Initiate IC transfer (if coupled)
-  Disc()->transfer( m_u );
+
+  Disc()->transfer( m_u ,  CkCallback(CkIndex_ALECG::transfer_complete(), thisProxy[thisIndex]));
 
   // Initialize nodal mesh volumes at previous time step stage
   d->Voln() = d->Vol();
@@ -1393,9 +1381,13 @@ ALECG::transfer()
 // *****************************************************************************
 {
   // Initiate solution transfer (if coupled)
-  //Disc()->transfer( m_u,
-  //  CkCallback(CkIndex_ALECG::stage(), thisProxy[thisIndex]) );
+
+#ifdef HAS_EXAM2M
+  Disc()->transfer( m_u, CkCallback(CkIndex_ALECG::stage(), thisProxy[thisIndex]));
+#else
   thisProxy[thisIndex].stage();
+#endif
+
 }
 
 //! [stage]
--- a/src/Inciter/Discretization.cpp
+++ b/src/Inciter/Discretization.cpp
@@ -190,7 +190,7 @@ Discretization::Discretization(
     if (thisIndex == 0) {
       exam2m::addMesh( thisProxy, m_nchare,
         CkCallback( CkIndex_Discretization::transferInit(), thisProxy ) );
-      std::cout << "Disc: " << m_meshid << " m2m::addMesh()\n";
+      //      std::cout << "Disc: " << m_meshid << " m2m::addMesh()\n";
     }
     #else
     transferInit();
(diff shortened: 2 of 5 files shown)


adityakpandare (commented):
Requesting @jbakosi's review too.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-:-NEaO8rT9MBWks_9XiMJ:b-pqeaav)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @jbakosi)

jbakosi (changes requested):
This looks a lot cleaner design than the one I originally did with communicating the callbacks in advance, which is now eliminated. Let's see if I understand:

There is now no need for the initial communication of the transfer callbacks, so in the `ALECG` ctor we simply call `Discretization::comfinal()`, which used to be the end-point of the callback-communication and which, after initializing `Discretization::m_mytransfer` (a subset of `m_transfer`), ends up reducing to `Transporter::comfinal()`.

`Discretization::transfer()` now takes a callback to the function that needs to be called when the transfer is complete on the given solver. This is `ALECG::transfer_complete()` during inititalization and `ALECG::stage()` during timestepping. This is stored in `Discretization::solver_transfer_complete_CB`.

Interfacing with ExaM2M now only requires passing a single callback and only on the destination mesh and this is always `Discretization::transferCompleteCBinDest()`. This function is called by ExaM2M when the transfer is complete.

Since `Discretization::transferCompleteCBinDest()` is always only called on a solver that was a destination to a transfer, it needs to looks up the corresponding source and let it know that the transfer is complete, which is done by calling Discretization::transferCompleteNotifyFromDest() of the source which then calls whatever is stored in its `Discretization::solver_transfer_complete_CB`. This eliminates the need to have to communicate callbacks.

I think this is great! I assume this works with the simple test-case we used to test this with and you can take a few time steps without dead-locks. It would be great to add two regression tests that exercises this simple case: one that only runs the initialization, e.g., setting the ICs without timestepping, and another one that also takes a 3-4 time steps. @adityakpandare does the CI test with ExaM2M?

I also have a couple of more location-specific comments and suggestions. See below. Otherwise looks good.

> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-:-NFo0BXJEBG98DaDmA3Y:b-up6x0u)* status: all files reviewed, 11 unresolved discussions (waiting on @ericjbohm)

___
*[`src/Inciter/Discretization.hpp` line 132 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo5Z7b2fTh4uTBxhxk:-NFo5Z7b2fTh4uTBxhxl:b-cejam2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L132)):*
> ```cpp
>     //!   callback previously exchanged in comcb during setup.  This callback
>     //!   is specific to the  transfer and will be used by both source and dest
>     //!   meshes to signal the completion of the transfer operation.
> ```

When doxygen parses these comments, it will get confused because the line that starts with 'Start solution transfer' is also a '\brief' by definition in doxygen. It would be better to document this in the function definition in Discretization.cpp, see below.
___
*[`src/Inciter/Discretization.hpp` line 138 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo7Bby3UuK4rbbMDmV:-NFo7Bby3UuK4rbbMDmW:b-fln6w3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L138)):*
> ```cpp
>     //! \brief Called by ExaM2M when the destination mesh completes
>     //!  triggers callback passed in by destination solver
>     //!  triggers call to discretization source
> ```

Same here. I would keep the first line 'Solution transfer ...' and move the rest of the comments to the definition, see below.
___
*[`src/Inciter/Discretization.hpp` line 143 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo8FrW2pjmX2on40Hg:-NFo8FrW2pjmX2on40Hh:b-fln6w3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L143)):*
> ```cpp
>     //! Solution transfer completed
>     //! \brief Called into source by destination when transfer completes
>     //!  triggers callback passed in by solver
> ```

Same here. I would keep the first line 'Solution transfer ...' and move the rest of the comments to the definition, see below.
___
*[`src/Inciter/Discretization.hpp` line 438 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo8ZosEogEyGUMwQPp:-NFo8ZosEogEyGUMwQPq:b-tz1knj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L438)):*
> ```cpp
>       p | m_meshid;
>       p | m_transfer_complete;
>       p | solver_transfer_complete_CB;
> ```

To make it consistent with the rest of the code, I would rename this to `m_solver_transfer_complete_cb`.
___
*[`src/Inciter/Discretization.hpp` line 509 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo8ixK701zpx1765CB:-NFo8ixK701zpx1765CC:b-s7izfr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L509)):*
> ```cpp
>     //! Solution/mesh transfer (coupling) information
>     //! \details This has the same size with the same src/dst information on
>     //!   all solvers.
> ```

These comments are now stale. Let's change them to:
```
//! \brief Charm++ callback of the function to call after a mesh-to-mesh
//!   solution transfer is complete
```
___
*[`src/Inciter/Discretization.hpp` line 519 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo9AR0EdFjy788XEgS:-NFo9AR0EdFjy788XEgT:b-ijmpuu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.hpp#L519)):*
> ```cpp
>     //! \details This is a subset of m_transfer, holding only those entries
>     //!   that this solver is involved in (either a source or a destination).
>     std::vector< Transfer > m_mytransfer;
> ```

As the source is searched in `Discretization::transfer()` (see below) using only `m_transfer`, this is now probably not needed. However, ..., see below.
___
*[`src/Inciter/Discretization.cpp` line 332 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo9Q0AB3XUFQYgw5Yd:-NFo9Q0AB3XUFQYgw5Ye:b-qyav84) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L332)):*
> ```cpp
>   //    }
>   //    std::cout << '\n';
>   //}
> ```

Basically no change here. But I don't believe `m_mytransfer` is now necessary. So we could probably do completely away with `Discretization::comfinal()` and reduce to `Transporter::comfinal()` directly from the `ALECG` ctor using the code below, slightly modified, i.e.,
```cpp
  // Signal the runtime system that the workers have been created
  std::vector< std::size_t > meshdata{ /* initial */ 1,  d->MeshId() };
  contribute( meshdata, CkReduction::sum_ulong,
    CkCallback(CkReductionTarget(Transporter,comfinal), d->Tr()) );
```
___
*[`src/Inciter/Discretization.cpp` line 343 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo6ZORACixb3x4rKFX:-NFo6ZORACixb3x4rKFY:b5r1133) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L343)):*
> ```cpp
> // *****************************************************************************
> //  Start solution transfer (if coupled)
> //! \param[in] u Solution to transfer from/to
> ```


```
    //! \param[in] cb This passed in callback is how we notify the solver associated
    //!   with this Discreziation's mesh.  If we are the destination of the
    //!   transfer, we will also have to notify the Discretization source mesh. This callback
    //!   is specific to the transfer and also to the source and destination and is used by both source and dest
    //!   meshes to signal the completion of the transfer operation.
```
___
*[`src/Inciter/Discretization.cpp` line 385 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo7KK0Ey7ReTkLGpyw:-NFo7KK0Ey7ReTkLGpyx:bpu4zr4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L385)):*
> ```cpp
> //! Called by destination when transfer completes
> //! triggers callback passed in by solver
> //! triggers call to discretization source
> ```

```
    //  Solution transfer completed CB
    //! \details This is called by ExaM2M on the destination mesh when the transfer completes.
    //!   Since this is called only on the destination, we find and notify the corresponding source
    //!   of the completion.
```
___
*[`src/Inciter/Discretization.cpp` line 401 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFoAlsf0d5UkuTrQvoH:-NFoAlsf0d5UkuTrQvoI:b-85h5ye) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L401)):*
> ```cpp
>   }
>   //! callback to the solver that initiated the transfer
>   solver_transfer_complete_CB.send();
> ```

I am thinking that while `transferCompleteCBinDest()` is only supposed to be called on the destination, so `solver_transfer_complete_CB.send()` at the bottom is only called on the destination, it would read cleaner and perhaps would be less error-prone if we wrote this "symmetrically":
```cpp
void Discretization::transferCompleteCBinDest() {
  for (auto& t : m_transfer) {
    // If we are the source, complete transfer
    if (m_meshid == t.src) {
      solver_transfer_complete_CB.send();
    }
    // If we are the destination, notify the source of completion
    // (which calls its own solver_transfer_complete_CB)
    else if (m_meshid == t.dst) {
      m_disc[ t.src ][ thisIndex ].transferCompleteNotifyFromDest();
    }
  }
}
```
I assume the original code here worked fine for the simple test case we used together to test this with. So before I make this change,  I would put in a regression test using the original code and verify that my thinking here is correct.
___
*[`src/Inciter/Discretization.cpp` line 408 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFo8JoN6JcdQzaDRBog:-NFo8JoN6JcdQzaDRBoh:b-fjcg31) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L408)):*
> ```cpp
> //! Solution transfer completed
> //! Called into source by destination when transfer completes
> //! triggers callback passed in by solver in the source mesh
> ```

    //  Solution transfer completed
    //! \details Called on the source only by the destination when a transfer step completes.

jbakosi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/564)* status: all files reviewed, 11 unresolved discussions (waiting on @ericjbohm)

___
*[`src/Inciter/Discretization.cpp` line 401 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-NFoAlsf0d5UkuTrQvoH:-NFoGtyx1m8KcZf0OGOb:b-xmayo1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/2f53d3ade04898c4cb2d05d0dbb9c40ee4d226d0/src/Inciter/Discretization.cpp#L401)):*

I am thinking that while `transferCompleteCBinDest()` is only supposed to be called on the destination, so `solver_transfer_complete_CB.send()` at the bottom is only called on the destination, it would read cleaner and perhaps would be less error-prone if we wrote this "symmetrically":
```cpp
void Discretization::transferCompleteCBinDest() {
  for (auto& t : m_transfer) {
    // If we are the source, complete transfer
    if (m_meshid == t.src) {
      solver_transfer_complete_CB.send();
    }
    // If we are the destination, notify the source of completion
    // (which calls its own solver_transfer_complete_CB)
    else if (m_meshid == t.dst) {
      m_disc[ t.src ][ thisIndex ].transferCompleteNotifyFromDest();
    }
  }
}
```
I assume the original code here worked fine for the simple test case we used together to test this with. So before I make this change,  I would put in a regression test using the original code and verify that my thinking here is correct.
</blockquote></details>

No, I'm not suggesting good code here. More like:
```cpp
oid Discretization::transferCompleteCBinDest() {
  for (auto& t : m_transfer) {
    if (m_meshid == t.dst) {
      solver_transfer_complete_CB.send();
      // If we are the destination, notify the source of completion
      // (which calls its own solver_transfer_complete_CB)
      m_disc[ t.src ][ thisIndex ].transferCompleteNotifyFromDest();
    }
  }
}
```

adityakpandare (commented):
Created a commit in the Quinoa repo, mimicking @ericjbohm's changes: https://github.com/quinoacomputing/quinoa/commit/d2db0b9b36fde6e4f1c7c3a7b96b9b19d9978497.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/564#-:-NQ2C9io2dyH1qge90Js:b-bywozy)* status: all files reviewed, 11 unresolved discussions (waiting on @ericjbohm)

adityakpandare:
See https://github.com/quinoacomputing/quinoa/pull/576

PR #580, opened 2023-06-08
Mat based storage

adityakpandare:
@WeizhaoLi2018, could you confirm that there was a bug in the for-loop index in `consistentMultiMatLimiting_P1()`? This change requires me to update multimat baselines, since they were incorrect.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580)

### DIFF
--- a/src/PDE/Limiter.cpp
+++ b/src/PDE/Limiter.cpp
@@ -899,8 +899,10 @@ VertexBasedMultiMat_FV(
     }
     else
     {
-      if (!g_inputdeck.get< tag::discr, tag::accuracy_test >())
-        consistentMultiMatLimiting_P1(nmat, rdof, e, U, P, phic, phip);
+      if (!g_inputdeck.get< tag::discr, tag::accuracy_test >()) {
+        std::vector< tk::real > phic_p2(ncomp, 1.0);
+        consistentMultiMatLimiting_P1(nmat, rdof, e, U, P, phic, phic_p2);
+      }
     }
 
     // apply limiter function
@@ -1577,10 +1579,10 @@ void consistentMultiMatLimiting_P1(
   else
   {
     // same limiter for all volume-fractions
-    for (std::size_t k=volfracIdx(nmat, 0); k<volfracIdx(nmat, nmat); ++k)
+    for (std::size_t k=0; k<nmat; ++k)
       phic_p1[volfracIdx(nmat, k)] = phi_al_p1;
     if(rdof > 4)
-      for (std::size_t k=volfracIdx(nmat, 0); k<volfracIdx(nmat, nmat); ++k)
+      for (std::size_t k=0; k<nmat; ++k)
         phic_p2[volfracIdx(nmat, k)] = phi_al_p2;
   }
 }
--- a/src/PDE/MultiMat/MultiMatIndexing.hpp
+++ b/src/PDE/MultiMat/MultiMatIndexing.hpp
@@ -24,31 +24,31 @@ namespace inciter {
 //! \param[in] kmat Index of required material
 //! \return Index of the required material volume fraction
 inline std::size_t volfracIdx( std::size_t /*nmat*/, std::size_t kmat )
-{ return kmat; }
+{ return (3 + 3*kmat); }
 
 //! Get the index of the required material continuity equation
-//! \param[in] nmat Number of materials
+// //! \param[in] nmat Number of materials
 //! \param[in] kmat Index of required material
 //! \return Index of the required material continuity equation
-inline std::size_t densityIdx( std::size_t nmat, std::size_t kmat )
-{ return (nmat+kmat); }
+inline std::size_t densityIdx( std::size_t /*nmat*/, std::size_t kmat )
+{ return (3 + 3*kmat+1); }
 
 //! Get the index of the required momentum equation component
-//! \param[in] nmat Number of materials
+// //! \param[in] nmat Number of materials
 //! \param[in] idir Required component direction;
 //!   0: X-component,
 //!   1: Y-component,
 //!   2: Z-component.
 //! \return Index of the required momentum equation component
-inline std::size_t momentumIdx( std::size_t nmat, std::size_t idir )
-{ return (2*nmat+idir); }
+inline std::size_t momentumIdx( std::size_t /*nmat*/, std::size_t idir )
+{ return idir; }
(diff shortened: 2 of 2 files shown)


WeizhaoLi2018 (commented):
@adityakpandare I am confused here. Do you mean the line:

```c++
for (std::size_t k=volfracIdx(nmat, 0); k<volfracIdx(nmat, nmat); ++k)
```

is not identical to

```c++
for (std::size_t k=0; k<nmat; ++k)
```

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-:-NXQjPz_6a9NPci9km4E:bdlxfi8)* status: 0 of 2 files reviewed, all discussions resolved

WeizhaoLi2018 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/PDE/Limiter.cpp` line 904 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-NXQkg1PB8ex_AiLizzN:-NXQkg1PB8ex_AiLizzO:b8q5yeg) ([raw file](https://github.com/quinoacomputing/quinoa/blob/8b293f3cf7d7725d44207a8be41c4d3a7e1acc64/src/PDE/Limiter.cpp#L904)):*
> ```cpp
>       if (!g_inputdeck.get< tag::discr, tag::accuracy_test >()) {
>         std::vector< tk::real > phic_p2(ncomp, 1.0);
>         consistentMultiMatLimiting_P1(nmat, rdof, e, U, P, phic, phic_p2);
> ```

I do think this is a bug. We don't need `phip` for consistent limiting right now and that argument is used to pass the limiting coefficients for third order solution coefficients.

WeizhaoLi2018 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/MultiMat/MultiMatIndexing.hpp` line 27 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-NXQlyCN7qI8AECeaEdX:-NXQlyCN7qI8AECeaEdY:bgy001a) ([raw file](https://github.com/quinoacomputing/quinoa/blob/8b293f3cf7d7725d44207a8be41c4d3a7e1acc64/src/PDE/MultiMat/MultiMatIndexing.hpp#L27)):*
> ```cpp
> //! \return Index of the required material volume fraction
> inline std::size_t volfracIdx( std::size_t /*nmat*/, std::size_t kmat )
> { return (3 + 3*kmat); }
> ```

If my understanding is correct, this should be
```c++
{ return (3 + 3*nmat); }
```
Is this the bug you are trying to find?

adityakpandare (commented):
Yes I am talking about that line. The previous for-loop looped `k` over the `volfracIdx`, but then passed this `k` into `volfracIdx` again to index into `phic`. That was a bug, because that is incorrectly index into `phic`. So it should either be a loop over `0` to `nmat-1` and then use `volfracIdx` to index into `phic`, or be a loop over `volfracIdx` and use that `k` it directly to index into `phic`.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-:-NXQplo5El5Pq_IOlYI4:b-l2jd2f)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @WeizhaoLi2018)

___
*[`src/PDE/Limiter.cpp` line 904 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-NXQkg1PB8ex_AiLizzN:-NXQqaCA1zFpMjznJeoW:bp9swqc) ([raw file](https://github.com/quinoacomputing/quinoa/blob/8b293f3cf7d7725d44207a8be41c4d3a7e1acc64/src/PDE/Limiter.cpp#L904)):*

I do think this is a bug. We don't need `phip` for consistent limiting right now and that argument is used to pass the limiting coefficients for third order solution coefficients.
</blockquote></details>

Yes, hence fixed it.
___
*[`src/PDE/MultiMat/MultiMatIndexing.hpp` line 27 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-NXQlyCN7qI8AECeaEdX:-NXQqhYxEDShQR8I17OP:b-bp5eoj) ([raw file](https://github.com/quinoacomputing/quinoa/blob/8b293f3cf7d7725d44207a8be41c4d3a7e1acc64/src/PDE/MultiMat/MultiMatIndexing.hpp#L27)):*

If my understanding is correct, this should be
```c++
{ return (3 + 3*nmat); }
```
Is this the bug you are trying to find?
</blockquote></details>

No, this is not a bug. This is correct as it is. The current unknown vector stores vars as follows:
```
mtm-x
mtm-y
mtm-z
alpha_1
alpha rho_1
alpha rho E_1
...
alpha_k
alpha rho_k
alpha rho E_k
...
alpha_n
alpha rho_n
alpha rho E_n
```
We are trying to find volume fraction of material `kmat`. So the index for alpha_k will be `3+3*kmat`.

adityakpandare (commented):
What I mean is, there was a bug in the previous version of the for-loop in `consistentMultiMatLimiting_P1()`. Due to that bug, our multimat baselines were incorrect. Now that I have fixed that bug, the baselines will have to be updated. I just wanted you to confirm that the previous for-loop was indeed buggy.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-:-NXQuzQm1GRw-qNaznc2:bbl3jsl)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @WeizhaoLi2018)

WeizhaoLi2018 (commented):
I see. It indeed is a bug. 
But I am still not sure why it will cause any difference in the regression testcases. Since k is equal to the return value of volfracIdx theoretically.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/580#-:-NXQudUaDiTPZThPubWK:b-mtba2b)* status: 0 of 2 files reviewed, 1 unresolved discussion

adityakpandare:
Closing this for now. Will follow up on this later.

PR #582, opened 2023-06-16
Updates about refinement layer algorithm and communications

WeizhaoLi2018:
This PR includes the changes to refinement layer for p-adaptive scheme. The new functions will generate more smooth transition between elements with different polynomial orders. Some `m_ndof` communications are removed since they are useless for the whole communication scheme.

This PR changes the baseline for pdg_gauss_hump and pdg_sedov_blast. I am running both cases on ARC. If the results look fine, I will update the baseline.

I will create PR about changes related with multi-material once we merged this changes since two PRs might be good for reviewing as well as for me to track all the changes in baseline testcase.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582)


adityakpandare (changes requested):
@WeizhaoLi2018, the code doesn't compile on the CI server. The error is:
```
src/Inciter/DG.cpp: In member function ‘void inciter::DG::comreco(int, const std::vector<long unsigned int>&, const std::vector<std::vector<double> >&, const std::vector<std::vector<double> >&)’:
src/Inciter/DG.cpp:742:13: error: ‘ndof’ was not declared in this scope; did you mean ‘tag::ndof’?
Assert( ndof.size() == tetid.size(), "Size mismatch in DG::comreco()" );
```

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-:-NY50N3C0EbLiPZfwqMD:b4dijid)* status: 0 of 3 files reviewed, all discussions resolved

adityakpandare (changes requested):
Thanks, @WeizhaoLi2018. I have a couple of questions about the ndof-smoothening algorithms below. If you clarify those, I will make the required changes. Otherwise looks good!

> Reviewed 6 of 61 files at r2, 5 of 5 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-:-NoNMZhO0pZdxzA8mTzm:b-ptx9gc)* status: 11 of 61 files reviewed, 3 unresolved discussions (waiting on @WeizhaoLi2018)

___
*[`src/Inciter/DG.cpp` line 1146 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-NoNKeTeCnyabLJgbvP1:-NoNKeTfAiZPKSNAxdx3:b-wg2iu2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/90ae47adad8669771b672beca481d7f5ca68c3c7/src/Inciter/DG.cpp#L1146)):*
> ```cpp
>       const auto& pesup = tk::cref_find(myGhosts()->m_esup, ip);
>       for(auto er : pesup)
>         node_ndof[ip] = std::max(m_ndof[er], node_ndof[ip]);
> ```

Isn't putting this inside an element loop unnecessary? Why not just loop over all points directly, and do the max calculation at each point using the same `m_esup` structure?
___
*[`src/Inciter/DG.cpp` line 1174 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-NoNLtCLC5_hNph_TRiT:-NoNLtCM5lr4hHTpyV5E:bbdvtvr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/90ae47adad8669771b672beca481d7f5ca68c3c7/src/Inciter/DG.cpp#L1174)):*
> ```cpp
>     else if(counter_p1 > 0 && m_ndof[e] < 4)
>       m_ndof[e] = 4;
>   }
> ```

Similar question about the loop above: Why not just loop over all points, and then within that loop over all elements surrounding that point using `m_esup` and then compute the value of `m_ndof[e]`? This way, we can actually condense both the point-loops (this one, and the one above with my comment). Am I missing something here?
___
*[`src/Inciter/DG.cpp` line 1220 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-NoNMShf9e0niYun4ulj:-NoNMShf9e0niYun4ulk:b2d5qhm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/90ae47adad8669771b672beca481d7f5ca68c3c7/src/Inciter/DG.cpp#L1220)):*
> ```cpp
>     else if(counter_p1 == 4 && m_ndof[e] == 1)
>       m_ndof[e] = 4;
>   }
> ```

Same questions for this function as above.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582)* status: 10 of 61 files reviewed, all discussions resolved

___
*[`src/Inciter/DG.cpp` line 1146 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-NoNKeTeCnyabLJgbvP1:-NoNfWus4a4BhyyKL8vd:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/90ae47adad8669771b672beca481d7f5ca68c3c7/src/Inciter/DG.cpp#L1146)):*

Isn't putting this inside an element loop unnecessary? Why not just loop over all points directly, and do the max calculation at each point using the same `m_esup` structure?
</blockquote></details>

Done.
___
*[`src/Inciter/DG.cpp` line 1174 at r4](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-NoNLtCLC5_hNph_TRiT:-NoNfYiK6gBlVhGWKHsu:b-l5xias) ([raw file](https://github.com/quinoacomputing/quinoa/blob/90ae47adad8669771b672beca481d7f5ca68c3c7/src/Inciter/DG.cpp#L1174)):*

Similar question about the loop above: Why not just loop over all points, and then within that loop over all elements surrounding that point using `m_esup` and then compute the value of `m_ndof[e]`? This way, we can actually condense both the point-loops (this one, and the one above with my comment). Am I missing something here?
</blockquote></details>

This is not equivalent to what is currently implemented (and what this algo is intended to do) since the value of `m_ndof` is changed at the end of the loop, and that would affect the next loop-iteration. So please ignore this.

adityakpandare (approved):
Looks good to merge :lgtm:

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582#-:-NoNhcG721iQn5taaWZ_:bb017xj)* status: 11 of 61 files reviewed, all discussions resolved

adityakpandare (approved):
> Reviewed 50 of 61 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/582)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @WeizhaoLi2018)

PR #576, opened 2023-03-08
bugfix: support integration with ExaM2M transfer every iteration

adityakpandare:
This pull request mirrors https://github.com/quinoacomputing/quinoa/pull/564, creating the commits in [this repo](https://github.com/quinoacomputing/quinoa/) instead of the [fork](https://github.com/ericjbohm/quinoa/tree/exam2m-callback-in-transfer).

Changes by @ericjbohm:
This revised the callback scheme to be compatible with recent changes to ExaM2M.  ExaM2M will trigger a single callback to the destination mesh when the solution transfer is complete. Discretization will set that callback to be transferCompleteCBinDest. That will call the callback passed in by the destination mesh solver in the transfer call, and call transferCompleteNotifyFromDest in the Discretization source mesh, which will call the callback passed in by the source mesh solver.

PR #584, opened 2023-07-28
Progress towards removing system of systems

jbakosi:
Note that this PR is targeting branch _exam2m-iterative_, because it builds on that line of work and that should be merged first. Once #576 is merged, the target branch should be changed to _develop_.

This changes parsing for multiple meshes from
```
compflow ... end
compflow ... end
couple ... end
```
to
```
compflow
  mesh
    filename "mesh1.exo"
    filename "mesh2.exo"
  end
end
```
and related consequences/details, such as

1. `depvar` is now removed from flow blocks, but not (yet) from transport blocks. In transport `depvar` is still there because I wanted to reduce the number of changed files and a bunch of regression tests baselines would have needed to be updated only because the name of the now automatically assigned dependent variable changes.

2. The PDE constructors, `CGPDE`, `FVPDE`, etc., now do _not_ take a 'c' argument that they internally have stored as `m_system`. However, I have not fully removed `m_system` from inside of them, to leave the places more visible to removing them completely in the next step. The constructors simply fill `m_system` to 0 for now. Removing `m_system` will/should involve removing a layer of vector from many/some of the containers in `g_inputdeck`, and since that is tied to parsing and PEGTL grammar, it is best to do that in a next, separate step.

3. The mesh2mesh coupling machinery in `Discretization` does not really change with this commit, again, due to keeping the changes manageable. We can revamp that next, in light of the simplified design that the first mesh is always the background/source and the subsequent ones are always destination meshes, which this change set prepares.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/584)


jbakosi (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/584)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Control/Inciter/Types.hpp` line 342 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/584#-NbFh81y4ck4Xx22AOQZ:-NbYEA_K940TKYm1rN4J:b178g9m) ([raw file](https://github.com/quinoacomputing/quinoa/blob/30a08de0093f387f65ac356005a52a9a4b154d72/src/Control/Inciter/Types.hpp#L342)):*

Does this mean that mesh location/orientation functionality is no more supported? Is there a way to keep that functionality?
</blockquote></details>

I removed it temporarily. We will put it back later.
___
*[`src/PDE/ConfigureCompFlow.cpp` line 129 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/584#-NbFnlfuDMX0yQPUJ0_V:-NbYEWPk3duJD6mOl8Bs:byipdjy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/b961a52f2bfa2d8a2f38f21069b1643f00dbaf10/src/PDE/ConfigureCompFlow.cpp#L129)):*

Is this code commented so that it can be fixed in the next round of changes involving removal of "systems"?
</blockquote></details>

It is commented because I did not want to deal with this because the semantics of what/how these nested vectors store their contents related to systems/meshes. Will re-enable later.

adityakpandare (approved):
:lgtm: Thanks @jbakosi!

> Reviewed 2 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/584#-:-NcXhF1y6EgINo5JOv10:b3iylw6)* status: 139 of 143 files reviewed, all discussions resolved (waiting on @jbakosi)

jbakosi (commented):
This passes all tests. Merge this, cancel #576. Thanks for the help with this.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/584#-:-NcY6M244i2vW0w-CaL0:b5zkl5j)* status: 137 of 150 files reviewed, all discussions resolved (waiting on @adityakpandare)

adityakpandare (commented):
> Reviewed 4 of 6 files at r2, 7 of 9 files at r3, 3 of 35 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/584)* status: 142 of 176 files reviewed, all discussions resolved

jbakosi (commented):
Excellent!

adityakpandare:
This is merged into `develop`.

PR #591, opened 2023-08-24
Remove equation systems

jbakosi:
_Note: targeting branch 'system1', which should be merged first._

This touches a lot of, but mostly straightforward, code.

This removes systems. Let me know if I missed something.

Another potential, hopefully final step, is to remove systems from the parser grammar, but the vectors originally intended for systems can now be repurposed for solvers operating on different meshes. Let's discuss any pros and cons of this and/or whether we should simply reuse the existing vectors in the inputdeck parser grammar or redesign that a bit. My vote is to use it as is for now, and a better use of time is to replace PEGTL with LUA eventually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591)


jbakosi:
It just came into my mind: Shouldn't we keep all this for passing down the meshid instead?

adityakpandare (commented):
"all this" as in? the systems?

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591#-:-Nch9phC5dkBEacKYZr1:beil91i)* status: 0 of 102 files reviewed, all discussions resolved

jbakosi:
"All this" meaning basically canceling this PR. No? We could rename 'system' to meshid but that's it. Otherwise, how will we pass down the meshid to all these functions? I think that will be necessary soon for m2m problems, no? We could reintroduce that with a last function arg defaulting to 0. Will that not be needed for pretty much all code this PR touches?

adityakpandare (commented):
My understanding from our previous discussion was: there's a separate discretization object for each mesh. If that's the case, each mesh will have its own solution vector, and there will be no need of the system arg. Isn't this true?

> Reviewed 1 of 102 files at r1.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591#-:-NchEKvK9OAz5xhdvq0B:b2zfjlq)* status: 1 of 102 files reviewed, all discussions resolved

jbakosi:
It is tru that that's the design and that would be true on the surface of storing amd passing this data down to the operators, but look at the diff of this PR and see what that system arg has been used for: mostly to look up values from inputdeck of which there is a single copy on each PE, storing global config data for all meshes. Surely at the lowest levels the different solvers will need to know what mesh they operate on. No?

adityakpandare (commented):
Okay so the question is whether we need to pass the "system" arg to query the values from the inputdeck. I'd say no, since now there's no support for multiple systems. This is because all meshes are stored in a single system. So even if the vectors in the inputdeck grammar are kept, the index that we will be querying  from these "system-vectors"  will be 0. So rather than passing a "fake" index (which is necessarily 0) to all these functions, why don't we just hard-code this 0 where the queries actually happen (I'm guessing that's in the inputdeck parser). Does that make sense? This way, it's also fool-proof- devs will not use the system arg and run into a segfault.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591#-:-NchKY3x2BNr8az9FLpQ:bf09ex7)* status: 1 of 102 files reviewed, all discussions resolved

jbakosi:
I'm talking about possibly keeping the vectors in the inputdeck but changing the semantics of what it contains, i.e., not for system index but for meshid. Would that be useful in the immediate future? At first I thought it might be useful, but I'm now struggling to find a use-case so it is probably not worth keeping it.

For example, I thought about a need to look up a different way of setting user-defined ICs dependent on which mesh/solver we are initializing, but that does not make much sense. Another example might be when needing to use a different material, e.g., gamma, depending on which solver (mesh) is being solved. But that also does not make sense because the fluid solved with all meshes will be the same. Will it always be?  (I'm considering the simplest initial use-case of coupling here with rigid bodies embedded in a background flow where the flow around each object is the same fluid.) Basically, I'm wondering if it makes sense to plan for a case where anything that is needed from the inputdeck would depend on the meshid during a run. Since the inputdeck is meshid-agnostic (it has data on all meshids), it is the same copy on all PEs and for all solvers.

But I think it is probably the best to eliminate this, which simplifies code, and reintroduce whatever is needed when/if needed.

adityakpandare (approved):
To qualify my argument a bit more: the places where "system" is used in the code you've removed is for stuff like reading material properties, other parameters (viz. interface-sharpening, pressure-relaxation), and for box/block initialization values of pressure etc. These will not vary between meshes anymore, since all meshes are under one system, hence all such properties/parameters will be the same for all meshes. Thus, using a "system" argument would be superfluous, since its value would always be 0.

About your questions:
1. Materials for different meshes: Since we use a "diffuse interface formulation", all materials are present in the entire domain. This means, that even if a mesh has to be initialized with material B, whereas the background mesh contained material A, both meshes will have to know about both materials A & B. This is just how the multi-material formulation works, and has nothing to do with separate meshes. That being said, even initializing one mesh with mat-A and the other with mat-B is a terrible idea numerically, since the interface capturing would give up advecting the interface at difference mesh resolutions. So this is not a use case. So although multimaterial problems are of interests, the way to initialize them on overset grids is not one material per mesh.
2. Whether we want to eliminate vectors from the inputdeck types: I agree that they should be removed. Whether now or later, that depends on what you think is easier: taking this one-step at a time and merging your current changes first, followed by removing inputdeck-vectors; or doing it all at once.

I hope this makes sense!

> Reviewed 101 of 102 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591#-:-NchPNaf-dX-yinLmfpb:b6u12ly)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbakosi)

___
*[`src/PDE/CompFlow/Problem/FieldOutput.cpp` line 41 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/591#-NchbR0I4AZ-OdUJkpwX:-NchbR0I4AZ-OdUJkpwY:b-86v91h) ([raw file](https://github.com/quinoacomputing/quinoa/blob/a3261f140251c918658474e45707084182975eb0/src/PDE/CompFlow/Problem/FieldOutput.cpp#L41)):*
> ```cpp
> 
> std::vector< std::vector< tk::real > > 
> CompFlowFieldOutput( ncomp_t,
> ```

These functions aren't used anymore. Great catch!

jbakosi:
Okay, thanks the explanation. I see no point in _not_ merging this PR then. As I mentioned in the intro, I will take a 3rd pass removing the vectors from the input deck. Please go ahead and merge system1 and system2 if/when they pass CI.

adityakpandare (approved):
> Reviewed 9 of 9 files at r2, 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/591)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @jbakosi)

PR #608, opened 2024-05-09
Solids non-conservative term corrections and bug fixes

adityakpandare:
1. correct error in non-conservative terms from LF flux, and update baselines due to this,
2. bug fix in resetting Cauchy stress tensor,
3. apply BC on stress prim-vector,
4. add g-entries in diagHeader().

The corrected non-conservative terms have a much better pressure-equilibrium retention for the solid-fluid advection problem. See below. Also, the Cu-Cu shocktube is also slightly better in similar ways.
mmsadv-debugncn

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/608)


airaudofacundo (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/608)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #625, opened 2024-08-12
JWL bisection

g-rydquist:
In some cases, when performing the bisection method to find the density of a JWL material from a known pressure and temperature, the guessed bounds do not contain the correct density. Added code shifts bounds until suitable bounds are found.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625)


adityakpandare (changes requested):
Nice catch! Just a couple of questions so that I can understand better.

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-:-O46sYni34RQcpm37TVC:btsepez)* status: all files reviewed, 2 unresolved discussions (waiting on @g-rydquist)

___
*[`src/PDE/EoS/JWL.cpp` line 358 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46rv4bBnnPv18M_yi0:-O46rv4bBnnPv18M_yi1:b-o8swdt) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c367d4e668c3b69c14ba8718e62948f07ef80fee/src/PDE/EoS/JWL.cpp#L358)):*
> ```cpp
>   while ( p_known < PfromRT( a, t_known)) {
>     b = a;
>     a *= 1e-6;
> ```

For my understanding: if the bounds are too high, the right bound is replaced with the left one, and the left bound is replaced with `1e-6*a`. This attempts to bracket the density, by shifting the bounds leftward since the initial bounds were too high. Is my understanding correct? 
___
*[`src/PDE/EoS/JWL.cpp` line 396 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46sT_89dSgXdjoGGZt:-O46sT_89dSgXdjoGGZu:b-sl5zdk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5e7e4a7617be7b71dc7609175d7e618661940b94/src/PDE/EoS/JWL.cpp#L396)):*
> ```cpp
>     {
>       Throw("JWL bisection for density resulted in left/right bound as "
>       "solution. Extend bounds for correctness");
> ```

Why is this removed?

g-rydquist (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/EoS/JWL.cpp` line 358 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46rv4bBnnPv18M_yi0:-O46zA-L60hUcUHmjNeB:bp8t4b1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c367d4e668c3b69c14ba8718e62948f07ef80fee/src/PDE/EoS/JWL.cpp#L358)):*

For my understanding: if the bounds are too high, the right bound is replaced with the left one, and the left bound is replaced with `1e-6*a`. This attempts to bracket the density, by shifting the bounds leftward since the initial bounds were too high. Is my understanding correct?
</blockquote></details>

That's right. I chose `1e6` since that is the initial factor separating `a` and `b`. I suppose this does assume that pressure increases monotonically with density, which, looking at the Menikoff paper, isn't always true for low temperature if we choose Er in Eq. 13 to not be 0, which I think is something we had discussed briefly at one point. So something to keep in mind I suppose?
___
*[`src/PDE/EoS/JWL.cpp` line 396 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46sT_89dSgXdjoGGZt:-O46wJR-E-new6RTSoTN:b-5tek3p) ([raw file](https://github.com/quinoacomputing/quinoa/blob/5e7e4a7617be7b71dc7609175d7e618661940b94/src/PDE/EoS/JWL.cpp#L396)):*

Why is this removed?
</blockquote></details>

A couple reasons. First, it should be superseded by the above change. Basically, this statement is only intended to be entered if the bounds do not contain the right density, in which case it would throw an error, which shouldn't end up being an issue now. Second, it didn't actually work because of a numerical precision issue. Eventually, the bounds would converge on `a = b`, but `a` and `b` were large enough that `b - a ~= 1e-11` indefinitely. In fact, looking at one of the break conditions (`(b - a)/2.0 < 1e-12`), I'm not sure it could actually ever be entered. So the simulation would run until `maxiter` was reached anyways.

g-rydquist (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625)* status: all files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/EoS/JWL.cpp` line 358 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46rv4bBnnPv18M_yi0:-O4740VR0c3PTmQwmDiv:bi3u6h3) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c367d4e668c3b69c14ba8718e62948f07ef80fee/src/PDE/EoS/JWL.cpp#L358)):*

That's right. I chose `1e6` since that is the initial factor separating `a` and `b`. I suppose this does assume that pressure increases monotonically with density, which, looking at the Menikoff paper, isn't always true for low temperature if we choose Er in Eq. 13 to not be 0, which I think is something we had discussed briefly at one point. So something to keep in mind I suppose?
</blockquote></details>

I suppose reading a bit more, Menikoff warns against using the JWL EoS at these temperatures anyways....

adityakpandare (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @g-rydquist)

___
*[`src/PDE/EoS/JWL.cpp` line 358 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-O46rv4bBnnPv18M_yi0:-O474Co24kVg0Cto3x5l:bvq9vlr) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c367d4e668c3b69c14ba8718e62948f07ef80fee/src/PDE/EoS/JWL.cpp#L358)):*

I suppose reading a bit more, Menikoff warns against using the JWL EoS at these temperatures anyways....
</blockquote></details>

Yes, agreed, JWL is inaccurate at lower temperatures/specific-volumes. So this looks good.

adityakpandare (approved):
It looks like clang caught an unused variable in this PR. Would you take a look at it?

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-:-O475Lfz0j6eCBfRobof:b-dw22qs)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @g-rydquist)

g-rydquist (commented):
Yes, it looks like it was a_o and b_o. I got rid of them.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625#-:-O476nERBFsCaefxn0Pw:b-7nu55n)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @adityakpandare)

adityakpandare (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/625)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @g-rydquist)

PR #632, opened 2024-11-20
Viscous heat flux added

adityakpandare:
This commit adds the heat flux component from viscosity in the energy equation. Some assumptions are made: Prandtl number is 0.71.

### DIFF
--- a/src/PDE/Integrate/Boundary.cpp
+++ b/src/PDE/Integrate/Boundary.cpp
@@ -433,6 +433,7 @@ bndSurfIntViscousFV(
   const StateFn& gradFn,
   const Fields& U,
   const Fields& P,
+  const Fields& T,
   const std::vector< int >& srcFlag,
   Fields& R,
   int intsharp )
@@ -457,6 +458,7 @@ bndSurfIntViscousFV(
 //!   at boundaries
 //! \param[in] U Solution vector at recent time step
 //! \param[in] P Vector of primitives at recent time step
+//! \param[in] T Vector of temperatures at recent time step
 //! \param[in] srcFlag Whether the energy source was added
 //! \param[in,out] R Right-hand side vector computed
 //! \param[in] intsharp Interface compression tag, an optional argument, with
@@ -551,6 +553,14 @@ bndSurfIntViscousFV(
         // Get BC for cell-averaged state
         auto varcc = state( ncomp, mat_blk, ucc,
           centroids[1][0], centroids[1][1], centroids[1][2], t, fn );
+        // Hard-coded temperature BC- only works for adiabatic 'noslipwall',
+        // 'symmetry', 'extrapolate' bc
+        std::array< std::vector< real >, 2 > Tcc{{
+          std::vector<real>(nmat), std::vector<real>(nmat) }};
+        for (std::size_t k=0; k<nmat; ++k) {
+          Tcc[0][k] = T(el, k*rdof);
+          Tcc[1][k] = Tcc[0][k];  // actual bc
+        }
 
         // Numerical viscous flux
         // ---------------------------------------------------------------------
@@ -575,9 +585,22 @@ bndSurfIntViscousFV(
           for (std::size_t j=0; j<3; ++j)
             dudx[i][j] = 0.5 * (grad[0][3*i+j] + grad[1][3*i+j]);
 
-        // 3. Compute flux
+        // 3. average dT/dx_j
+        std::vector< std::array< real, 3 > > dTdx(nmat,
+          std::array< real, 3 >{{0, 0, 0}});
+        for (std::size_t k=0; k<nmat; ++k) {
+          auto mark = k*rdof;
+          for (std::size_t j=0; j<3; ++j) {
+            dTdx[k][j] =
+                dBdx_l[j][1] * T(el, mark+1)
+              + dBdx_l[j][2] * T(el, mark+2)
+              + dBdx_l[j][3] * T(el, mark+3);
+          }
+        }
+
+        // 4. Compute flux
         auto fl = modifiedGradientViscousFlux(nmat, ncomp, fn, centroids, var,
-          varcc, dudx);
+          varcc, Tcc, dudx, dTdx);
 
         // Add the surface integration term to the rhs
         for (ncomp_t c=0; c<ncomp; ++c)
--- a/src/PDE/Integrate/Boundary.cpp
+++ b/src/PDE/Integrate/Boundary.cpp
@@ -588,15 +588,6 @@ bndSurfIntViscousFV(
         // 3. average dT/dx_j
         std::vector< std::array< real, 3 > > dTdx(nmat,
           std::array< real, 3 >{{0, 0, 0}});
-        for (std::size_t k=0; k<nmat; ++k) {
-          auto mark = k*rdof;
-          for (std::size_t j=0; j<3; ++j) {
-            dTdx[k][j] =
-                dBdx_l[j][1] * T(el, mark+1)
-              + dBdx_l[j][2] * T(el, mark+2)
-              + dBdx_l[j][3] * T(el, mark+3);
-          }
-        }
 
         // 4. Compute flux
         auto fl = modifiedGradientViscousFlux(nmat, ncomp, fn, centroids, var,
--- a/src/PDE/Integrate/Boundary.hpp
+++ b/src/PDE/Integrate/Boundary.hpp
@@ -104,6 +104,7 @@ bndSurfIntViscousFV(
   const StateFn& gradFn,
   const Fields& U,
   const Fields& P,
+  const Fields& T,
   const std::vector< int >& srcFlag,
   Fields& R,
   int intsharp );
--- a/src/PDE/Integrate/Surface.cpp
+++ b/src/PDE/Integrate/Surface.cpp
@@ -510,6 +510,7 @@ surfIntViscousFV(
   const Fields& geoElem,
   const Fields& U,
   const Fields& P,
+  const Fields& T,
   const std::vector< int >& srcFlag,
   Fields& R,
   int intsharp )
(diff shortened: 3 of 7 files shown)


g-rydquist (changes requested):
Nice work, Aditya. I had trouble following the calculation of the temperature in `computeTemperaturesFV()`, but things made sense assuming they were calculated correctly. My concern, which I commented about in `modifiedGradientViscousFlux()`, is that it seems like it would make more sense for an average temperature and conductivity to be used to calculate the conduction instead of per material, but perhaps there is some physical reason for this I am missing.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-:-OCAWa0y3OkckpBaUmdx:b-emokz5)* status: all files reviewed, 3 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/Reconstruction.cpp` line 816 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCATNec9H3KYcZVd2HH:-OCATNec9H3KYcZVd2HI:bpgcno7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Reconstruction.cpp#L816)):*
> ```cpp
>         T(e, mark+idof) = R[mark+idof] / L[idof];
>     }
>   }
> ```

Having some trouble understanding what's going on in this function. From `Surface::surfIntViscousFV()`, it looks like we're calculating `T` on average and at 3 other locations (?) by calculating the integral in the details and inverting, but I don't really understand the details of where it's calculated or where the equation comes from.
___
*[`src/PDE/Integrate/Surface.cpp` line 772 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAMOW71f-7UBVr0IFi:-OCAMOW71f-7UBVr0IFj:b7d3eag) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Surface.cpp#L772)):*
> ```cpp
>     conduct_mat[k] = inciter::getmatprop< tag::mu >(k) *
>       inciter::getmatprop< tag::cv >(k) * inciter::getmatprop< tag::gamma >(k)
>       / 0.71;
> ```

Any reason not to just add a `conduct_mat` keyword?
___
*[`src/PDE/Integrate/Surface.cpp` line 798 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAPU1eAOernsz28Y6W:-OCAPU1eAOernsz28Y6X:b-qgpd04) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Surface.cpp#L798)):*
> ```cpp
>       for (std::size_t i=0; i<3; ++i)
>         energyFlux[k][j] += uAvg[i] * tau[inciter::stressCmp[i][j]] +
>           conduct_mat[k] * dTdx_m[k][j];
> ```

Is this suggesting that material `k` in one cell will only conduct heat with material `k`the neighboring cell, and not with other materials? I suppose it makes more sense to me that the conduction would be in an average sense (for both temperatures and conductivities) instead of for individual components, and then would be distributed to the materials according to the volume fractions, like with the the energy transport with the stress.
___
*[`src/PDE/Integrate/Boundary.cpp` line 562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAJPjY-MAnZpN4iMid:-OCAJPjY-MAnZpN4iMie:bxteeb9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Boundary.cpp#L562)):*
> ```cpp
>         for (std::size_t k=0; k<nmat; ++k) {
>           Tcc[0][k] = T(el, k*rdof);
>           Tcc[1][k] = Tcc[0][k];  // actual bc
> ```

Surely not particularly urgent, but actually being able to have a constant temperature BC might be kind of nice.

adityakpandare (commented):
Thank you for taking a look @g-rydquist. I'm also unclear about how the viscous terms look like in a multi-temperature single-velocity model. Perhaps practically more important, we are not supposed to enable viscosity for a multi-material case, only for single-material one. I have commented on your questions. Please let me know if you have more!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-:-OCArI9KEwQL4itdJT45:b-vrtty5)* status: all files reviewed, 3 unresolved discussions (waiting on @g-rydquist)

___
*[`src/PDE/Reconstruction.cpp` line 816 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCATNec9H3KYcZVd2HH:-OCAkfIG2S5BHwgnbDkQ:b-tmylxn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Reconstruction.cpp#L816)):*

Having some trouble understanding what's going on in this function. From `Surface::surfIntViscousFV()`, it looks like we're calculating `T` on average and at 3 other locations (?) by calculating the integral in the details and inverting, but I don't really understand the details of where it's calculated or where the equation comes from.
</blockquote></details>

This function uses the weak form to compute the temperature degrees of freedom (dofs) based on dofs of the other quantities that we have available (density, energy, etc). The weak form equation for each element `e` is:
```
T_i \int_e B_i B_j = \int_e T_{EOS}(rho, e, u)
```
The integral on the left hand side is the mass matrix `L` (which is diagonal for our choice of B's which are basis functions), and the integral on the right side has to be computed using numerical quadrature, by evaluating the temperature at each quadrature point using the EOS. Inverting `L` gives us `T_i` which is in this case, a second order polynomial of temperature: `T_1 B_1 + T_2 B_2 + T_3 B_3 + T_4 B_4`. The first coefficient `T_1` is the cell-average, and the others are high-order moments.
___
*[`src/PDE/Integrate/Boundary.cpp` line 562 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAJPjY-MAnZpN4iMid:-OCAmGXU0ymiEQ5Eyoz5:b-bfhv9j) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Boundary.cpp#L562)):*

Surely not particularly urgent, but actually being able to have a constant temperature BC might be kind of nice.
</blockquote></details>

Yes, I have that in my mind for the future. This is an adiabatic bc, and an isothermal one will be good to have.
___
*[`src/PDE/Integrate/Surface.cpp` line 772 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAMOW71f-7UBVr0IFi:-OCAmSAy01j3HiQsHhA7:byz2g8k) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Surface.cpp#L772)):*

Any reason not to just add a `conduct_mat` keyword?
</blockquote></details>

Mostly because the Prandtl number is usually always assumed to be 0.71, and with that, we have all the information we need to compute conductivity. Asking the user for that is superfluous at this point.
___
*[`src/PDE/Integrate/Surface.cpp` line 798 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAPU1eAOernsz28Y6W:-OCAn4zG78tN5ihQXLEY:b-dbe4sp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Surface.cpp#L798)):*

Is this suggesting that material `k` in one cell will only conduct heat with material `k`the neighboring cell, and not with other materials? I suppose it makes more sense to me that the conduction would be in an average sense (for both temperatures and conductivities) instead of for individual components, and then would be distributed to the materials according to the volume fractions, like with the the energy transport with the stress.
</blockquote></details>

At this point, yes, `k` only conducts with `k` through this term. The real reason for this is that it is unclear to me what the viscous PDEs look like for a multi-temperature single-velocity model. One requirement is that when the total energy equations of all materials are added up, they should result in a traditional compressible Navier-Stokes equation for the bulk total energy. I haven't seen this formulation in a paper yet. One major difference between the temperature stresses vs the viscous stress is that this is a single velocity formulation- so the viscous stress term `tau` is more easily defined. But more importantly, viscosity is supposed to be used only for single material flows as of now, so this does not matters. Have you seen a paper which includes the multi-material equations with the viscous terms? If there is, I can modify these terms accordingly.

g-rydquist (approved):
:lgtm:

> Dismissed @g-rydquist from a discussion.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-:-OCESqy-2umbNLZBAOg2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[`src/PDE/Integrate/Surface.cpp` line 798 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-OCAPU1eAOernsz28Y6W:-OCEPjsS2DETh5CM8Sh0:bbmfai0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c3febe2160fbfae270365a55899cd143a798f320/src/PDE/Integrate/Surface.cpp#L798)):*

At this point, yes, `k` only conducts with `k` through this term. The real reason for this is that it is unclear to me what the viscous PDEs look like for a multi-temperature single-velocity model. One requirement is that when the total energy equations of all materials are added up, they should result in a traditional compressible Navier-Stokes equation for the bulk total energy. I haven't seen this formulation in a paper yet. One major difference between the temperature stresses vs the viscous stress is that this is a single velocity formulation- so the viscous stress term `tau` is more easily defined. But more importantly, viscosity is supposed to be used only for single material flows as of now, so this does not matters. Have you seen a paper which includes the multi-material equations with the viscous terms? If there is, I can modify these terms accordingly.
</blockquote></details>

Yes, that makes sense. I have not seen any literature on compressible, multi-material, viscous flows!

g-rydquist (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/632#-:-OCFHiAK806Nk90RIElU:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #641, opened 2025-03-05
Rigid body motion with 2 dofs

adityakpandare:
This set of commits adds 2 degree-of-freedom rigid body motion capability to the overset mesh solver. Only the 2 linear DOFs are considered. This is intended to be used for cases with a symmetry plane where motion of the body in direction normal to the symmetry plane is restricted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/641)


g-rydquist (changes requested):
Nice job @adityakpandare! I didn't see any real issues, and just left a few nit-picky comments and suggestions. Otherwise, things look pretty good to me!

> Reviewed 19 of 20 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-:-OKlqsVN-gdTxd6n-i_M:b-tj2r8q)* status: all files reviewed, 6 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 99 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlapVA6GIemQnTFxju:-OKlapVA6GIemQnTFxjv:b-4qf58x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L99)):*
> ```cpp
>   m_nodeblockidc(),
>   m_ixfer(0),
>   m_surfForce({{0, 0, 0}})
> ```

Is there a reason two sets of brackets are used here?
___
*[`src/Inciter/OversetFE.cpp` line 1210 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlkVTyB9pgBpY5pWyx:-OKlkVTyB9pgBpY5pWyy:b-diqxs7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L1210)):*
> ```cpp
>         g_inputdeck.get< tag::rigid_body_motion >().get< tag::symmetry_plane >();
> 
>       m_surfForce[sym_dir] = 0.0;
> ```

Do you think it would be useful to, instead of specifying x, y, or z as the symmetry plane, to instead supply a normal vector? So it might be that `m_surfForce -= tk::dot(m_surfForce, sym_vector)` or something like that. I guess generally the plane would align with our axes, but perhaps it might not in some situations.
___
*[`src/Inciter/OversetFE.cpp` line 1230 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlm59zDCuRlbaIOnUO:-OKlm59zDCuRlbaIOnUP:bkul4ye) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L1230)):*
> ```cpp
>         for (std::size_t i=0; i<3; ++i) {
>           // mesh acceleration
>           auto a_mesh = m_surfForce[i] / mass_mesh;
> ```

Only really need to calculate each `a_mesh` component once per `p` so you could swap the loops and move outside I suppose.
___
*[`tests/regression/inciter/compflow/Euler/Mesh2Mesh/CMakeLists.txt` line 87 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlqccw-eoxBwA3nrzo:-OKlqccw-eoxBwA3nrzp:bck107v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/tests/regression/inciter/compflow/Euler/Mesh2Mesh/CMakeLists.txt#L87)):*
> ```CMake
> #                      TEXT_RESULT out.1.hist.p1
> #                      TEXT_DIFF_PROG_CONF mvngoverset_hist.ndiff.cfg
> #                      LABELS "oversetfe;m2m")
> ```

Should we just remove this test completely if we're moving to a different way to do this overset mesh movement? Or is it just being left here as a reminder to add a new one later?
___
*[`src/Inciter/Transporter.hpp` line 201 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlnFCND1A2Ig08bcQ3:-OKlnFCND1A2Ig08bcQ4:b34s0c4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/Transporter.hpp#L201)):*
> ```cpp
> 
>     //! Reduction target that computes minimum timestep across meshes
>     void minDtAcrossMeshes( CkReductionMsg* advMsg );
> ```

Perhaps this function should be renamed since it also has force information?
___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 538 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlpx185dATZepT0yxe:-OKlpx185dATZepT0yxf:b-1a2eca) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/PDE/CompFlow/CGCompFlow.hpp#L538)):*
> ```cpp
>         real reA = U(N[0],4);
>         real reB = U(N[1],4);
>         real reC = U(N[2],4);
> ```

Are the functions like, e.g., `momentumDofIdx` out of scope at this point, and that's why these are just manually written out?

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/641)* status: 15 of 20 files reviewed, 6 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 99 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlapVA6GIemQnTFxju:-OKm-tnPCKm4QhIgXJ3l:bhzvxgf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L99)):*

Is there a reason two sets of brackets are used here?
</blockquote></details>

That's how c++ std::arrays are initialized. I'm not sure why this is- has to do with arrays being an "aggregate type". See https://en.cppreference.com/w/cpp/container/array. In any case, compilers complain if double braces are not used.
___
*[`src/Inciter/OversetFE.cpp` line 1210 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlkVTyB9pgBpY5pWyx:-OKm2OkB6jxUJTIQxDXe:b-dx8x62) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L1210)):*

Do you think it would be useful to, instead of specifying x, y, or z as the symmetry plane, to instead supply a normal vector? So it might be that `m_surfForce -= tk::dot(m_surfForce, sym_vector)` or something like that. I guess generally the plane would align with our axes, but perhaps it might not in some situations.
</blockquote></details>

This is a very good point. However, using a general plane for symmetry will complicate adding rotation (the 3rd DOF). We should get the 3DOF solver working and verified first, and then make this change. Let me create an issue for this, so that we don't forget. 
___
*[`src/Inciter/OversetFE.cpp` line 1230 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlm59zDCuRlbaIOnUO:-OKm0n41CwXDdBxWjLuS:b-u6irt0) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L1230)):*

Only really need to calculate each `a_mesh` component once per `p` so you could swap the loops and move outside I suppose.
</blockquote></details>

Yes, I realized this when working on the 3DOF motion. Is it okay if I pushed this change in that change-set?
___
*[`src/Inciter/Transporter.hpp` line 201 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlnFCND1A2Ig08bcQ3:-OKm3HPa7vtJYRZvHV0F:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/Transporter.hpp#L201)):*

Perhaps this function should be renamed since it also has force information?
</blockquote></details>

Done.
___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 538 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlpx185dATZepT0yxe:-OKm1-hw3lnsqT3HZIQX:b-fnxhek) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/PDE/CompFlow/CGCompFlow.hpp#L538)):*

Are the functions like, e.g., `momentumDofIdx` out of scope at this point, and that's why these are just manually written out?
</blockquote></details>

Two reasons: `momentumDofIdx` and other such indexing functions are only defined for `MultiMat` and `MultiSpecies`. This is `CompFlow`, so those functions aren't defined for this PDE type. Secondly, each unknown was manually written out for vectorization- part reason why the ALECG and OversetFE with single mesh are so fast compared to FV. Of course, we have to give up on this for FV-multimaterial since the equation indices go through the indexing functions, and compilers usually cannot vectorize when function calls appear within the kernel.
___
*[`tests/regression/inciter/compflow/Euler/Mesh2Mesh/CMakeLists.txt` line 87 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlqccw-eoxBwA3nrzo:-OKm21LS92J1IyPgF814:b2lkkik) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/tests/regression/inciter/compflow/Euler/Mesh2Mesh/CMakeLists.txt#L87)):*

Should we just remove this test completely if we're moving to a different way to do this overset mesh movement? Or is it just being left here as a reminder to add a new one later?
</blockquote></details>

Yes, this is a reminder to add a new test, and a reference to how to add one (since having moving meshes complicates the filenames, and we need to get those right to have the test pass).

g-rydquist (approved):
:lgtm:

> Reviewed 5 of 5 files at r3, all commit messages.
> Dismissed @g-rydquist from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-:-OKmC9Rh9bCJSprNdw91:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 99 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlapVA6GIemQnTFxju:-OKmCJAE8i0Cq7ckEwJT:b-4ncgfz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L99)):*

That's how c++ std::arrays are initialized. I'm not sure why this is- has to do with arrays being an "aggregate type". See https://en.cppreference.com/w/cpp/container/array. In any case, compilers complain if double braces are not used.
</blockquote></details>

Huh, learn something new everyday.
___
*[`src/Inciter/OversetFE.cpp` line 1230 at r2](https://reviewable.io/reviews/quinoacomputing/quinoa/641#-OKlm59zDCuRlbaIOnUO:-OKmCVFl7W_465BMQwVA:brlnc56) ([raw file](https://github.com/quinoacomputing/quinoa/blob/336041ccfd23c720824ae90c2fc4f34866912a59/src/Inciter/OversetFE.cpp#L1230)):*

Yes, I realized this when working on the 3DOF motion. Is it okay if I pushed this change in that change-set?
</blockquote></details>

Good with me.

PR #643, opened 2025-03-12
Rigid body 3dof motion

adityakpandare:
This set of commits adds rotation to the 2 DOF solver, resulting in a 3 DOF capability (translation in 2 directions, and rotation about the third direction).

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)


g-rydquist (changes requested):
Nice job, @adityakpandare , looks pretty good to me. There was one spot where I think there was an issue with the calculation of the linear accelerations resulting from the rotations, but it looks like that was replaced in the latest commit by just using `tk::rotatePoint` to move the points. Perhaps those should just be removed if they aren't being used anymore?

> Reviewed 17 of 17 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-:-OLZr-1m6-ZQGD-PhqOP:b-1q9xbm)* status: all files reviewed, 4 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Control/Inciter/InputDeck/LuaParser.cpp` line 687 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLGJn4d5j1rvAg4UyrR:-OLGJn4d5j1rvAg4UyrS:b6ghq4l) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Control/Inciter/InputDeck/LuaParser.cpp#L687)):*
> ```cpp
>         mesh_deck[i].get< tag::mass >(), 0.0);
> 
>       // moment of inertia. this is currently only configured for planar motion
> ```

Is it within the scope of this work to calculate full, 3-axis rotation of the object? Not necessarily for this PR, but maybe a future one. I suppose the user would need to specify a full inertia tensor, and that tensor would need to be transformed according to whatever angle the body rotates. Sounds a little tedious but maybe not too bad.
___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 527 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLGP0ql7jUgvS3ar6BX:-OLGP0ql7jUgvS3ar6BY:bypkhfw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/PDE/CompFlow/CGCompFlow.hpp#L527)):*
> ```cpp
>         real reC = U(N[2],4);
>         // compute face normal
>         real nx, ny, nz;
> ```

Just noticed as I was going through, why are these `real` instead of `tk::real`?
___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OLZkZ_7962Mbrv98b-2:b-2yr4qi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*
> ```cpp
>         auto theta = std::acos(rCM[i2]/r_mag);
>         auto a1 = a_tgt*std::cos(theta);
>         auto a2 = a_tgt*std::sin(theta);
> ```

I think there is an issue in how these are calculated that doesn't properly account for the orientation of the `rCM` vector. I've attached an image below. Basically, for a positive angular acceleration around the *z*-direction, it looks like the accelerations may be incorrect, and, in the two cases I've highlighted, would calculated the same values, even though the should actually be different. I think the correct way to do this would be the following code snippet, though you may want to double check me. Although I'm not sure I'm seeing where `u_mesh`/`MeshVel()` end up getting used...





_Code snippet:_
```cpp
auto theta = std::atan2(rCM[i2],rCM[i1]);
auto a1 = -a_tgt*std::sin(theta);
auto a2 = a_tgt*std::cos(theta);
```
___
*[`src/Inciter/OversetFE.cpp` line 1294 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZjdmK7uKZ9KdzDHyc:-OLZjdmK7uKZ9KdzDHyd:bqnrpn2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1294)):*
> ```cpp
>         std::array< tk::real, 3 > angles{{ 0, 0, 0 }},
>           point{{d->Coord()[0][p], d->Coord()[1][p], d->Coord()[2][p]}};
>         angles[sym_dir] = dtheta * 180.0/(4.0*std::atan(1.0));
> ```

Is there not a built-in value of pi somewhere that we can use?

g-rydquist (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1293 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_ZFiq6vGVpuRYO0eD:-OL_ZFiq6vGVpuRYO0eE:bao06yy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1293)):*
> ```cpp
>         // add contribution of rotation to mesh displacement
>         std::array< tk::real, 3 > angles{{ 0, 0, 0 }},
>           point{{d->Coord()[0][p], d->Coord()[1][p], d->Coord()[2][p]}};
> ```

Also, should `point` be `rCM`, not the actual coordinates of the point? Since this point is what is being rotated about? then I suppose you would need to add the results back to the coordinates. I noticed that the sphere had some motion with rotation that it didn't when rotation was removed, and I wonder if this could be responsible.

g-rydquist (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 6 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1256 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_mryf45aJk22f7Jhr:-OL_mrygC3HMtz4zcsZo:b-q0jaqv) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1256)):*
> ```cpp
>           d->Coord()[0][p] - m_centMass[0],
>           d->Coord()[1][p] - m_centMass[1],
>           d->Coord()[2][p] - m_centMass[2] }};
> ```

Okay, sorry to keep spamming this PR, but I was trying a couple things, and I discovered that `m_centMass` is zero here! It looks like it never actually gets read in, from what I can see. So that seems to be why the no rotation case and the rotational case aren't matching up: it's rotating about the origin.

adityakpandare (commented):
Thanks for the quick response! And good catch on the projection of the acceleration! Let me know what you think. Pushing changes in a bit.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-:-OL_FGg_4yI6DoTeHmzD:b-8bvymt)* status: all files reviewed, 6 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Control/Inciter/InputDeck/LuaParser.cpp` line 687 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLGJn4d5j1rvAg4UyrR:-OLZx3E434UlAiZGA_fJ:b-omrsc7) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Control/Inciter/InputDeck/LuaParser.cpp#L687)):*

Is it within the scope of this work to calculate full, 3-axis rotation of the object? Not necessarily for this PR, but maybe a future one. I suppose the user would need to specify a full inertia tensor, and that tensor would need to be transformed according to whatever angle the body rotates. Sounds a little tedious but maybe not too bad.
</blockquote></details>

Let's discuss offline about 'scope of work' the next time we meet. For this pull request however, full 3D rotations are out of scope since they are not required anyway for 3DOF motion.
___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OL_ELC9CWvNnl9i2vET:bdft6t6) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*

I think there is an issue in how these are calculated that doesn't properly account for the orientation of the `rCM` vector. I've attached an image below. Basically, for a positive angular acceleration around the *z*-direction, it looks like the accelerations may be incorrect, and, in the two cases I've highlighted, would calculated the same values, even though the should actually be different. I think the correct way to do this would be the following code snippet, though you may want to double check me. Although I'm not sure I'm seeing where `u_mesh`/`MeshVel()` end up getting used...


</blockquote></details>

Nice catch! And it is important to get `u_mesh` right, because it is used to assign mesh velocities while computing fluxes. Would your suggestion work in the 3rd and 4th quadrant? The correct trig here would take the cosine and sine of the angle made by the `a_tgt` vector with the +ve i1 axis. That would work for all quadrants. But in c++, no way to get angles with respect to the +ve i1 axis from the coordinates, so we need an extra test. See if the new code makes sense to you.
___
*[`src/Inciter/OversetFE.cpp` line 1293 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_ZFiq6vGVpuRYO0eD:-OL_jJrTEQ8w_I5Ox_Ax:b-r67gaq) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1293)):*

Also, should `point` be `rCM`, not the actual coordinates of the point? Since this point is what is being rotated about? then I suppose you would need to add the results back to the coordinates. I noticed that the sphere had some motion with rotation that it didn't when rotation was removed, and I wonder if this could be responsible.
</blockquote></details>

Yes, you're right- it should be `rCM`. The way it is currently, it rotates the points about (0, 0, 0). The points should be rotated about the center of mass. For this, `rCM` should be rotated, and then, the center of mass coordinates should be added back to the rotation to get the final position.
___
*[`src/Inciter/OversetFE.cpp` line 1294 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZjdmK7uKZ9KdzDHyc:-OLZw__u0XKdha-DdSW7:b-52y035) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1294)):*

Is there not a built-in value of pi somewhere that we can use?
</blockquote></details>

C++ 20 added pi to the std lib as `std::numbers::pi`. But this would require us to have c++ 20 as a requirement (which could  be a breaking change for systems with older compilers like gcc < 9.0 or so). I believe we only require c++ 17 as of now. Which is why I didn't use the c++ 20 functionality. Please correct me if I'm wrong- in which case I'll replace this with `std::numbers::pi`. In any case, I have precomputed pi outside this loop, since it is required multiple times now.
___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 527 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLGP0ql7jUgvS3ar6BX:-OLZvLnm2fAy3rLrDhd-:bf9heho) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/PDE/CompFlow/CGCompFlow.hpp#L527)):*

Just noticed as I was going through, why are these `real` instead of `tk::real`?
</blockquote></details>

Good question. `tk` is within scope for this function, which is why the `tk::` namespace doesn't need to be specified. This is typically when the class (`CompFlow` here) is within the `tk` namespace. Not sure if that's the case here, or whether it inherts `tk`'s scope somehow. In any case, this wouldn't have compiled if it wasn't in scope of `tk`.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: 16 of 17 files reviewed, 6 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1256 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_mryf45aJk22f7Jhr:-OL_r_zN1CB6j4m-bkVS:b-beayi2) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1256)):*

Okay, sorry to keep spamming this PR, but I was trying a couple things, and I discovered that `m_centMass` is zero here! It looks like it never actually gets read in, from what I can see. So that seems to be why the no rotation case and the rotational case aren't matching up: it's rotating about the origin.
</blockquote></details>

Yes, I just spotted this too. I see weird rotations for that test. Thanks! And you're not spamming, this is my buggy code. Thanks for being patient with it!

g-rydquist (changes requested):
> Reviewed 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OL_vAFK1FcJhdiJLglu:b7fgx) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*

Nice catch! And it is important to get `u_mesh` right, because it is used to assign mesh velocities while computing fluxes. Would your suggestion work in the 3rd and 4th quadrant? The correct trig here would take the cosine and sine of the angle made by the `a_tgt` vector with the +ve i1 axis. That would work for all quadrants. But in c++, no way to get angles with respect to the +ve i1 axis from the coordinates, so we need an extra test. See if the new code makes sense to you.
</blockquote></details>

Yes, that looks fine to me. What you have written is actually equivalent to what I have, because of the identity `cos(pi/2 + theta) = -sin(theta)` and `sin(pi/2 + theta) = cos(theta)`. Is it just a limitation of c++ that it needs a positive angle? From the tests I had run, it is returning the correct components, even in the 3rd and 4th quadrants when the angle is negative.
___
*[`src/Inciter/OversetFE.cpp` line 1303 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_yNJx-QN-pqszTyTM:-OL_yNJy90iStGOczPee:b490uhk) ([raw file](https://github.com/quinoacomputing/quinoa/blob/50b1b10fba40bf2997e704b1e6384fdb56bfb7a5/src/Inciter/OversetFE.cpp#L1303)):*
> ```cpp
>         tk::rotatePoint(angles, rCM);
>         for (std::size_t i=0; i<3; ++i)
>           d->Coord()[i][p] = rCM[i] + m_centMass[i];
> ```

It looks like this is now overwriting the translational velocity. Maybe one way to go about it would be to calculate the difference in `rCM` with something like the following:



_Code snippet:_
```cpp
auto rCM_post = rCM;
tk::rotatePoint(angles, rCM_post);
for (std::size_t i=0; i<3; ++i)
  d->Coord()[i][p] += rCM_post[i] - rCM[i];
```

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OLa1DLaDzKhJXuu7zzV:b-n9qe5d) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*

Yes, that looks fine to me. What you have written is actually equivalent to what I have, because of the identity `cos(pi/2 + theta) = -sin(theta)` and `sin(pi/2 + theta) = cos(theta)`. Is it just a limitation of c++ that it needs a positive angle? From the tests I had run, it is returning the correct components, even in the 3rd and 4th quadrants when the angle is negative.
</blockquote></details>

It's possible, maybe I misunderstood the components in your image. If you're okay with what's in the code now, let's leave it as it is? Or would you prefer your version?
___
*[`src/Inciter/OversetFE.cpp` line 1303 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_yNJx-QN-pqszTyTM:-OLa0n3A0i_Wr1BnGTZy:b-9ctpx4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/50b1b10fba40bf2997e704b1e6384fdb56bfb7a5/src/Inciter/OversetFE.cpp#L1303)):*

It looks like this is now overwriting the translational velocity. Maybe one way to go about it would be to calculate the difference in `rCM` with something like the following:
</blockquote></details>

Argh... Yes. Question though: why not just `d->Coord()[i][p] += rCM[i] + m_centMass[i]`? This can be obtained using an additive translation and rotation.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1303 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_yNJx-QN-pqszTyTM:-OLa2paU62B0u1IWUeSE:b-f4vl7o) ([raw file](https://github.com/quinoacomputing/quinoa/blob/50b1b10fba40bf2997e704b1e6384fdb56bfb7a5/src/Inciter/OversetFE.cpp#L1303)):*

Argh... Yes. Question though: why not just `d->Coord()[i][p] += rCM[i] + m_centMass[i]`? This can be obtained using an additive translation and rotation.
</blockquote></details>

Well, more like `d->Coord()[i][p] = dC_translation[i] + rCM[i] + m_centMass[i]`

g-rydquist (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OLa1sdNCyf9qhgklEqZ:b-elvyn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*

It's possible, maybe I misunderstood the components in your image. If you're okay with what's in the code now, let's leave it as it is? Or would you prefer your version?
</blockquote></details>

Doesn't bother me either way, but I do think the `if (theta < 0.0) theta = (2*pi + theta);` line may not actually do anything if c++ is able to handle negative numbers, since `cos(2*pi + theta) = cos(theta)` and same for sine. Either way, though, I believe the code is doing the correct calculation.
___
*[`src/Inciter/OversetFE.cpp` line 1303 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_yNJx-QN-pqszTyTM:-OLa2ojTBIuiiMcBRfHK:b-pqpzli) ([raw file](https://github.com/quinoacomputing/quinoa/blob/50b1b10fba40bf2997e704b1e6384fdb56bfb7a5/src/Inciter/OversetFE.cpp#L1303)):*

Well, more like `d->Coord()[i][p] = dC_translation[i] + rCM[i] + m_centMass[i]`
</blockquote></details>

Ah yeah, that does make sense to me as well. Just tried it out, and they seem to come out the same. I think this latest method makes the most sense to me, since we can see all of the components contributing to the motion in one spot.

adityakpandare (commented):
Thank you for your patience on this one @g-rydquist!

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-:-OLdjknOAuPH6HlQSZlM:bvgvyhu)* status: 16 of 17 files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1273 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OLZkZ_7962Mbrv98b-1:-OLdcxch01Kh0f6H6Dr3:btz3b2d) ([raw file](https://github.com/quinoacomputing/quinoa/blob/3b761327de4f57acd45cc3e664bb1002b3379aa1/src/Inciter/OversetFE.cpp#L1273)):*

Doesn't bother me either way, but I do think the `if (theta < 0.0) theta = (2*pi + theta);` line may not actually do anything if c++ is able to handle negative numbers, since `cos(2*pi + theta) = cos(theta)` and same for sine. Either way, though, I believe the code is doing the correct calculation.
</blockquote></details>

Yes, that's right. Removed.
___
*[`src/Inciter/OversetFE.cpp` line 1303 at r3](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-OL_yNJx-QN-pqszTyTM:-OLdc_pdEAcskoCWwobF:blkbbgl) ([raw file](https://github.com/quinoacomputing/quinoa/blob/50b1b10fba40bf2997e704b1e6384fdb56bfb7a5/src/Inciter/OversetFE.cpp#L1303)):*

Ah yeah, that does make sense to me as well. Just tried it out, and they seem to come out the same. I think this latest method makes the most sense to me, since we can see all of the components contributing to the motion in one spot.
</blockquote></details>

Okay, this is implemented. Thanks!

g-rydquist (approved):
No worries. I think we are ready to go on this one then! :lgtm:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643#-:-OLdk7oF3SQPO6HE8HN0:b8clyid)* status: 16 of 17 files reviewed, all discussions resolved (waiting on @adityakpandare)

g-rydquist (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/643)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #650, opened 2025-04-03
tpg handle wider temperature bounds, add clipping

madpeck3:
Addresses some necessary changes to get more difficult TPG cases to run (hypersonic cylinder). In those cases, the energy dips significantly when initializing the run. To address this:

1) The hard bounds on the temperature range are removed. Instead, corrections are applied cp, enthalpy, to avoid a poor extrapolation of the NASA polynomials. The implementation follows what is described in the FUN3D manual:
image
To avoid repeating code, the functions for cp and enthalpy are moved into the .hpp file, where they are now treated in nondimensional form (which is the form originally given in the McBride paper with the NASA coefficients).

2) Clipping is applied to negative internal energies when calculating temperature. This is intended as a standin until a positivity preserver is implemented elsewhere in the code. Clipping is also applied to negative pressures when calculating sound speed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650)

### DIFF
--- a/src/PDE/EoS/ThermallyPerfectGas.cpp
+++ b/src/PDE/EoS/ThermallyPerfectGas.cpp
@@ -188,7 +188,7 @@ ThermallyPerfectGas::temperature(
 
   // Solve for temperature - Newton's method
   tk::real temp = 1500;     // Starting guess
-  tk::real tol = std::max(1e-8, 1e-8 * std::abs(e)); // Stopping condition
+  tk::real tol = std::max(1e-8, 1e-8 * e); // Stopping condition
   tk::real err;
   std::size_t maxiter = 10;
   std::size_t i(0);
--- a/src/PDE/EoS/ThermallyPerfectGas.cpp
+++ b/src/PDE/EoS/ThermallyPerfectGas.cpp
@@ -153,17 +153,10 @@ ThermallyPerfectGas::totalenergy(
   auto R = m_R;
 
   tk::real temp = pr / (rho * R);
-  // Identify what temperature range this falls in
-  std::size_t t_rng_idx = get_t_range(temp);
 
   // h = h_poly(T) + h_ref = e + R T (perfect gas)
-  tk::real e = R * (-m_cp_coeff[t_rng_idx][0] * std::pow(temp, -1) +
-      m_cp_coeff[t_rng_idx][1] * std::log(temp) + (m_cp_coeff[t_rng_idx][2] - 1) * temp +
-      m_cp_coeff[t_rng_idx][3] * std::pow(temp, 2) / 2 +
-      m_cp_coeff[t_rng_idx][4] * std::pow(temp, 3) / 3 +
-      m_cp_coeff[t_rng_idx][5] * std::pow(temp, 4) / 4 +
-      m_cp_coeff[t_rng_idx][6] * std::pow(temp, 5) / 5 + m_cp_coeff[t_rng_idx][7]) +
-      m_dH_ref;
+  auto h = calc_h(temp) * R * temp; // dimensionalize the results out of the calc_h func
+  tk::real e = h - R * temp;
 
   return (rho * e + 0.5 * rho * (u*u + v*v + w*w));
 }
@@ -199,30 +192,15 @@ ThermallyPerfectGas::temperature(
   std::size_t maxiter = 10;
   std::size_t i(0);
   while (i < maxiter) {
-    // Identify what temperature range the current guess is in
-    std::size_t t_rng_idx = get_t_range(temp);
-
-    // With correct polynomial coefficients, construct e(temp) and de(temp)/dT
-    tk::real f_T = R * (-m_cp_coeff[t_rng_idx][0] * std::pow(temp, -1) +
-      m_cp_coeff[t_rng_idx][1] * std::log(temp) + (m_cp_coeff[t_rng_idx][2] - 1) * temp +
-      m_cp_coeff[t_rng_idx][3] * std::pow(temp, 2) / 2 +
-      m_cp_coeff[t_rng_idx][4] * std::pow(temp, 3) / 3 +
-      m_cp_coeff[t_rng_idx][5] * std::pow(temp, 4) / 4 +
-      m_cp_coeff[t_rng_idx][6] * std::pow(temp, 5) / 5 + m_cp_coeff[t_rng_idx][7]) +
-      m_dH_ref - e;
+    // Construct e(temp) and de(temp)/dT
+    tk::real h = calc_h(temp) * R * temp;
+    tk::real f_T = h - R * temp - e;
 
     err = abs(f_T);
 
-    // Get derivative - df/dT. For loop is working through polynomial.
-    tk::real fp_T = 0;
-    tk::real power = -2;
-    for (std::size_t k=0; k<m_cp_coeff[t_rng_idx].size()-1; ++k)
-    {
-      fp_T += m_cp_coeff[t_rng_idx][k] * std::pow(temp, power);
-      if (k == 2) fp_T += -1;
-      power += 1;
-    }
-    fp_T = fp_T * R;
+    // Get derivative - df/dT.
+    tk::real cp = calc_cp(temp) * R;
+    tk::real fp_T = cp - R;
 
     // Calculate next guess
     temp = temp - f_T / fp_T;
--- a/src/PDE/EoS/ThermallyPerfectGas.cpp
+++ b/src/PDE/EoS/ThermallyPerfectGas.cpp
@@ -187,7 +187,7 @@ ThermallyPerfectGas::temperature(
 
   // Solve for temperature - Newton's method
   tk::real temp = 1500;     // Starting guess
-  tk::real tol = 1e-8 * std::abs(e); // Stopping condition
+  tk::real tol = std::max(1e-8, 1e-8 * std::abs(e)); // Stopping condition
   tk::real err;
   std::size_t maxiter = 10;
   std::size_t i(0);
--- a/src/PDE/EoS/ThermallyPerfectGas.cpp
+++ b/src/PDE/EoS/ThermallyPerfectGas.cpp
@@ -125,7 +125,8 @@ ThermallyPerfectGas::soundspeed(
 {
   auto g = m_gamma;
 
-  tk::real a = std::sqrt( g * pr / rho );
+  auto p_eff = std::max( 1.0e-15, pr );
+  tk::real a = std::sqrt( g * p_eff / rho );
 
   return a;
 }
--- a/src/PDE/EoS/ThermallyPerfectGas.cpp
+++ b/src/PDE/EoS/ThermallyPerfectGas.cpp
@@ -193,7 +193,7 @@ ThermallyPerfectGas::temperature(
 
   // Solve for temperature - Newton's method
   tk::real temp = 1500;     // Starting guess
-  tk::real tol = 1e-8 * e; // Stopping condition
+  tk::real tol = 1e-8 * std::abs(e); // Stopping condition
   tk::real err;
   std::size_t maxiter = 10;
   std::size_t i(0);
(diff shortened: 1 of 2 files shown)


madpeck3 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650)* status: all files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 70 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx0rL_9bwLBMcSMEec:-OMxJeN_E5sxKyLlr0bN:b-370a3o) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L70)):*

This part has been repeated in `calc_cp()`. Should we export this to a function so that it can be reused?
</blockquote></details>

The two reasons I didn't repeat it as a function are 1) the version in `calc_cp()` doesn't need any modifications after, so the flag `correction_needed` is superfluous, and 2) I thought the function would be pretty messy with needing to return `t_rng_idx`, `temp_poly` and `correction_needed`. 

Now that I think about it, having it as a `void` return where the arguments are modified would probably be a better way to do this (rather than packing and unpacking some tuple like thing). Up to you on if we want to pull this aside, and the `correction_needed` will just be ignored in `calc_cp()`
___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 85 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx-qQpAX4bs37iU8x9:-OMxKF1-4aWTqWeCIC8M:bcfyrgu) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L85)):*

I'm comparing this with Eq. B19 in the image you've attached- I'm confused why `h` is multiplied with `temp_poly` (the division by `temp` is to non-dimensionalize `h` I presume).
</blockquote></details>

`h` gets multiplied by `temp_poly` because the external call to the function thinks the nondimensionalizing temperature is `temp` but in reality the function internally used `temp_poly`. In the case of a correction being needed, `temp` and `temp_poly`

(To make writing the equation not horrible, T = temp and T* = temp_poly) Writing out dimensional h, post the function call:
h(T) = [ h(T*) R T* + (T - T*) cp(T*) R ]
Then make h(T) dimensionless
h(T) / R T = h(T*) T* / T + (T - T*) cp(T*) / T
___
*[`src/PDE/EoS/ThermallyPerfectGas.cpp` line 160 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx2Atn7k1T80O53Qj3:-OMxSEt6ADq3iOlsRgFd:b-dqvlsw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/94bea8e01277038fdb767fe3b690b25f01e247a5/src/PDE/EoS/ThermallyPerfectGas.cpp#L160)):*

Question about this: is the `-1*temp` missing in the new version (in `calc_h`)?
</blockquote></details>

I think what you're referencing showed up in the old code because previously I was calculating `e` directly, so I wrapped the `-R temp` term up in the NASA polynomial.

madpeck3 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 31 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMwV_OC9qTJmL9dl4B2:-OMxTaGcELeDTNzFZ3QB:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L31)):*

For Doxygen to be able to parse multi-line comments, let's add `\brief` at the beginning of this comment.
</blockquote></details>

Done.
___
*[`src/PDE/EoS/ThermallyPerfectGas.cpp` line 191 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMwVrds-fcfO6EqoHvf:-OMxTb7X6sCb38ebp0Sl:b-896fix) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.cpp#L191)):*

If we have the above clipping for internal energy, why is the `abs()` needed? Can we remove it?
</blockquote></details>

Done.

madpeck3 (commented):
I think everything is addressed at this point.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-:-OMy1sVn2pdL0LkntREW:b-9xon3h)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @adityakpandare)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 70 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx0rL_9bwLBMcSMEec:-OMy1fKzA7jFhyX9DrXw:bq31cm8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L70)):*
<details><summary><i>Previously, madpeck3 wrote…</i></summary><blockquote>

The two reasons I didn't repeat it as a function are 1) the version in `calc_cp()` doesn't need any modifications after, so the flag `correction_needed` is superfluous, and 2) I thought the function would be pretty messy with needing to return `t_rng_idx`, `temp_poly` and `correction_needed`. 

Now that I think about it, having it as a `void` return where the arguments are modified would probably be a better way to do this (rather than packing and unpacking some tuple like thing). Up to you on if we want to pull this aside, and the `correction_needed` will just be ignored in `calc_cp()`
</blockquote></details>

I just pushed a changed suggesting how to wrap up the common code together in the get_t_range function. I eliminated the `correction_needed` variable and just now check if `temp` and `temp_poly` are different. Let me know what you think.

adityakpandare (changes requested):
Thanks for the explanations. I have one follow-up question.

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-:-ON0EpxR4S3x_4GK-63m:b-mrnx4o)* status: all files reviewed, 1 unresolved discussion (waiting on @madpeck3)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 70 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx0rL_9bwLBMcSMEec:-OMyZo9aD4fowpckT9RW:b-58y1yg) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L70)):*
<details><summary><i>Previously, madpeck3 wrote…</i></summary><blockquote>

I just pushed a changed suggesting how to wrap up the common code together in the get_t_range function. I eliminated the `correction_needed` variable and just now check if `temp` and `temp_poly` are different. Let me know what you think.
</blockquote></details>

Looks good. Thanks!
___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 85 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx-qQpAX4bs37iU8x9:-ON0ER7T6fzR6La1KZF1:b-lavpez) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L85)):*
<details><summary><i>Previously, madpeck3 wrote…</i></summary><blockquote>

`h` gets multiplied by `temp_poly` because the external call to the function thinks the nondimensionalizing temperature is `temp` but in reality the function internally used `temp_poly`. In the case of a correction being needed, `temp` and `temp_poly`

(To make writing the equation not horrible, T = temp and T* = temp_poly) Writing out dimensional h, post the function call:
h(T) = [ h(T*) R T* + (T - T*) cp(T*) R ]
Then make h(T) dimensionless
h(T) / R T = h(T*) T* / T + (T - T*) cp(T*) / T
</blockquote></details>

In that case, shouldn't the same "normalization" happen with the (T-T*)*Cp term? i.e.
h(T)/RT = h(T*) T*/T + (T-T*) T*/T Cp/T 
___
*[`src/PDE/EoS/ThermallyPerfectGas.cpp` line 160 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx2Atn7k1T80O53Qj3:-OMyaQ1U5D2ptbXHB6-A:bwjgz9v) ([raw file](https://github.com/quinoacomputing/quinoa/blob/94bea8e01277038fdb767fe3b690b25f01e247a5/src/PDE/EoS/ThermallyPerfectGas.cpp#L160)):*
<details><summary><i>Previously, madpeck3 wrote…</i></summary><blockquote>

I think what you're referencing showed up in the old code because previously I was calculating `e` directly, so I wrapped the `-R temp` term up in the NASA polynomial.
</blockquote></details>

Yes, I see now. Thank you.

madpeck3 (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 85 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx-qQpAX4bs37iU8x9:-ON0FOglA0t1SbLW-IFi:b-j11edp) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L85)):*

In that case, shouldn't the same "normalization" happen with the (T-T*)*Cp term? i.e.
h(T)/RT = h(T*) T*/T + (T-T*) T*/T Cp/T
</blockquote></details>

No, since the nondimensional form of cp is just cp / R, there is no resetting of the nondimensionalization temperature for the value computed by `calc_cp`

In this equation: h(T) = \[ h(T\*) __R T\*__ + (T - T\*) cp(T\*) __R__ \], the __bolded__ terms are dimensionalizing the h, cp terms calculated as functions of T*.

adityakpandare (approved):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @madpeck3)

___
*[`src/PDE/EoS/ThermallyPerfectGas.hpp` line 85 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-OMx-qQpAX4bs37iU8x9:-ON0_Dyb2ugiDJ7KNqwy:b-qi0kdm) ([raw file](https://github.com/quinoacomputing/quinoa/blob/691a055e0b6c84bd3bf864a28096b00952a1e110/src/PDE/EoS/ThermallyPerfectGas.hpp#L85)):*
<details><summary><i>Previously, madpeck3 wrote…</i></summary><blockquote>

No, since the nondimensional form of cp is just cp / R, there is no resetting of the nondimensionalization temperature for the value computed by `calc_cp`

In this equation: h(T) = \[ h(T\*) __R T\*__ + (T - T\*) cp(T\*) __R__ \], the __bolded__ terms are dimensionalizing the h, cp terms calculated as functions of T*.
</blockquote></details>

Understood. Thanks!

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/650#-:-ON0_Htj1CEvr7HwT0ai:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @madpeck3)

PR #667, opened 2025-08-04
Alecg energy source initialization updates

adityakpandare:
This set of commits updates the energy source initial conditions for ALECG. This code was quite stale, and needed update. Thanks for finding this @clong83!

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/667)


g-rydquist (changes requested):
Looks pretty good to me. I did try running the tests and noticed `inciter:compflow_engsrc_propagating_alecg_pe4` failed with a negative density error on the first time step, though, and I didn't look into it too much to see what was going on.

> Reviewed 7 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-:-OWupRYd1i7FKFAubshA:bdao74y)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 1550 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-OWujl1K5_Gk3v7arFox:-OWujl1K5_Gk3v7arFoy:b-ph7ail) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6f0d9a5009438936429e2247416c17c2e6bd24aa/src/PDE/CompFlow/CGCompFlow.hpp#L1550)):*
> ```cpp
>                 //auto S = amplE * std::sin(pi*(node[2]-s0)/wFront);
>                 // arbitrary shape form
>                 auto S = amplE;
> ```

Do we want to keep the half-sine wave form? It seems simple enough to add by just multiplying `std::sin(pi*(r_e - rFront)/wFront)` here, and then scaling the above amplitude by `pi/2`.

adityakpandare (commented):
Yes, that inputdeck needs to be fixed. I will do that next.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-:-OWurchX9DnuGzcU9ECY:buj41a3)* status: 4 of 15 files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

___
*[`src/PDE/CompFlow/CGCompFlow.hpp` line 1550 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-OWujl1K5_Gk3v7arFox:-OWur1cO8SUuHEUmlcUi:b-8x7jf4) ([raw file](https://github.com/quinoacomputing/quinoa/blob/6f0d9a5009438936429e2247416c17c2e6bd24aa/src/PDE/CompFlow/CGCompFlow.hpp#L1550)):*

Do we want to keep the half-sine wave form? It seems simple enough to add by just multiplying `std::sin(pi*(r_e - rFront)/wFront)` here, and then scaling the above amplitude by `pi/2`.
</blockquote></details>

I don't think it is much different than the form we have now in case of a planar wave (I had done comparisons a couple of years back with three different forms of energy sourcing: sine, square, and an arbitrary shape, and found minimal differences). Secondly, the sine-wave form would be more complicated for a point-source growing sphere (non-planar wave), since there's no "cross-sectional area" like there was for a planar front. So we should just get rid of this code.

adityakpandare (commented):
Fixed the test. Had to update baselines due to change in how we source in the energy, but verified that it looks good. Also tested a problem with larger mesh and compared with available solutions to check, and that looks good too.

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-:-OWw4_rs-CsZTXHQ9lfe:b-ka8x9r)* status: 4 of 15 files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

g-rydquist (approved):
Worked for me!
:lgtm:

> Reviewed 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/667#-:-OWzfhPWDL2x0aN1PR6V:b-kgy17d)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #670, opened 2025-09-08
use 'orientation' keyword in control files to rotate overset mesh

adityakpandare:
The orientation keyword inside the mesh block in the control file is used to rotate meshes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670)

### DIFF
--- a/src/Control/Inciter/InputDeck/InputDeck.hpp
+++ b/src/Control/Inciter/InputDeck/InputDeck.hpp
@@ -1885,7 +1885,10 @@ class InputDeck : public tk::TaggedTuple< ConfigMembers > {
 
       keywords.insert({"orientation", "Configure orientation",
         R"(Configure orientation of an IC box for rotation about centroid of
-        box; or configure orientation of a mesh relative to another.)",
+        box (when specified within an IC 'box' block); or configure orientation
+        of a mesh (when specified within a 'mesh' block). Requires specification
+        of three angles about which the entity (box or mesh) is to be rotated.
+        The entity is rotated about the cartesian coordinate axes.)",
         "vector of 3 reals"});
 
       keywords.insert({"initiate", "Initiation type",
--- a/src/Control/Inciter/InputDeck/InputDeck.hpp
+++ b/src/Control/Inciter/InputDeck/InputDeck.hpp
@@ -1948,9 +1948,10 @@ class InputDeck : public tk::TaggedTuple< ConfigMembers > {
       keywords.insert({"filename", "Set filename",
         R"(Set filename, e.g., mesh filename for solver coupling.)", "string"});
 
-      keywords.insert({"location", "Configure location",
-        R"(Configure location of a mesh relative to another.)",
-        "vector of 3 reals"});
+      keywords.insert({"location", "Configure location of mesh",
+        R"(Configure location of a mesh relative to its local coordinate
+        system. Requires specification of three distances which are used to
+        relocate the mesh.)", "vector of 3 reals"});
 
       keywords.insert({"moment_of_inertia", "Moment of inertia of rigid body",
         R"(Moment of inertia of rigid body for rotational motion)", "real"});
--- a/src/Inciter/Partitioner.cpp
+++ b/src/Inciter/Partitioner.cpp
@@ -119,14 +119,44 @@ Partitioner::Partitioner(
 
   // Rotate mesh if specified
   if (rotate_mesh) {
+    auto mesh_cm =
+      g_inputdeck.get< tag::mesh >()[meshid].get< tag::center_of_mass >();
+    auto& x = m_coord[0];
+    auto& y = m_coord[1];
+    auto& z = m_coord[2];
+    for (std::size_t i=0; i<x.size(); ++i) {
+      std::array< tk::real, 3 >
+        point{{ x[i]-mesh_cm[0], y[i]-mesh_cm[1], z[i]-mesh_cm[2] }};
+      tk::rotatePoint(
+        {{ mesh_orientation[0], mesh_orientation[1], mesh_orientation[2] }},
+        point );
+      x[i] = point[0]+mesh_cm[0];
+      y[i] = point[1]+mesh_cm[1];
+      z[i] = point[2]+mesh_cm[2];
+    }
+  }
+
+  // Check if mesh location specified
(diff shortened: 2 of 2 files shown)


g-rydquist (changes requested):
> @g-rydquist reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`src/Inciter/Partitioner.cpp` line 111 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-OZjT-Fa0H3eAEzGYkR1:-OZjT-Fa0H3eAEzGYkR2:bygp9cy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c6bb43c9f9412bb596a7ad463d561f8fa6ede79c/src/Inciter/Partitioner.cpp#L111)):*
> ```cpp
>   // Check if mesh rotation specified
>   auto mesh_orientation =
>     g_inputdeck.get< tag::mesh >()[meshid].get< tag::orientation >();
> ```

So it seems like this keyword is also being used to orient an IC box (I'm seeing this in `PDE/CompFlow/CGCompFlow.hpp`). Should the description of this keyword be updated in `InputDeck.hpp` to reflect that it can be used for rotating the mesh if it is in a mesh block instead of an IC block?
___
*[`src/Inciter/Partitioner.cpp` line 129 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-OZjUrg9-GGy9hT6AT4m:-OZjUrg9-GGy9hT6AT4n:b-mnmspf) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c6bb43c9f9412bb596a7ad463d561f8fa6ede79c/src/Inciter/Partitioner.cpp#L129)):*
> ```cpp
>       tk::rotatePoint(
>         {{ mesh_orientation[0], mesh_orientation[1], mesh_orientation[2] }},
>         point );
> ```

This rotates about the origin, correct? Do we want to rotate about the center of mass? I suppose technically this rotation could be applied to the background mesh, and I don't think the CoM is readily available for this mesh... Perhaps the user could supply a point to rotate about in the mesh block?

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Inciter/Partitioner.cpp` line 111 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-OZjT-Fa0H3eAEzGYkR1:-OZlQHx19tJb2WjSjpsC:bo899ly) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c6bb43c9f9412bb596a7ad463d561f8fa6ede79c/src/Inciter/Partitioner.cpp#L111)):*

So it seems like this keyword is also being used to orient an IC box (I'm seeing this in `PDE/CompFlow/CGCompFlow.hpp`). Should the description of this keyword be updated in `InputDeck.hpp` to reflect that it can be used for rotating the mesh if it is in a mesh block instead of an IC block?
</blockquote></details>

Yes, I updated the comment for 'orientation' in Control/Inciter/InputDeck/InputDeck.hpp
___
*[`src/Inciter/Partitioner.cpp` line 129 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-OZjUrg9-GGy9hT6AT4m:-OZlQsvA7YlJqlaX-MTy:bl6983s) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c6bb43c9f9412bb596a7ad463d561f8fa6ede79c/src/Inciter/Partitioner.cpp#L129)):*

This rotates about the origin, correct? Do we want to rotate about the center of mass? I suppose technically this rotation could be applied to the background mesh, and I don't think the CoM is readily available for this mesh... Perhaps the user could supply a point to rotate about in the mesh block?
</blockquote></details>

This is a possibility, but makes it harder to align overset meshes (especially important when we have symmetry planes and mesh boundaries are very close to each other). If we rotate about origin, it's easier to imagine how the mesh will end up after rotation. I'll leave it to you- let me know what you suggest.

adityakpandare (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Inciter/Partitioner.cpp` line 129 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-OZjUrg9-GGy9hT6AT4m:-OZo19qe8lJs42N5mTDT:b-fl8e2x) ([raw file](https://github.com/quinoacomputing/quinoa/blob/c6bb43c9f9412bb596a7ad463d561f8fa6ede79c/src/Inciter/Partitioner.cpp#L129)):*

This is a possibility, but makes it harder to align overset meshes (especially important when we have symmetry planes and mesh boundaries are very close to each other). If we rotate about origin, it's easier to imagine how the mesh will end up after rotation. I'll leave it to you- let me know what you suggest.
</blockquote></details>

Thinking about this some more, it is more appropriate to rotate about the center of mass given by the user. I will update this accordingly. Thanks for pointing this out! Also, I am going to add the 'location' keyword that allows placing the overset mesh at a location specified by user. Planning to push that change to this branch as well.

adityakpandare (commented):
Okay, pushed changes with the rotation about the center-of-mass, and relocation of the mesh (useful for placing overset meshes at user-specified locations, for example).

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-:-OZoFuGn0FtCkar9FEZp:ba3md5j)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @g-rydquist)

g-rydquist (approved):
Nice work!
:lgtm:

> @g-rydquist reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/670#-:-OZp8nXU9cGhYroyEvI4:bntxkck)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @adityakpandare)

PR #672, opened 2025-09-11
Fixing some minor overset bugs

g-rydquist:
This commit fixes three things:
1. The criteria for the mesh moving is that there is some force on it. Technically, the mesh could be moving even if it has no force applied to it if it has some velocity from a previous time step. This is a pretty unlikely, edge case scenario that I don't imagine would really ever happen, but maybe it could if there is some body moving at the exact speed as a background flow.
2. The displacement from the rotation used the center of mass location at the end of the previous RK stage, not the previous time step.
3. `u_mesh` does something similar as the rotation displacement. It also doesn't seem to properly take into account the angular velocity the mesh might already have.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672)

### DIFF
--- a/src/Inciter/OversetFE.cpp
+++ b/src/Inciter/OversetFE.cpp
@@ -1276,9 +1276,11 @@ OversetFE::solve()
       }
     }
 
-    // Mark if mesh moved
+    // Mark if mesh moved or is moving
     if (std::sqrt(tk::dot(m_surfForce, m_surfForce)) > 1e-12 ||
-      std::sqrt(tk::dot(m_surfTorque, m_surfTorque)) > 1e-12)
+      std::sqrt(tk::dot(m_surfTorque, m_surfTorque)) > 1e-12 ||
+      std::sqrt(tk::dot(m_centMassVeln, m_centMassVeln)) > 1e-12 ||
+      std::sqrt(m_angVelMeshn * m_angVelMeshn) > 1e-12 )
       m_movedmesh = 1;
     else
       m_movedmesh = 0;
@@ -1317,6 +1319,7 @@ OversetFE::solve()
         }
         r_mag = std::sqrt(r_mag);
         auto a_tgt = alpha_mesh*r_mag;
+        auto v_tgt = m_angVelMeshn*r_mag;
 
         // get the other two directions
         auto i1 = (sym_dir+1)%3;
@@ -1326,6 +1329,8 @@ OversetFE::solve()
         auto theta = std::atan2(rCM[i2],rCM[i1]);
         auto a1 = a_tgt*std::cos((pi/2.0)+theta);
         auto a2 = a_tgt*std::sin((pi/2.0)+theta);
+        auto v1 = v_tgt*std::cos((pi/2.0)+theta);
+        auto v2 = v_tgt*std::sin((pi/2.0)+theta);
 
         // angle of rotation
         auto dtheta = m_angVelMesh*dtp + 0.5*alpha_mesh*dtp*dtp;
@@ -1344,16 +1349,16 @@ OversetFE::solve()
           // mesh displacement from translation
           dsT = m_centMassVel[i]*dtp + 0.5*a_mesh[i]*dtp*dtp;
           // mesh displacement from rotation
-          dsR = rCM[i] + m_centMass[i] - d->Coordn()[i][p];
+          dsR = rCM[i] + m_centMassn[i] - d->Coordn()[i][p];
           // add both contributions
           d->Coord()[i][p] = d->Coordn()[i][p] + dsT + dsR;
           // mesh velocity change from translation
-          u_mesh(p,i) += a_mesh[i]*dtp;
+          u_mesh(p,i) = m_centMassVeln[i] + a_mesh[i]*dtp;
         }
 
         // add contribution of rotation to mesh velocity
-        u_mesh(p,i1) += a1*dtp;
-        u_mesh(p,i2) += a2*dtp;
+        u_mesh(p,i1) += v1 + a1*dtp;
+        u_mesh(p,i2) += v2 + a2*dtp;
       }
 
       // update angular velocity


adityakpandare (changes requested):
Looks good! Just one question to make sure I understood correctly.

> @adityakpandare reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-:-OZuiEFd-OdK1TlBLokO:b-ejl6mo)* status: all files reviewed, 2 unresolved discussions (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1283 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZtIcKbBW64mm0B2oWw:-OZtIcKbBW64mm0B2oWx:b-73m6vw) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1283)):*
> ```cpp
>       std::sqrt(tk::dot(m_surfTorque, m_surfTorque)) > 1e-12 ||
>       std::sqrt(tk::dot(m_centMassVeln, m_centMassVeln)) > 1e-12 ||
>       std::sqrt(m_angVelMeshn * m_angVelMeshn) > 1e-12 )
> ```

Nice catch!
___
*[`src/Inciter/OversetFE.cpp` line 1361 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZuhhNUBhijP0Yk_9ra:-OZuhhNUBhijP0Yk_9rb:b1k9ia1) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1361)):*
> ```cpp
>         // add contribution of rotation to mesh velocity
>         u_mesh(p,i1) += v1 + a1*dtp;
>         u_mesh(p,i2) += v2 + a2*dtp;
> ```

So, the `u_mesh` is made of 4 components: `m_centMassVeln`, `v1/2`, `a_mesh*dt`, and `a1/2*dt`. The first two are from the previous time step, and the other two are from the current time stage. So, the first two contributions will stay constant during a time step (for all 3 RK stages). Is my understanding right?

g-rydquist (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1361 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZuhhNUBhijP0Yk_9ra:-OZyJK3B3udIpA7fq9go:bt7dv7k) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1361)):*

So, the `u_mesh` is made of 4 components: `m_centMassVeln`, `v1/2`, `a_mesh*dt`, and `a1/2*dt`. The first two are from the previous time step, and the other two are from the current time stage. So, the first two contributions will stay constant during a time step (for all 3 RK stages). Is my understanding right?
</blockquote></details>

Yes, that is correct. Are you thinking perhaps there is a more efficient way to do this that doesn't involve unnecessarily adding the first two 3 times?

adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672)* status: all files reviewed, 1 unresolved discussion (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1361 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZuhhNUBhijP0Yk_9ra:-O_CuS6g5bSub5nFlBkA:bf3i8g8) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1361)):*

Yes, that is correct. Are you thinking perhaps there is a more efficient way to do this that doesn't involve unnecessarily adding the first two 3 times?
</blockquote></details>

No, I'm just thinking about correctness. We are effectively zeroing out `u_mesh` and computing it from scratch (line with `u_mesh(p,i) = m_centMassVeln[i] + a_mesh[i]*dtp`), rather than incrementally change it like `u_(n+1) = u_n + c*dt*R_n`. So it poses two questions: is this correct? is this consistent with the unknown vector `m_u` (which is what we were attempting to do with this change)?

g-rydquist (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672)* status: all files reviewed, 1 unresolved discussion (waiting on @adityakpandare)

___
*[`src/Inciter/OversetFE.cpp` line 1361 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZuhhNUBhijP0Yk_9ra:-O_DlRAFAcxWOU_npYIn:bcd9ahy) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1361)):*

No, I'm just thinking about correctness. We are effectively zeroing out `u_mesh` and computing it from scratch (line with `u_mesh(p,i) = m_centMassVeln[i] + a_mesh[i]*dtp`), rather than incrementally change it like `u_(n+1) = u_n + c*dt*R_n`. So it poses two questions: is this correct? is this consistent with the unknown vector `m_u` (which is what we were attempting to do with this change)?
</blockquote></details>

Good questions. I think it must be consistent based on how the overset mesh is translated, right? I can't quite get my head around how we could have `u_mesh` change incrementally each RK stage and have variables like`m_centMass` not. Also, I think this must be the case based on the `rkcoef` variable, which is `rkcoef{{ 1.0/3.0, 1.0/2.0, 1.0 }}`. If we were incrementally updating, I would imagine these would need to sum to 1, but instead it seems like we do a third-time step, which informs a half-time step, both of which inform a full time step.

It also looks like earlier in `Overset::solve()`, around line 1259, we get the below line, which looks to me as though it is using the previous time step's `m_un`, as well as the current RK coefficient times the time step to update at a given stage.



_Code snippet:_
```cpp
m_u(i,c) = m_un(i,c) + m_blank[i] * rkcoef[m_stage] * dtp * m_rhs(i,c)
        / d->Vol()[i];
```

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-:-O_N8Q770gJxEWM1II2r:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @g-rydquist)

___
*[`src/Inciter/OversetFE.cpp` line 1361 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/672#-OZuhhNUBhijP0Yk_9ra:-O_N8MNU1BQFcbFZ-_OE:b-u7nlxi) ([raw file](https://github.com/quinoacomputing/quinoa/blob/223bbb43d0175af7eb5308e08531462f6ee9a694/src/Inciter/OversetFE.cpp#L1361)):*

Good questions. I think it must be consistent based on how the overset mesh is translated, right? I can't quite get my head around how we could have `u_mesh` change incrementally each RK stage and have variables like`m_centMass` not. Also, I think this must be the case based on the `rkcoef` variable, which is `rkcoef{{ 1.0/3.0, 1.0/2.0, 1.0 }}`. If we were incrementally updating, I would imagine these would need to sum to 1, but instead it seems like we do a third-time step, which informs a half-time step, both of which inform a full time step.

It also looks like earlier in `Overset::solve()`, around line 1259, we get the below line, which looks to me as though it is using the previous time step's `m_un`, as well as the current RK coefficient times the time step to update at a given stage.
</blockquote></details>

Based on our offline discussions, this looks good. Thanks!

PR #683, opened 2025-12-08
Fixed a bug where the last stage of IMEX was unnecessarily calling rhs

airaudofacundo:
Also, I was wrongly defining the variable 'rhsprev'. This small change fixes both issues.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/683)

### DIFF
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -1514,17 +1514,24 @@ DG::solve( tk::real newdt )
   if (imex_runge_kutta) {
     if (m_stage == 0)
     {
-      // Save previous rhs
-      m_rhsprev = m_rhs;
       // Initialize m_stiffrhs to zero
       m_stiffrhs.fill(0.0);
       m_stiffrhsprev.fill(0.0);
     }
   }
 
-  g_dgpde[d->MeshId()].rhs( physT, pref, myGhosts()->m_geoFace,
-    myGhosts()->m_geoElem, myGhosts()->m_fd, myGhosts()->m_inpoel, m_boxelems,
-    myGhosts()->m_coord, m_u, m_p, m_ndof, d->Dt(), m_rhs );
+  if (!imex_runge_kutta) {
+    g_dgpde[d->MeshId()].rhs( physT, pref, myGhosts()->m_geoFace,
+      myGhosts()->m_geoElem, myGhosts()->m_fd, myGhosts()->m_inpoel, m_boxelems,
+      myGhosts()->m_coord, m_u, m_p, m_ndof, d->Dt(), m_rhs );
+  }
+  else if (m_stage < m_nstage-1) {
+    // Save previous rhs
+    m_rhsprev = m_rhs;
+    g_dgpde[d->MeshId()].rhs( physT, pref, myGhosts()->m_geoFace,
+      myGhosts()->m_geoElem, myGhosts()->m_fd, myGhosts()->m_inpoel, m_boxelems,
+      myGhosts()->m_coord, m_u, m_p, m_ndof, d->Dt(), m_rhs );
+  }
 
   if (imex_runge_kutta) {
     // Implicit-Explicit time-stepping using RK3 to discretize time-derivative
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -1520,14 +1520,8 @@ DG::solve( tk::real newdt )
     }
   }
 
-  if (!imex_runge_kutta) {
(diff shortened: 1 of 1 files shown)


adityakpandare (changes requested):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/683)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @airaudofacundo)

___
*[`src/Inciter/DG.cpp` line 1534 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/683#-Ofz--UmCTYGlCNJVfDF:-Ofz--UmCTYGlCNJVfDG:b-kjmxjn) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ec162353007e311ada0ad91fa03f7361949bcd50/src/Inciter/DG.cpp#L1534)):*
> ```cpp
>       myGhosts()->m_geoElem, myGhosts()->m_fd, myGhosts()->m_inpoel, m_boxelems,
>       myGhosts()->m_coord, m_u, m_p, m_ndof, d->Dt(), m_rhs );
>   }
> ```

The `rhs()` calls in the two if-tests are the same, correct? If yes, what do you think about combining the two if-tests into one? one way I can think of is:
```
if (!imex_runge_kutta || (m_st < m_st-1)) {
  if (imex_runge_kutta && (m_st < m_st-1)) m_rhsprev = m_rhs;
  ...rhs(...);
}
```
This just makes it very clear that the rhs call is the same for IMEX or otherwise, at the expense of making the if-test complicated. My slight preference would be to make it this way, but I'm not super opposed to the way you have it either. I'll leave it to you.
___
*[`src/Inciter/DG.cpp` line 1518 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/683#-Ofyzkmg0m5gJxQKnPyT:-Ofyzkmg0m5gJxQKnPyU:bytu4j9) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e722614ad459d2ea07079b2426fe62c4a39fa4bb/src/Inciter/DG.cpp#L1518)):*
> ```cpp
>     {
>       // Save previous rhs
>       m_rhsprev = m_rhs;
> ```

This line basically stored the RHS at t_n i.e. R(t_n), into the var `m_rhsprev`. So we don't need that anymore?

airaudofacundo (commented):
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/683)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @adityakpandare)

___
*[`src/Inciter/DG.cpp` line 1518 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/683#-Ofyzkmg0m5gJxQKnPyT:-OfzJnAE6g4sjB_hZJGC:b9rclt5) ([raw file](https://github.com/quinoacomputing/quinoa/blob/e722614ad459d2ea07079b2426fe62c4a39fa4bb/src/Inciter/DG.cpp#L1518)):*

This line basically stored the RHS at t_n i.e. R(t_n), into the var `m_rhsprev`. So we don't need that anymore?
</blockquote></details>

Correct, it gets updated below.
___
*[`src/Inciter/DG.cpp` line 1534 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/683#-Ofz--UmCTYGlCNJVfDF:-OfzJwX-7A7J6OLCfw3w:brfflqz) ([raw file](https://github.com/quinoacomputing/quinoa/blob/ec162353007e311ada0ad91fa03f7361949bcd50/src/Inciter/DG.cpp#L1534)):*

The `rhs()` calls in the two if-tests are the same, correct? If yes, what do you think about combining the two if-tests into one? one way I can think of is:
```
if (!imex_runge_kutta || (m_st < m_st-1)) {
  if (imex_runge_kutta && (m_st < m_st-1)) m_rhsprev = m_rhs;
  ...rhs(...);
}
```
This just makes it very clear that the rhs call is the same for IMEX or otherwise, at the expense of making the if-test complicated. My slight preference would be to make it this way, but I'm not super opposed to the way you have it either. I'll leave it to you.
</blockquote></details>

Sure, implemented that.

adityakpandare (approved):
:lgtm: Good to merge after tests pass.

> @adityakpandare reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/683#-:-OfzKuf3Cu50qEFEVmZ2:bmnkhvl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @airaudofacundo)

PR #685, opened 2025-12-10
Reorder integrations in IMEX

airaudofacundo:
IMEX was incorrectly integrating the explicit equations after performing the implicit integrations, making it so the implicit variables used older versions of explicit variables.

Here we fix that by integrating the explicit part of all equations first, and performing the implicit operations afterwards.

In addition, due to the call to EvalPolynomialSol in stiff_rhs generating issues, we replaced it with a call to eval_state, which does the same thing minus the THINC reconstruction. A few argument were removed from stiff_rhs as a result.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/685)

### DIFF
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -2350,8 +2350,7 @@ std::vector< tk::real > DG::nonlinear_func(std::size_t e,
 
   // Compute stiff_rhs
   g_dgpde[d->MeshId()].stiff_rhs( e, myGhosts()->m_geoElem,
-    myGhosts()->m_inpoel, myGhosts()->m_coord,
-    m_u, m_p, m_ndof, m_stiffrhs );
+    m_u, m_ndof, m_stiffrhs );
 
   // Store f
   std::vector< tk::real > f(n, 0.0);
--- a/src/Inciter/DG.cpp
+++ b/src/Inciter/DG.cpp
@@ -2173,21 +2173,33 @@ DG::imex_integrate()
     m_stiffrhsprev = m_stiffrhs;
 
     // Compute the imex update
+    const auto nelem = myGhosts()->m_fd.Esuel().size()/4;
+    const auto neq = m_u.nprop()/rdof;
 
-    // Integrate explicitly on the imex equations
-    // (To use as initial values)
-    for (std::size_t e=0; e<myGhosts()->m_nunk; ++e) {
+    for (std::size_t e=0; e<nelem; ++e) {
       auto vole = myGhosts()->m_geoElem(e,0);
+      // Integrate explicitly on all equations
+      for (std::size_t c=0; c<neq; ++c)
+      {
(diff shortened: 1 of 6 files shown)


adityakpandare (changes requested):
> @adityakpandare reviewed 6 of 6 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/685)* status: all files reviewed, 1 unresolved discussion (waiting on @airaudofacundo)

___
*[`src/Inciter/DG.cpp` line 2190 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/685#-OgA7BusCGXk_ETJLDCU:-OgA7BusCGXk_ETJLDCV:b-tjbhqd) ([raw file](https://github.com/quinoacomputing/quinoa/blob/731ae09300968c0a1726865fc95cce3756df203b/src/Inciter/DG.cpp#L2190)):*
> ```cpp
>           m_u(e, rmark) = m_un(e, rmark) + d->Dt() * (
>             expl_rkcoef[0][m_stage] * m_rhsprev(e, mark)/(vole*mass_dubiner[k])
>             + expl_rkcoef[1][m_stage] * m_rhs(e, mark)/(vole*mass_dubiner[k]));
> ```

This just occurred to me- what happens in the last RK stage when `m_rhs` is not freshly computed in `DG::solve()`? is `m_rhs =` m_rhsprev`? Is this update accurate then?

airaudofacundo:
*[`src/Inciter/DG.cpp` line 2190 at r1](https://reviewable.io/reviews/quinoacomputing/quinoa/685#-OgA7BusCGXk_ETJLDCU:-OgDKnZQAoFNS0iHVF2V:bv2jmg) ([raw file](https://github.com/quinoacomputing/quinoa/blob/731ae09300968c0a1726865fc95cce3756df203b/src/Inciter/DG.cpp#L2190)):*

This just occurred to me- what happens in the last RK stage when `m_rhs` is not freshly computed in `DG::solve()`? is `m_rhs =` m_rhsprev`? Is this update accurate then?
</blockquote></details>

We only say rhsprev = rhs before calling the rhs function for m_stage = 0 or 1. For m_stage = 2 the update is in the same if-statement as the function call, so:
m_stage = 0:
  m_rhsprev = 0;
  m_rhs = R_0
m_stage = 1:
  m_rhsprev = R_0;
  m_rhs = R_1;
m_stage = 2:
  m_rhsprev = R_0;
  m_rhs = R_1;

adityakpandare (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/685#-:-OgDUzE_BJBZrp8K02ip:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @airaudofacundo)

PR #689, opened 2026-01-09
Overset body force

g-rydquist:
Adds an optional body force per unit mass to the overset mesh. One thing I found kind of strange here was that the body force gets added each RK iteration, which is why I added the `if (m_stage==0)` statement. Is this expected behavior? I didn't think so, because I figured the body force would get overwritten during  each iteration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/689)


g-rydquist (commented):
Everything looks solid to me, and it seemed to work when I tested it on the case I had been working on. I'm good to merge it.
:lgtm:

> @g-rydquist reviewed 26 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/689#-:-OnMvqkV5f6I-SNpgwaD:b-kxxr8q)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on adityakpandare).

PR #693, opened 2026-01-26
fixed index in get_t_range to properly deal with temperatures above upper bound

airaudofacundo:
For ranges (200, 1000), (1000, 6000), (6000, 20000), the t_range array looks like (200, 1000, 6000, 20000), so the size is 4, but we should be placed at 6000, which corresponds to index 2, so size-2.  Since that number is going to be used in arrays like cp_coeff which are indexed based on the interval, so there are only 3 intervals, idx=3 would give an error.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/693)


adityakpandare (approved):
:lgtm: Nice catch!

> @adityakpandare reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/693#-:-OjvPwTJDTXd_6M3ssxW:bqvm1h6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @airaudofacundo).

PR #701, opened 2026-03-11
Fixed bug where some triangles were excluded from proper sideset marking, causing spurious forces in the rigid body motion solver

clong83:
Confirmed that moving the block creates triangles marked as slipwall that should not be.  Instead of moving the block, we now check all the corners of the triangle and if any are a slipwall, the triangle gets marked.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/701)

### DIFF
--- a/src/Inciter/ALECG.cpp
+++ b/src/Inciter/ALECG.cpp
@@ -206,8 +206,11 @@ ALECG::queryBnd()
   // to obtain force on overset walls
   m_slipwallbctri.resize( m_triinpoel.size()/3, 0 );
   for (std::size_t e=0; e<m_triinpoel.size()/3; ++e)
-    if (m_slipwallbcnodes.find(m_triinpoel[e*3+0]) != end(m_slipwallbcnodes))
+    if (m_slipwallbcnodes.find(m_triinpoel[e*3+0]) != end(m_slipwallbcnodes) ||
+        m_slipwallbcnodes.find(m_triinpoel[e*3+1]) != end(m_slipwallbcnodes) ||
+        m_slipwallbcnodes.find(m_triinpoel[e*3+2]) != end(m_slipwallbcnodes)){
       m_slipwallbctri[e] = 1;
+    }
 
   // Prepare unique set of time dependent BC nodes
   m_timedepbcnodes.clear();
--- a/src/Inciter/OversetFE.cpp
+++ b/src/Inciter/OversetFE.cpp
@@ -220,8 +220,11 @@ OversetFE::getBCNodes()
   // to obtain force on overset walls
   m_slipwallbctri.resize( m_triinpoel.size()/3, 0 );
   for (std::size_t e=0; e<m_triinpoel.size()/3; ++e)
-    if (m_slipwallbcnodes.find(m_triinpoel[e*3+0]) != end(m_slipwallbcnodes))
+    if (m_slipwallbcnodes.find(m_triinpoel[e*3+0]) != end(m_slipwallbcnodes) ||
+        m_slipwallbcnodes.find(m_triinpoel[e*3+1]) != end(m_slipwallbcnodes) ||
+        m_slipwallbcnodes.find(m_triinpoel[e*3+2]) != end(m_slipwallbcnodes)){
       m_slipwallbctri[e] = 1;
+    }
 
   // Prepare unique set of time dependent BC nodes
   m_timedepbcnodes.clear();


clong83:
Passes all unit tests in debug mode.  Still fails the moving mesh os test, but this is nothing new for me on rocinante.

adityakpandare (commented):
Thanks, @clong83! Can you update the baselines for that test? That should fix it (the reason it previously failed was fixed with https://github.com/quinoacomputing/quinoa/commit/d9ebc861c61056658ae72e257b53673be1bdbb37).

> @adityakpandare made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/701#-:-OnTZVmN5WAygEdFElv7:b-w0m4ic)* status: 0 of 2 files reviewed, all discussions resolved (waiting on g-rydquist).

clong83:
Aditya-

I made the branch from fdc12a9, so I should have that fix already.

Here is the more verbose error I get:
182: -- Binary diff found match for 'statbackground_pe4.std.exo.0' in 'out.0.e-s.0.4.0'
182: -- Binary diff found match for 'statbackground_pe4.std.exo.1' in 'out.0.e-s.0.4.1'
182: -- Binary diff found match for 'statbackground_pe4.std.exo.2' in 'out.0.e-s.0.4.2'
182: -- Binary diff found match for 'statbackground_pe4.std.exo.3' in 'out.0.e-s.0.4.3'
182: CMake Error at /usr/projects/ea/Quinoa/quinoaLUA/cmake/test_runner.cmake:342 (message):
182:   Binary diff command failed:
182:   /usr/projects/ea/Quinoa/quinoa-tpl/install/bin/exodiff -m -q -f
182:   m2m_moving.cfg mvngoverset0_pe4.std.exo.0 out.1.e-s.0.4.0, error: 2

adityakpandare (commented):
Yes, but you will still need to update those baselines (`statbackground_pe4.std.exo.*` and `mvngoverset0_pe4.*`) because when you changed how the triangle-faces in a side set are determined, the resulting force (and moments) in overset will also change. So that will change the solutions, thus needing a baseline update.

> @adityakpandare made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/701#-:-OnTqsmlC007S2TlHdTu:bnxukrq)* status: 0 of 2 files reviewed, all discussions resolved (waiting on g-rydquist).

clong83:
Ahh, yes now I see.

clong83:
Interestingly, I only have one set of overset mesh results.  That is, I have out.1.e-s.0.4.{0-3}, but no out.1.e-s.2.4.{0-3}.

That will definitely cause the test to fail, right?  I can confirm it successfully ran two time steps.  I can otherwise update the static mesh files.

clong83:
I updated the overset mesh files I do have locally, and it now predictably complains that the second time step files are missing, which is the same problem I always had with it previously.  

Nothing appears to have gone wrong in the simulation, I am not sure why these files did not get written since this should include your fix.

I will pull a fresh install without these changes to see if this bug is persistent for me.

182:   output: , error output: Exodus Library Warning/Error: [ex__check_file_type]
182:   	ERROR: Could not open file
182:   '/usr/projects/ea/Quinoa/quinoaLUA/build/tests/inciter/compflow/Euler/Mesh2Mesh/inciter:compflow_euler_movingmesh_overset_pe4/out.1.e-s.2.4.0',
182:   error = No such file or directory.  Exodus Library Warning/Error:
182:   [ex_open_int] 	ERROR: failed to open

clong83:
I can confirm that a fresh pull of develop passes this test as expected given your recent fix, @adityakpandare 

However, the introduction of this bug fix in OversetFE.cpp in the slipwall BC assignment somehow prevents the out.1.e-s.2.4.{0-3} files from being written.  I have no idea why that should be the case.

The forces in the diag file are all less than 1e-12, so presumably there is no mesh motion at all.  If the rigid body solver is now bypassed because the forces are too low to matter, would that do it?

g-rydquist (approved):
I'm going to be out all of next week, so in the interest of getting this merged I'll approve it. Hopefully I don't need to re-review it after the baseline changes are pushed.

:lgtm:

> @g-rydquist reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/701#-:-OnbHrcFB9h3TEM7itVE:b-w3ccmr)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on clong83).

PR #725, opened 2026-08-05
Fixing bug in if statement logic

clong83:
Fixed if statements so the else statement does not ignore P0P1 in Boundary/Surface.cpp.  Created a do nothing if statement in Volume.cpp to prevent exiting when running P0P1.

I looked at the history of DGMultiSpecies.cpp, and it looks like before this merge there was no volume integral call for P0P1?  If I don't have that right and Volume.cpp needs to call something, please let me know.

- - -
This change is [Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/725)

### DIFF
--- a/src/PDE/Integrate/Boundary.cpp
+++ b/src/PDE/Integrate/Boundary.cpp
@@ -1189,14 +1189,14 @@ bndSurfIntViscousMultiSpecies(
       coord, fd, geoFace, geoElem, U, P, t, state, gradFn, R );
   }
 
-  if (ndof == 4) {
+  else if (ndof == 4) {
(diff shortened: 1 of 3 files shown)


adityakpandare (approved):
Thanks, @clong83! Nice catch! And yes, there's no volume integral for p0p1, since the spatial derivatives of the basis functions are zero for p0. 
:lgtm:

> @adityakpandare reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/quinoacomputing/quinoa/725#-:-OzMbopW7QCvTLRpjzF0:bsc0bus)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on samuelh-nd).

